PR #27, opened 2015-03-31
Python3 support

gagern:
The `python` binary invoked by `emsdk` might be Python 3. But the script isn't ready to run under Python 3. With these changes, I at least managed to get a setup installed and activated using Python 3, so it looks good even though I can't guarantee that I got the unicode versus bytes distinction correct in all places. Python 2.7 should be fine as well. And I doubt we should care too much for Python versions older than that.

Compare kripken/emscripten#792 about Python 3 support in emscripten itself.

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -692,7 +692,7 @@ class Tool:
   def __init__(self, data):
     # Convert the dictionary representation of the tool in 'data' to members of this class for convenience.
     for key in data:
-      if sys.version_info < (3,) and isinstance(data[key], str):
+      if sys.version_info < (3,) and isinstance(data[key], unicode):
         setattr(self, key, data[key].encode('Latin-1'))
       else:
         setattr(self, key, data[key])
--- a/emsdk
+++ b/emsdk
@@ -313,7 +313,7 @@ def download_text_file(url, dstpath, download_even_if_exists=False):
   fix_lineendings(os.path.join(emsdk_path(), filename))
 
 def run_get_output(cmd, cwd=None):
-  process = subprocess.Popen(cmd, cwd=cwd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, stdin=subprocess.PIPE, env=os.environ.copy())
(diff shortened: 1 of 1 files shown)


emsdk
```diff
@@ -684,7 +692,7 @@ class Tool:
   def __init__(self, data):
     # Convert the dictionary representation of the tool in 'data' to members of this class for convenience.
     for key in data:
-      if isinstance(data[key], unicode):
+      if sys.version_info < (3,) and isinstance(data[key], str):
```
juj: This changed `unicode` to `str`, why is that? I think this will break people, see https://github.com/juj/emsdk/commit/81aaba9b4e34010564551649bc2e44c8321463b7 and https://github.com/kripken/emscripten/issues/3147 .

juj:
The changes look good in general. I am ok with supporting both python2 and python3 at the same time. Did you test this in python2.7? How about 2.6?

emsdk
gagern: Oh, that's indeed a mistake. `2to3` changed the `unicode` to `str` and I didn't change it back again even though I adjusted the line.

gagern:
Python 2.7 looks good, even though I haven't tested a full compile and install in a non-ASCII directory. Python 2.6 should in theory work as well, since it already provides the `print_function` future import, but since my distribution no longer supports Python 2.6 I haven't tested that.

aidanhs:
Out of curiosity, why was 'universal_newlines' added? That seems like an unrelated change?

gagern:
`universal_newlines` has the side effect of converting between bytes and string using the current locale. Quoting from [the docs](https://docs.python.org/3/library/subprocess.html):

> If `universal_newlines` is `False` the file objects *stdin*, *stdout* and *stderr* will be opened as binary streams, and no line ending conversion is done.  
> If `universal_newlines` is `True`, these file objects will be opened as text streams in universal newlines mode using the encoding returned by `locale.getpreferredencoding(False)`.

juj:
Merged, thanks!

PR #30, opened 2015-04-10
Allow path separator characters in branch names

gunyarakun:
Branch names may contain path separator characters like '/', especially under git-flow.
emsdk fails to build when path separators are contained.
I fixed it with replacing path separators to '-'.


juj:
Looks ok, thanks!

PR #31, opened 2015-04-15
Use $0 to get execution path

fchiba:
$_ is "The last argument of the previous command." (cf. man zshparam) on zsh, so SRC will be "]".

### DIFF
--- a/emsdk_env.sh
+++ b/emsdk_env.sh
@@ -17,7 +17,7 @@
 
 SRC="$BASH_SOURCE"
 if [ "$SRC" = "" ]; then
-  SRC="$_"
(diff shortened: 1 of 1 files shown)


juj:
Ping @aidanhs who worked on previous emsdk changes and knows linux'y behavior much better than me, how does this look to you?

aidanhs:
Yup, looks fine. `$0` in zsh is apparently equivalent to `$BASH_SOURCE` in bash.
I don't know why I retained `$_` in my last PR...

When this is merged I'll follow up with something to fix fish shell.

juj:
Merged, thanks!

PR #40, opened 2015-10-06
Make .emscripten config file relocatable when using the --embedded flag

mhenschel:
I'd like to copy the emsdk directory around freely. The --embedded option almost does the job but it still leaves absolute paths in the .emscripten config file. This change will make all absolute paths derived from the location of the .emscripten file itself. I tested this briefly for my setup and it seems to work. Could you please merge this?


juj:
Thanks, looks good, except a path `\` vs `/` issue. I'll merge and fix locally.

mhenschel:
Thanks for fixing all the issues with that.

PR #43, opened 2015-11-30
Fix tag compilation for Linux 64bit and Unix 32bit

asRIA:
Before this patch it was impossible to compile from tag 64bit version of SDK on Linux host and 32bit version of SDK under Unix operation system.

Tested on Ubuntu with command:
```
./emsdk install sdk-tag-1.35.0-64bit
```

### DIFF
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -842,8 +842,20 @@
     "version": "tag-%tag%",
     "bitness": 64,
     "uses": ["clang-tag-e%tag%-64bit", "node-4.1.1-64bit", "emscripten-tag-%tag%-64bit"],
+    "os": "linux"
+  },
+  {
+    "version": "tag-%tag%",
+    "bitness": 32,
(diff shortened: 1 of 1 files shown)


asRIA:
Actually I don't see any other occurrence of `"os": "unix"`
It could be not supported at all.

asRIA:
poke @juj

juj:
Oh, very sorry for the long delay, email notification about this PR had fallen through my attention. Definitely feel free to ping back earlier if that happens again!

Looking at this, the intent was that in emsdk_manifest, `os: 'unix'` should mean "OS X or linux", so the existing manifest should be ok, but it was in `emsdk` that this check for `unix` was not correct (it was only used in the form of `unix_url` before). Authored this commit https://github.com/juj/emsdk/commit/01f5271887c65e4f84c9c631b4382f3beaa2dac0 to fix this, if you update to that, you should see the 64bit linux tag installable now, and that commit should replace the need for this pull request. Does that work properly?

asRIA:
It works after changes. Thank you!

PR #49, opened 2016-04-13
Explicitly specify python2

bdero:
Allows for compatability with environments where Python 3
is default (like Arch based Linux distros).

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 from __future__ import print_function
 import sys, optparse, subprocess, os, os.path, errno, zipfile, string, json, platform, shutil, tarfile, tempfile, multiprocessing, re, stat, copy


jonas-schievink:
Ping. Without this, a manual change is required on Arch Linux before emsdk will work.

juj:
Oh, sorry, this had slipped through my attention, thanks for pinging.

This PR is not good, as it will break OS X from being able to call emsdk unqualified. We had the same type of issue in Emscripten repository a few years ago.

The resolution for emsdk was to make the emsdk script executable in both python 2.x and 3.x simultaneously, that was done in #27. I believe this means that on Arch Linux one would install python 3 to be able to call this unqualified. I'll close this as fixed with the assumption that that works.

jonas-schievink:
> I believe this means that on Arch Linux one would install python 3 to be able to call this unqualified.

Sadly, that does not work, since `iteritems` was removed in Python 3.

```
  File "./emsdk", line 1797, in <module>
    sys.exit(main())
  File "./emsdk", line 1496, in main
    load_sdk_manifest()
  File "./emsdk", line 1291, in load_sdk_manifest
    for p, v in vars(t2).iteritems():
AttributeError: 'dict' object has no attribute 'iteritems'
```

PR #50, opened 2016-06-22
Fix VS2015 build

bsergeev:
Without this change, building with VS2015 would fail because a wrong MSBuild.exe would be found (through `search_paths_old`).


juj:
Thanks for posting this and sorry for missing this. I think the fix is already in, so latest emsdk master branch should work correctly.

PR #52, opened 2016-07-04
Add a generated file to gitignore

shlomif:


shlomif:
On Thu, 05 Jan 2017 02:23:12 -0800
juj <notifications@github.com> wrote:

> Thanks! These files should get deleted after constructing the environment, so
> should not be seen on the command line. But it's possible to somehow get a
> stray file remain, so good to have these. I pushed this in the form of having
> the `.bat` file in there as well.
> 

Thanks!

-- 
-----------------------------------------------------------------
Shlomi Fish       http://www.shlomifish.org/
https://youtu.be/n6KAGqjdmsk - “Hurt Me Tomorrow”

 <rindolf>  I LOVE YOU ALL WITH YAAKOV’S GREAT HUGE LOVE.
    —  http://www.shlomifish.org/humour/fortunes/sharp-perl.html

Please reply to list if it's a mailing list post - http://shlom.in/reply .

PR #53, opened 2016-08-17
Add option to enable WebAssembly in LLVM

arximboldi:
Recent versions of `sdk-incoming` have WebAssembly support, but there was no way in `emsdk` to build LLVM with support for native Wasm generation (instead of Asm.js to Wasm conversion). 

This commit adds a flag `--enable-wasm` to the `emsdk` tool to enable that in turn adds `-DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=WebAssembly` when building LLVM. This should allow adventurers to try this feature using their beloved `emsdk` without needing to compile LLVM manually.

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -11,7 +11,7 @@ else:
   from urllib2 import urlopen, HTTPError
 
 # EMSDK_DEV is a developer mode flag, which, if true, the SDK is downloaded from a 'staging' online source,
-# instead of the public source. New releases are first deployed to the staging source for testing, before 
+# instead of the public source. New releases are first deployed to the staging source for testing, before
 # being published to the public. Don't enable this unless you develop EMSDK itself and need to access the
 # staging source repository instead.
 EMSDK_DEV = bool(os.getenv('EMSDK_DEV')) if os.getenv('EMSDK_DEV') != None else False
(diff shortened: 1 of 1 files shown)


emsdk
```diff
@@ -1869,6 +1877,10 @@ def main():
       elif sys.argv[i] == '--disable-assertions':
         ENABLE_LLVM_ASSERTIONS = 'OFF'
         sys.argv[i] = ''
+      elif sys.argv[i] == '--enable-wasm':
+        global ENABLE_WASM
+        ENABLE_WASM = True
+	sys.argv[i] = ''
```
juj: The indentation drops here on `sys.argv[i] = ''` by one character. (or perhaps a tabs vs spaces difference?)

emsdk
```diff
@@ -1644,6 +1649,9 @@ def main():
                                   passed to the emsdk activate command to
                                   activate the desired version.
 
+                   --enable-wasm: Try to enable WebAssembly support in the
```
juj: Perhaps just say "Enable WebAssembly support" instead of "Try to enable ..."?

juj:
Thanks! This looks good, happy to have this in. Two minor comments inline.

emsdk
arximboldi: The wording was because at the time it only worked with `sdk-incoming`, but you are right, something more direct is better.

emsdk
arximboldi: You are right. I normally do Python with four spaces so I might have just messed it up. I'll fix it.

arximboldi: Ha! No, it actually was tabs vs spaces :D Fixing anyways.

arximboldi:
Just pushed hopefully addressing your comments.

Bogidon:
This is awesome! I've been having trouble properly compiling the toolchain manually so I can try out the LLVM backend. I think this would make things a lot easier. Any ETA for a merge @juj?

juj:
Thanks for the patience on this PR. Looks great now!

PR #58, opened 2016-12-25
Call emsdk_env.bat without changing curr dir

rongjiecomputer:
Previously user needs to switch current directory to the location of `emsdk_env.bat`, call it, then switch back to previous directory to continue his works.


juj:
Thanks!

PR #60, opened 2017-01-27
Explicitly request a compatible interpreter

3goliad:


rwtolbert:
Seems like this is overcome by events, now that Python 3 is preferred.

sbc100:
And the file got renamed to `emsdk.py`, but still if you would like to replace this with python3 this change would still be welcome.

rwtolbert:
this actually isn't my work. Was just looking through some of the backlog to see if I could help or if some of it was simply out-of-date

PR #64, opened 2017-02-16
Add support for SOCKS proxies to emsdk.

de-wim:
The emsdk command now checks for the SOCKS_PROXY environment variable and tries to
use the PySocks library to wrap `socket.socket` when it is set.

This allows emsdk to be used on IPv6-only hosts, since GitHub is IPv4 only and IPv6 support
for AWS is poor at best.

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -2,6 +2,20 @@
 
 from __future__ import print_function
 import sys, optparse, subprocess, os, os.path, errno, zipfile, string, json, platform, shutil, tarfile, tempfile, multiprocessing, re, stat, copy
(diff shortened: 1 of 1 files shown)


wvdschel:
I use this on a rented server that I use as a build bot. IPv4 addresses cost extra, so it's currently IPv6 only.

I hope this change is acceptable to be merged upstream, as `emsdk update` overwrites the emsdk script and I have to repatch it every time I update at the moment.

sbc100:
I assume this is no longer needed?

PR #66, opened 2017-03-03
Fix fatal error when git shallow clone is enabled

Shnatsel:
Fix "error: pathspec 'incoming' did not match any file(s) known to git." error when GIT_CLONE_SHALLOW is true.

Background: http://stackoverflow.com/a/23710503/585725


sbc100:
I assume this is no longer needed?

PR #74, opened 2017-05-05
Fix random errors fetching from github

parasyte:
It seems github requires a specific URL format for reaching repos. `emsdk update-tags` will randomly fail when the convention is not followed.


maxbrunsfeld:
Without this, the `update-tags` command does not work for me, with no output.

juj:
Thanks, looks good!

PR #76, opened 2017-05-26
Add Cygwin Support

kawakami-o3:
treat Cygwin enviroment as Linux.

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -36,7 +36,7 @@ if os.name == 'nt':
   ENVPATH_SEPARATOR = ';'
 
 LINUX = False
-if platform.system() == 'Linux':
+if platform.system() == 'Linux' or os.name == 'posix':
   LINUX = True
   ENVPATH_SEPARATOR = ':'
 


tdaede:
This gets farther, however it's not enough as emcc gets run with cygwin python2, but emcc can't handle cygwin paths.

kawakami-o3:
Thanks. I'll continue to test my patch.
Could you show me any error logs?

kawakami-o3:
I got this error.
```
[ 99%] Linking CXX shared library ../../../../bin/cygclang-4.dll
c++: error: Files/Emscripten/clang/tag-e1.37.13/build_tag-e1.37.13_64/tools/clang/tools/libclang/libclang.exports: No such file or directory
make[2]: *** [tools/clang/tools/libclang/CMakeFiles/libclang.dir/build.make:656: bin/cygclang-4.dll] error 1
make[1]: *** [CMakeFiles/Makefile2:12424: tools/clang/tools/libclang/CMakeFiles/libclang.dir/all] error 2
```
I will make an additional patch.

Any other errors? or suggestion?

juj:
Thanks, this looks good, I think.

The above error might be worked around by locating emsdk to some directory that does not have a space in it, i.e. instead of using the NSIS installer, bootstrap with `git clone https://github.com/juj/emsdk.git` to a custom directory without spaces in it. (We are in the process of removing support for the NSIS installer althogether in the future)

juj:
Actually, I realize this is not quite correct, since it will cause Cygwin to attempt to download Linux binaries. Also, I now see that there is a regression that OS X was misidentified as LINUX as well, so this  should probably be reverted. I did a workaround at above commit that reidentifies OS X as only OS X for now.

PR #82, opened 2017-06-10
Remove trailing slashes from git URLs

postmillenial:
breaks on git, or github, or something...

I believe this is the issue being described in https://github.com/WebAssembly/binaryen/issues/938


juj:
Thanks @postmillenial !

When I navigate to e.g. https://github.com/kripken/emscripten-fastcomp, and click on the clone button, I see that it prints `https://github.com/kripken/emscripten-fastcomp.git`, i.e. it gives the suffix `".git"` on all these URLs. So let me do this change in that format and add that suffix to the URLs as well, perhaps that is the most recommended format by GitHub. I hope that should work as well, let me know if it causes any issues.

PR #91, opened 2017-08-02
Avoid pywin32 dependency on Windows

segevfiner:
Instead use `winreg`/`_winreg` for the registry and `ctypes` for the `SendMessage` call.

This dependency has not been described anywhere. It's not a problem for someone using a pre-built emsdk, which bundles Python with pywin32, but it's slightly annoying for someone using a Git checkout.

I think it also won't work at all from a mingw built Python (In MSYS2 for example, when using the mingw Python and not the MSYS Python), though that's not really a supported configuration now, we may want to support it in the future (emsdk thinks it's on Windows, but it doesn't really know how to handle Windows and a MSYS2/bash shell, emsdk_env.sh doesn't work since it only generates `emsdk_set_env.bat`).

I didn't fully test `win_set_environment_variable_direct`. I only tested the code in the interpreter by copy and paste.

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -168,12 +168,16 @@ def win_set_environment_variable_direct(key, value, system=True):
     if py:
       py_path = to_native_path(py.expand_vars(py.activated_path))
       os.environ['PATH'] = os.environ['PATH'] + ';' + py_path
-    import win32api, win32con
+    try:
+        import winreg
+    except ImportError:
+        import _winreg as winreg
+    import ctypes.wintypes
     if system: # Read globally from ALL USERS section.
-      folder = win32api.RegOpenKeyEx(win32con.HKEY_LOCAL_MACHINE, 'SYSTEM\\CurrentControlSet\\Control\\Session Manager\\Environment', 0, win32con.KEY_ALL_ACCESS)
+      folder = winreg.OpenKeyEx(winreg.HKEY_LOCAL_MACHINE, 'SYSTEM\\CurrentControlSet\\Control\\Session Manager\\Environment', 0, winreg.KEY_ALL_ACCESS)
     else: # Register locally from CURRENT USER section.
-      folder = win32api.RegOpenKeyEx(win32con.HKEY_CURRENT_USER, 'Environment', 0, win32con.KEY_ALL_ACCESS)
-    win32api.RegSetValueEx(folder, key, 0, win32con.REG_EXPAND_SZ, value)
+      folder = winreg.OpenKeyEx(winreg.HKEY_CURRENT_USER, 'Environment', 0, winreg.KEY_ALL_ACCESS)
+    winreg.SetValueEx(folder, key, 0, winreg.REG_EXPAND_SZ, value)
     if VERBOSE: print('Set key=' + key + ' with value ' + value + ' in registry.')
   except Exception as e:
     if e[0] == 5 or e[2] == 'Access is denied.':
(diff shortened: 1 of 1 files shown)


juj:
Thanks, in general I think this is a good approach if it allows us to avoid the dependency. I'll do testing to figure out if this works well, the registry setting rules for PATH are quite subtle.

sbc100:
I know its old but any interest in landing this still?  Or should we close it out?  Seems like its useful.

aminya:
I have had many problems with this. pywin32 also does not work properly most of the time! Can we get this rebased?

sbc100:
What do you mean by "pywin32 also does not work properly most of the time"? 

I'm not sure how this change would help with that since we ship pywin32 with the emsdk version of python so it will still be found and used consistently.

It sounds like you want an option to simply not use it at all?   Perhaps you could describe the problems you are seeing and we can see if we can find a solution that works consistently.

aminya:
> What do you mean by "pywin32 also does not work properly most of the time"?

- It uses "Machine" environment variables instead of "User". This requires "administrator" elevation for adding environment variables and updating PATH. 
- It fails to add the environment variables randomly.
- If it is missing, the whole emsdk will be broken.

One of the reasons I made chocolatey-emscripten was this. It works just with "PowerShell". No dependency is needed, and it works perfectly.
https://github.com/aminya/chocolatey-emscripten

sbc100:
But it looks like chocolatey-emscripten uses emsdk under the hood.. so how does it solve the problem?

aminya:
> But it looks like chocolatey-emscripten uses emsdk under the hood.. so how does it solve the problem?

See [add_envs](https://github.com/aminya/chocolatey-emscripten/blob/master/tools/add_envs.ps1), [add_paths](https://github.com/aminya/chocolatey-emscripten/blob/master/tools/add_paths.ps1), and [remove_envs](https://github.com/aminya/chocolatey-emscripten/blob/master/tools/remove_envs.ps1)

sbc100:
> > What do you mean by "pywin32 also does not work properly most of the time"?
> 
> * It uses "Machine" environment variables instead of "User". This requires "administrator" elevation for adding environment variables and updating PATH.
> * It fails to add the environment variables randomly.
> * If it is missing, the whole emsdk will be broken (#615 will fix this).

We should look into fixing all these for sure.  Sounds like (3) is fixed already.   

We should certainly look at (1) and (3).   From looking at the code it looks like we only try set permanent environment variables if `--global` is passed.    Is that the mode you are talking about.   I think there some discussion of adding `--machine` which could be combined with `--global` and could default to false.

aminya:
> We should certainly look at (1) and (3). From looking at the code it looks like we only try set permanent environment variables if `--global` is passed. Is that the mode you are talking about. I think there some discussion of adding `--machine` which could be combined with `--global` and could default to false.
 
You can set environment variables permanently for the current user only. You don't need "admin" access for that.
Does `--global` mean for all users on a system?

sbc100:
> > We should certainly look at (1) and (3). From looking at the code it looks like we only try set permanent environment variables if `--global` is passed. Is that the mode you are talking about. I think there some discussion of adding `--machine` which could be combined with `--global` and could default to false.
> 
> You can set environment variables permanently for the current user only. 

Right, but emsdk currently doesn't have an option for that I think, or at least it got removed at some point.  Would welcome a patch to at it back.

> Does `--global` mean for all users on a system?

I think `--global` today means "in the registry" whereas without `--global` only the current shell is effected.   The meaning may have changed over time I'm just looking at what it currently means.

sbc100:
Ah, we already have a patch here .. we should try to land it: https://github.com/emscripten-core/emsdk/pull/468

PR #105, opened 2017-09-29
Support PowerShell environment

### LINKED ISSUE #100: add PowerShell env support
PowerShell is the new default shell for Windows 10, and it would be nice if emsdk could generate the proper env for it.

PowerShell syntax (in a `emsdk_set_env.ps1` file) would look like:
```
$env:PATH = "C:\proj\emsdk;$env:PATH"
$env:EMSDK = "C:\proj\emsdk"
```
etc.

SaschaNaz:
Fixes #100.

### DIFF
--- a/emsdk.bat
+++ b/emsdk.bat
@@ -1,5 +1,3 @@
-@SET PREVPATH=%PATH%
-
 :: Find python from an explicit location relative to the Emscripten SDK.
 @IF EXIST "%~dp0python\2.7.5.3_64bit\python.exe" (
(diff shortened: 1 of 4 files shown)


emsdk.bat
```diff
@@ -1,5 +1,3 @@
-@SET PREVPATH=%PATH%
```
juj: Hmm, I know I had some grand reason to guarding the `PATH` variable before and after, but searching the commit history, I am unable to find it. This may have been a leftover, perhaps I had at some point appended the directory to PYTHON to PATH, instead of invoking it explicitly.

juj:
Thanks!

PR #114, opened 2017-11-13
Fix powershell env directory

saschanaz:
It should use `./` to access the current directory.

(I added emsdk to the system path so I couldn't catch this. Sorry 😅)

### DIFF
--- a/emsdk.ps1
+++ b/emsdk.ps1
@@ -29,9 +29,9 @@ $env:EMSDK_POWERSHELL = 1
 # python is not able to set environment variables to the parent calling process, so
 # therefore have it craft a .ps1 file, which we invoke after finishing python execution,
 # to set up the environment variables
-if (Test-Path "emsdk_set_env.ps1") {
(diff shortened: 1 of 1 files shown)


jakobmulvad:
Was just about to open the same pull request. This was my fix as well.

DTasev:
@juj Can confirm this works with the current master, `./emsdk_env.ps1` can now properly set the environment. Please merge this in or remove the PowerShell scripts as of right now the only thing they do is get you halfway there and give an error.

saschanaz:
Sorry to bother you twice, but would be happy if you review this 😆 @juj

kripken:
@juj, if you're too busy to review&merge stuff in a timely manner in this repo, perhaps someone else should be given merge powers?

juj:
Sorry for the delay, this had fallen through my attention!

@kripken : that's a good observation, even if this was my full time attention, we should still have more admins on this repo. I'll make you an administrator on this as well to get started on getting more collaborators.

PR #115, opened 2017-11-13
Catch KeyboardInterupt in download_file

tejom:
When interrupting a download with something like control c the program will quit but won't clean it self up and will leave an invalid file. There is a handler for Exception, but this wont catch interrupts. If you try to rerun the program it won't try to re-download it the next time.

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -487,6 +487,10 @@ def download_file(url, dstpath, download_even_if_exists=False, filename_prefix =
     print("Error downloading URL '" + url + "': " + str(e))
     rmfile(file_name)
     return None
+  except KeyboardInterrupt as e:
+    print("Received Interrupt, exiting")
+    rmfile(file_name)
+    exit(0)
   return file_name
 
 def download_text_file(url, dstpath, download_even_if_exists=False, filename_prefix=''):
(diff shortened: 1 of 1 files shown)


emsdk
```diff
@@ -487,6 +487,10 @@ def download_file(url, dstpath, download_even_if_exists=False, filename_prefix =
     print("Error downloading URL '" + url + "': " + str(e))
     rmfile(file_name)
     return None
+  except KeyboardInterrupt as e:
+    print("Received Interrupt, exiting")
```
juj: Can you change this to something like `"Aborted by User"`?

emsdk
```diff
@@ -487,6 +487,10 @@ def download_file(url, dstpath, download_even_if_exists=False, filename_prefix =
...
+    rmfile(file_name)
+    exit(0)
```
juj: `exit(1)` would be better here, or otherwise if this was chained to scripts, it might continue executing the next script believing `emsdk` run had succeeded?

tejom: Makes sense, the exit code the way it is now is 1.

juj:
Perfect, thanks!

PR #119, opened 2017-11-26
Add a license file.

shlomif:


shlomif:
Ping @juj!

juj:
Indeed good to have a license file here as well. The intent has been that emsdk is carried with the same MIT license as Emscripten, as mentioned here. Added a LICENSE file in this commit https://github.com/juj/emsdk/commit/1b59d943c9336f2a15b34f8f51eeb074a89997e4 , which slightly adjusts the authors line to refer to Emscripten's authors, otherwise the same. Thanks!

shlomif:
On Thu, 01 Feb 2018 22:55:13 +0000 (UTC)
juj <notifications@github.com> wrote:

> Indeed good to have a license file here as well. The intent has been that
> emsdk is carried with the same MIT license as Emscripten, as mentioned here.
> Added a LICENSE file in this commit
> https://github.com/juj/emsdk/commit/1b59d943c9336f2a15b34f8f51eeb074a89997e4 ,
> which slightly adjusts the authors line to refer to Emscripten's authors,
> otherwise the same. Thanks!
> 

thanks!



-- 
-----------------------------------------------------------------
Shlomi Fish       http://www.shlomifish.org/
Humanity - Parody of Modern Life - http://shlom.in/humanity

When people agree with me I always feel I must be wrong.
    — https://en.wikiquote.org/wiki/Oscar_Wilde (via fortune-mod)

Please reply to list if it's a mailing list post - http://shlom.in/reply .

PR #122, opened 2017-12-22
Use key to sort versions

saschanaz:
Python supports sorting by arrays:

```py
[1,1,0] > [1,0,0] # True
[1,1,0] > [1,3,0] # False
[1,1,0] < [1,3,0] # True
[1,0,0,1] > [1,0,0] # True
```

We can use this to reduce codes.


juj:
Oh, this is just superb, quite a big cleanup!

PR #123, opened 2017-12-22
Provide Node.js 8 also for precompiled tools

saschanaz:
Current emsdk only provides it for incoming and tagged downloads, it should also provide it for precompiled ones.


juj:
Very good find, thanks!

PR #125, opened 2018-01-03
list sdk depends

umeboshi2:
I did this to list the dependencies on each sdk, so that I can install the packages without interfering with my nodejs environment.  I don't think that this can be used with ```--old``` due to ```sys.argv``` being used directly, instead of using argparse.  I am not sure the formatting is acceptable, but I think it is valuable to gather the info quickly on the command line, instead of digging through the json manifest.

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -2168,6 +2168,7 @@ def main():
   if cmd == 'list':
     print('')
     show_old = len(sys.argv) >= 3 and sys.argv[2] == '--old'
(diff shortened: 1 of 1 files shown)


umeboshi2:
I should have clarified, but ```--uses``` isn't in the help text because it seems to conflict with ```--old```.

juj:
Looks like a nice feature, thanks!

PR #126, opened 2018-01-03
update tags

umeboshi2:
This is just an update of the tag lists.  It looks like the tag lists haven't been updated in the repo as frequently as in the past.  I do understand the instructions do say to run ```./emsdk update-tags``` after a clone, however many may feel that the tags are pretty close to updated when the repo is cloned.  I had to look at the emsdk script to determine what "old" means, and it's all but the last two tags.  By that measurement, even the pull request contains old versions.  Perhaps the repo can be automated to update the tag lists every two versions.


juj:
The tags are semi-periodically updated, unfortunately there is no automated mechanism to do this. It looks like as part of a recent fix we got the tags in already, so I'll close this one, but the new tags are present in the tree now so should be good. Thanks!

PR #128, opened 2018-01-11
Python 3: version_key() should return list

saschanaz:
`map()` on Python 3 does not return a list but an iterator, so this PR fixes it. (Sorry for this 😅)


juj:
Thanks!

PR #131, opened 2018-01-29
Update emsdk

francoguilherme:


juj:
Not quite sure about this, this looks like short-circuiting, perhaps some leftover test code?

PR #141, opened 2018-04-27
Do not link to emsdk-portable.*, link to the main website docs

kripken:
We no longer have the resources to produce and maintain `emsdk-portable.*`. The emscripten main docs have already been changed to reflect that, but the emsdk repo had links to them. This replaces those links with a link to the main website docs, so we just have a single central location for how to get started with the sdk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,10 +6,7 @@ You can also set up Emscripten from source, without the pre-built SDK, see "Inst
 
 ## Downloads
 
-To get started with Emscripten development, grab one of the packages below:
-
-* Windows: [emsdk-portable.zip](https://s3.amazonaws.com/mozilla-games/emscripten/releases/emsdk-portable.zip)
-* Linux and Mac OS X: [emsdk-portable.tar.gz](https://s3.amazonaws.com/mozilla-games/emscripten/releases/emsdk-portable.tar.gz)
+To get started with Emscripten development, see the [Emscripten website documentation](https://kripken.github.io/emscripten-site/docs/getting_started/downloads.html).
 
 **Old Releases** are available in the **Archived Releases** section below.
 


juj:
> We no longer have the resources to produce and maintain emsdk-portable.*

The buildbot infrastructure that is already up at http://ec2-54-213-252-128.us-west-2.compute.amazonaws.com:8112/#/builders has the machinery to produce the portable zips. The new zips are created by going to this page http://ec2-54-213-252-128.us-west-2.compute.amazonaws.com:8112/#/builders and clicking on each of the `osx-emsdk-updater`, `ubuntu-emsdk-updater` and `win-emsdk-updater` and then clicking on the `force-emsdk-updater` button on the top right. For example, clicking on `ubuntu-emsdk-updater` leads to http://ec2-54-213-252-128.us-west-2.compute.amazonaws.com:8112/#/builders/10, which has `force-emsdk-updater` button to kick off a new deployment of the portable zips.

Or do you mean that you are looking to close down this machinery, so we should not depend on these anymore?

kripken:
Yeah, sorry for not being clear - the scripts are there, but in practice, there is only me to run them, and I don't have a way to test them on non-linux platforms and fix issues if any arise. So I don't think we have a choice but to not depend on them, at least for now - I do hope we find the resources to pick this up again some day, I'm trying my best at that.

juj:
Ah right, I understand. Yeah, let's then go with this plan of action.

One issue with depending on `git clone` to bootstrap is that existing `emsdk-portable.zip` based installations will no longer receive new updates when they do `emsdk update`, since the triggers are no longer run, but they will need to manually change from the emsdk-portable.zip method over to a `git clone` based installation. In that method, `git pull` is then used to update emsdk.

Perhaps we should make a `stable` vs `master` branch distinction in emsdk as well, and make `stable` branch the default? So that `master` branch can then receive new commits that might not be as tested yet, and then merge to `stable` when tested to work? In any case, this PR looks good, I'll merge that in.

kripken:
About branches, maybe they could mirror the emscripten branches, so `incoming/master` instead of `master/stable`?

PR #145, opened 2018-05-15
Build binaryen from source when building tagged sdk versions from source

jgravelle-google:
We already include a build of binaryen (as a subdirectory of clang) when fetching prebuilt binaries, but when building from source we don't build binaryen as well. As we move to requiring binaryen be set via BINARYEN_ROOT in the config file (which emsdk already handles writing), we'll always need binaryen built. This should make that easier for users.

### DIFF
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -982,67 +982,67 @@
   {
     "version": "incoming",
     "bitness": 32,
-    "uses": ["clang-incoming-32bit", "node-8.9.1-32bit", "python-2.7.13.1-32bit", "java-8.152-32bit", "emscripten-incoming-32bit"],
+    "uses": ["clang-incoming-32bit", "node-8.9.1-32bit", "python-2.7.13.1-32bit", "java-8.152-32bit", "emscripten-incoming-32bit", "binaryen-incoming-32bit"],
     "os": "win"
   },
   {
     "version": "incoming",
     "bitness": 64,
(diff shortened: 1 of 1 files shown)


kripken:
Hmm I've never looked at this code before. What does this manifest file do?

jgravelle-google:
It initializes the `Tool`s that emsdk uses to build/fetch things: https://github.com/juj/emsdk/blob/a4474e59db658cea570c78254fa71119cf688db5/emsdk#L1658-L1662

So, the manifest is how it specifies what repos to sync, or where precompiled binaries are. It also has support for sdks, which are collections of tools that go together. This is how we get clang+node+python+java+emscripten when using emsdk to get latest.

kripken:
I see, thanks. Well, the change looks good to me as best as I can tell. Maybe we can wait a little more to see if @juj can take a look. If not I can merge this.

I guess there are no tests for this?

jgravelle-google:
I am ok with waiting, especially because I have no idea if there are tests for this.

jgravelle-google:
@juj @kripken ?
(I was going to just hit merge and hope for the best, but turns out I can't actually do that)

juj:
Thanks, this is great. Originally I did not make this the default, since Binaryen was optional for such a long time, and people might want to download Emscripten and use it with only asm.js without using Binaryen, but that was a long time ago and now since Wasm is the default, this makes sense to do.

juj:
> I am ok with waiting, especially because I have no idea if there are tests for this.

The tests for this are basically part of emslave when it pulls and builds incoming or any of the tags for testing or tags packaging, so the operation in the emslave repo. But apart from that, there is no dedicated testing suite, it has been manual testing.

PR #146, opened 2018-05-30
Support for building upstream LLVM from source

jgravelle-google:
@kripken @juj @dschuff

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -791,7 +791,16 @@ def xcode_sdk_version():
     return subprocess.checkplatform.mac_ver()[0].split('.')
 
 def build_fastcomp_tool(tool):
-  if VERBOSE: print('build_fastcomp_tool(' + str(tool) + ')')
+  return build_llvm_tool(tool, build_js=True)
+
+def build_upstream_llvm_tool(tool):
+  global ENABLE_WASM
+  # Upstream LLVM has no JSBackend, relies on wasm backend
+  ENABLE_WASM = True
+  return build_llvm_tool(tool, build_js=False)
+
+def build_llvm_tool(tool, build_js):
+  if VERBOSE: print('build_llvm_tool(' + str(tool) + ')')
   fastcomp_root = tool.installation_path()
   fastcomp_src_root = os.path.join(fastcomp_root, 'src')
   if hasattr(tool, 'git_branch'): # Does this tool want to be git cloned from github?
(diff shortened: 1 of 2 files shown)


jgravelle-google:
@juj @kripken ?
This is probably going to start being useful pretty quick.
We do also need to get an upstream LLVM build on the prebuilts, which I'm not sure how we do.

emsdk_manifest.json
```diff
@@ -145,6 +145,34 @@
...
   },
+  {
+    "id": "upstream-clang",
+    "version": "master",
+    "bitness": 32,
+    "install_path": "clang/upstream",
+    "git_branch": "master",
+    "url": "https://github.com/llvm-mirror/llvm.git",
+    "clang_url": "https://github.com/llvm-mirror/clang.git",
+    "lld_url": "https://github.com/llvm-mirror/clang.git",
```
juj: Should this be `https://github.com/llvm-mirror/lld.git`?

jgravelle-google: Yes it should... why did I think this worked locally?
Oh wow I had the 64-bit version correct, so it happened to work on my machine.

emsdk
```diff
@@ -821,7 +834,8 @@ def build_fastcomp_tool(tool):
 
   enable_assertions = ENABLE_LLVM_ASSERTIONS.lower() == 'on' or (ENABLE_LLVM_ASSERTIONS == 'auto' and build_type.lower() != 'release' and build_type.lower() != 'minsizerel')
 
-  args = ['-DLLVM_TARGETS_TO_BUILD=X86;JSBackend', '-DLLVM_INCLUDE_EXAMPLES=OFF', '-DCLANG_INCLUDE_EXAMPLES=OFF', '-DLLVM_INCLUDE_TESTS=' + tests_arg, '-DCLANG_INCLUDE_TESTS=' + tests_arg, '-DLLVM_ENABLE_ASSERTIONS=' + ('ON' if enable_assertions else 'OFF')]
+  js_target = ';JSBackend' if build_js else ''
```
juj: How about this:

```python
only_supports_wasm = hasattr(tool, 'only_supports_wasm')

targets_to_build = 'X86'
if not only_supports_wasm: targets_to_build += ';JSBackend'
```

and then below:

```c++
if ENABLE_WASM or only_supports_wasm:
  args += ['-DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=WebAssembly']
```

That would then allow deleting the function `build_upstream_llvm_tool()` altogether (`ENABLE_WASM` variable can be left in whatever user-specified state in that case), and `def build_llvm_tool(tool, build_js):` -> `def build_llvm_tool(tool):`

emsdk_manifest.json
```diff
@@ -145,6 +145,34 @@
...
+    "bitness": 64,
+    "install_path": "clang/upstream",
+    "git_branch": "master",
+    "url": "https://github.com/llvm-mirror/llvm.git",
+    "clang_url": "https://github.com/llvm-mirror/clang.git",
+    "lld_url": "https://github.com/llvm-mirror/lld.git",
+    "custom_install_script": "build_llvm",
+    "activated_path": "%installation_dir%/%fastcomp_build_bin_dir%",
+    "activated_cfg": "LLVM_ROOT='%installation_dir%/%fastcomp_build_bin_dir%'",
+    "cmake_build_type": "Release"
```
juj: and then add `only_supports_wasm: true` here?

emsdk
jgravelle-google: Sure, done

juj:
Perfect, many thanks!

Issue #148, opened 2018-06-22
Unable to compile incoming after  #145


asRIA:
Hi,
After #145 was merged I started to observe error in binaryen compilation: 
Executed command: 
```
./emsdk install --build=MinSizeRel sdk-incoming-64bit -j8
```

Result:
```
[100%] Linking CXX executable optimizer
[100%] Built target optimizer
Done installing tool 'emscripten-incoming-64bit'.
Manifest error: No tool by name 'binaryen-incoming-64bit' found! This may indicate an internal SDK error!
Traceback (most recent call last):
  File "./emsdk", line 2350, in <module>
    sys.exit(main())
  File "./emsdk", line 2333, in main
    success = tool.install()
  File "./emsdk", line 1337, in install
    success = tool.install()
AttributeError: 'NoneType' object has no attribute 'install'
```

Moreover after `emsdk list`
```
...
The following SDKs can be compiled from source:
         sdk-incoming-32bit       
Manifest error: No tool by name 'binaryen-incoming-64bit' found! This may indicate an internal SDK error!
Manifest error: No tool by name 'binaryen-incoming-64bit' found! This may indicate an internal SDK error!
         sdk-incoming-64bit       
         sdk-master-32bit         
         sdk-master-64bit         
         sdk-tag-1.38.5-32bit     
```

Issue was introduced by 16125272b28ba8a6a65ddde81924aa894d65080b

CC:
@jgravelle-google

asRIA:
After some investigation I've found that adding
```
  {

    "id": "binaryen",

    "version": "incoming",

    "bitness": 32,

    "append_bitness": false,

    "url": "https://github.com/WebAssembly/binaryen.git",

    "git_branch": "master",

    "activated_cfg": "BINARYEN_ROOT='%installation_dir%%generator_prefix%_32bit_binaryen'",

    "activated_path": "%installation_dir%%generator_prefix%_32bit_binaryen/bin",

    "activated_env": "BINARYEN_ROOT=%installation_dir%%generator_prefix%_32bit_binaryen",

    "cmake_build_type": "Release",

    "custom_install_script": "build_binaryen",

    "custom_is_installed_script": "is_binaryen_installed",

    "custom_uninstall_script": "uninstall_binaryen"

  },

  {

    "id": "binaryen",

    "version": "incoming",

    "bitness": 64,

    "append_bitness": false,

    "url": "https://github.com/WebAssembly/binaryen.git",

    "git_branch": "master",

[...]

PR #149, opened 2018-06-22
Fixed #148 - unable to compile sdk-incoming on linux

asRIA:
An attempt to fix problem with compiling emscripten with command: 
```
./emsdk install sdk-incoming-64bit -
```

PR #150, opened 2018-06-22
There is no binaryen-incoming, point incoming sdks at binaryen-master

jgravelle-google:
Fixes https://github.com/juj/emsdk/issues/148


asRIA:
I'm fine as long as it fixes problem.
One comment though is that without context either of fix looks like a copy-paste error. I'm closing #149

asRIA:
I can confirm that it fixes #148 (y)

PR #154, opened 2018-07-03
Add --notty option to force non-tty output

sbc100:
This is useful on travis or circle-ci were we dont
want to show hundreds of lines of downloading progress.
These systems preport to be TTYs but don't support \t
or \b.

Also, preform logging through a separate function and
direct it to stderr so it gets automatically flushed.


emsdk
```diff
@@ -18,7 +18,7 @@ else:
 EMSDK_DEV = bool(os.getenv('EMSDK_DEV')) if os.getenv('EMSDK_DEV') != None else False
 
 if EMSDK_DEV:
-  print('EMSDK_DEV active.')
+  log_message('EMSDK_DEV active.')
```
kripken: don't we need to define `log_message` before this first use?

kripken:
I opened #155 with travis support.

emsdk
```diff
@@ -158,9 +159,12 @@ def cmake_generator_prefix():
   elif CMAKE_GENERATOR == 'MinGW Makefiles': return '_mingw'
   return '' # Unix Makefiles and Visual Studio 2013 do not specify a path prefix for backwards path compatibility
 
+def log_message(msg, end='\n'):
+  sys.stderr.write(msg + end)
```
juj: This changes all printing to go to `stderr` instead of `stdout`? We certainly don't want that. If you want unbuffered output, you can run the script with `python -u emsdk.py` or set `PYTHONUNBUFFERED` environment variable. See https://stackoverflow.com/questions/107705/disable-output-buffering

(If we diddwant to wrap printing for some reason, it would be better to assign over the `print` function instead of change `print` to `log_message` everywhere)

emsdk
```diff
@@ -462,35 +466,46 @@ def download_file(url, dstpath, download_even_if_exists=False, filename_prefix =
...
     with open(file_name, 'wb') as f:
       file_size = get_content_length(u)
-      if file_size > 0: print("Downloading: %s from %s, %s Bytes" % (file_name, url, file_size))
-      else: print("Downloading: %s from %s" % (file_name, url))
+      if file_size > 0: log_message("Downloading: %s from %s, %s Bytes" % (file_name, url, file_size))
+      else: log_message("Downloading: %s from %s" % (file_name, url))
 
       file_size_dl = 0
+      progress_shown = 0
+      progress_max = 76
```
juj: What is `progress_max = 76`?

juj:
Force-disabling TTY for environments that do report they are a TTY based environment but in practice are not is a good feature, though let's do that in a separate PR, I think most of the changes in this PR are not actually needed in order to get rid of output buffering.

PR #156, opened 2018-07-05
binaryen.js is no longer bundled in binaryen

kripken:
And emscripten doesn't need it anyhow (it's for handwritten js users).

This binaryen change broke tagged version building,

http://ec2-54-213-252-128.us-west-2.compute.amazonaws.com:8112/#/builders/5/builds/193

Hopefully this will fix it, although I'm unsure how the bots receive updates to the emsdk script - @juj, do you know?

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -937,7 +937,6 @@ def build_binaryen_tool(tool):
   shutil.copytree(os.path.join(src_root, 'scripts'), os.path.join(build_root, 'scripts'))
   remove_tree(os.path.join(build_root, 'src', 'js'))
   shutil.copytree(os.path.join(src_root, 'src', 'js'), os.path.join(build_root, 'src', 'js'))
-  shutil.copyfile(os.path.join(src_root, 'bin', 'binaryen.js'), os.path.join(build_root, 'bin', 'binaryen.js'))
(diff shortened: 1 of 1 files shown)


kripken:
Thanks @juj, sounds good!

@asRIA - that is a related issue, caused by the change that made this fix necessary. It should be fixed already on incoming, see https://github.com/kripken/emscripten/pull/6815

asRIA:
No worries at all. These projects are awesome, and you are doing outstanding work. If some help is needed with releases I might help.

asRIA:
After some reflection, this change actually makes impossible to use every version of emscripten what's tagged. 
Maybe let's make it opt-in somehow? Or keep current logic for compatibility reason (what's a technical dept unfortunately)

kripken:
Looks like 1.38.7 built ok on the bots on ubuntu and windows so far, os x still working. So I think we are ok now?

Regarding helping - testing builds and notifying about any issues is very helpful, thanks @asRIA!

kripken:
Oh wait, we aren't ok yet, as the emscripten fix for this issue landed after 1.38.7. I'll tag 1.38.8 now.

kripken:
Tagged 1.38.8, hopefully after that all is good.

asRIA:
1.38.8 is stable 👍

kripken:
Great, thanks for confirming!

PR #159, opened 2018-07-10
Added tests folder so that emsdk verification can be done

Mikerah:
While installing and verifying my installation of emsdk, I noticed that the tests folder that the verification how-to mentions is not in the repo.

I have added it so that others can easily test their installation of emsdk.

### DIFF
--- a/tests/hello_world.cpp
+++ b/tests/hello_world.cpp
@@ -1,0 +1,7 @@
+#include <iostream>
+
+using namespace std;
+
+int main() {
+    cout << "Hello World!" << endl;
+}


kripken:
Thanks!

Which how-to doc did you see that in? It's possible it's a mistake we should fix there. (Although, there's not much harm in adding this file here.)

Mikerah:
The document is here: https://kripken.github.io/emscripten-site/docs/building_from_source/verify_emscripten_environment.html

kripken:
I see, thanks @Mikerah. Those docs look ok, so yeah, let's add this file.

However, how about adding a more minimal hello world, like this?
```
#include <stdio.h>

int main() {
  printf("hello, world!\n");
  return 0;
}
```

Using C++ iostreams will also build libc++ and libc++abi, which takes a lot longer.

Mikerah:
I don't think it matters that much since it's a small enough example that the fact that libc++abi and libc++ need to be built is unnoticeable (at least on my machine).

juj:
The file in question that the docs refer to is located here: https://github.com/kripken/emscripten/blob/incoming/tests/hello_world.cpp

The intent of the docs is to ask user to move to the Emscripten directory after installing a particular version with emsdk. That is stated at

```
The next test is to actually build some code! On the command prompt navigate to the Emscripten directory for the current SDK and try to build the hello_world.cpp test code:
```

in https://kripken.github.io/emscripten-site/docs/building_from_source/verify_emscripten_environment.html#build-a-basic-example.

Perhaps the docs should be improved instead? Authored https://github.com/kripken/emscripten/pull/6851 for one example how to do so.

PR #164, opened 2018-07-18
Add upstream waterfall tools and sdk

jgravelle-google:
This fetches the latest revision of the prebuilt wasm backend off the wasm waterfall (https://wasm-stat.us)
Only tested on linux, and the linux build is the only one that's regularly green at the moment, so that's all that's supported for now.

Going forward I'm not sure what sort of release process we want for mapping emscripten tags with upstream LLVM commits. Some ideas I can think of are:
1) Fork LLVM, create a new branch off of latest master when we tag an emscripten version, with the only diff being the addition of an `emscripten-version.txt`
2) Don't track LLVM commits, but LLVM releases (6.0.0, 5.0.2, etc)
3) Some explicitly managed lookup for the revisions, living either in emsdk or emscripten
4) Include `emscripten-version.txt` in upstream LLVM

I think I like 3) the best but in theory we don't need to answer that right away.

### DIFF
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -176,8 +176,8 @@
     "cmake_build_type": "Release"
   },
   {
-    "id": "upstream-clang",
-    "version": "%waterfall-lkgr%",
+    "id": "clang",
+    "version": "upstream-%waterfall-lkgr%",
     "bitness": 64,
     "linux_url": "https://storage.googleapis.com/wasm-llvm/builds/linux/%waterfall-lkgr%/wasm-binaries.tbz2",
     "install_path": "upstream/%waterfall-lkgr%",
@@ -185,8 +185,8 @@
     "activated_cfg": "LLVM_ROOT='%installation_dir%/bin'"
   },
   {
-    "id": "upstream-emscripten",
-    "version": "%waterfall-lkgr%",
+    "id": "emscripten",
+    "version": "upstream-%waterfall-lkgr%",
     "bitness": 64,
     "linux_url": "https://storage.googleapis.com/wasm-llvm/builds/linux/%waterfall-lkgr%/wasm-binaries.tbz2",
     "install_path": "upstream/%waterfall-lkgr%",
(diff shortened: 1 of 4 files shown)


jgravelle-google:
Other details:
This change grabs emscripten and binaryen as well (it also grabs fastcomp, wabt, d8, and the spec interpreter, basically everything the waterfall builds is in that `.tbz2` archive) for synchronicity. For consistency they're modeled as separate tools.
We re-download the lkgr json from the waterfall when the user runs `update-tags`, for somewhat-consistency. The old lkgr build numbers aren't stored anywhere after being updated.

emsdk_manifest.json
```diff
@@ -1597,6 +1625,12 @@
     "version_filter": [
       ["%precompiled_tag64%", ">", "1.37.22"]
     ]
+  },
+  {
+    "version": "upstream-%waterfall-lkgr%",
+    "bitness": 64,
+    "uses": ["clang-upstream-%waterfall-lkgr%-64bit", "emscripten-upstream-%waterfall-lkgr%-64bit", "binaryen-upstream-%waterfall-lkgr%-64bit"],
+    "os": "linux"
```
juj: There is an implicit feature that "latest listed SDK is the newest stable version" that comes from https://github.com/juj/emsdk/blob/master/emsdk#L1463, so this added sdk section should probably be located after this line: https://github.com/jgravelle-google/emsdk/blob/f221912051bf1e7e92b38ac5655957c5f03af7f3/emsdk_manifest.json#L1219 so that `emsdk install latest` will not start downloading these waterfall builds instead of stable precompiled builds.

kripken: Hmm, we do `install latest` in the test here, and it still downloaded the proper one, this didn't seem to mess it up?

kripken: Yeah, the test here runs on linux on travis - which seems like what should have broken?

juj:
Looks great!

PR #165, opened 2018-07-19
LLVM_ROOT

juj:
This PR adds the `LLVM_ROOT` environment variable to also be registered (in addition to setting it in cfg) so that users can easily locate where clang executable resides.

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -1270,14 +1270,13 @@ class Tool:
   # Returns true if the system environment variables requires by this tool are currently active.
   def is_env_active(self):
     if hasattr(self, 'activated_env'):
-      (key, value) = parse_key_value(self.activated_environment())
-      if not key in os.environ or to_unix_path(os.environ[key]) != to_unix_path(value):
-        if VERBOSE: print(str(self) + ' is not active, because environment variable key="' + key + '" has value "' + str(os.getenv(key)) + '" but should have value "' + value + '"')
-        return False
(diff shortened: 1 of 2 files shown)


emsdk
```diff
@@ -1270,14 +1270,13 @@ class Tool:
...
     if hasattr(self, 'activated_env'):
-      (key, value) = parse_key_value(self.activated_environment())
-      if not key in os.environ or to_unix_path(os.environ[key]) != to_unix_path(value):
-        if VERBOSE: print(str(self) + ' is not active, because environment variable key="' + key + '" has value "' + str(os.getenv(key)) + '" but should have value "' + value + '"')
-        return False
-#      if WINDOWS:
-#        env_var = win_get_active_environment_variable(key)
-#        if env_var != value:
-#          return False
+      envs = self.activated_environment().split(';')
```
kripken: I don't really know this code, but it seems suspicious it checks `hasattr(self, 'activated_env')` and then uses `self.activated_environment()` (first `environment` is abbreviated to `env`, later it isn't). Is that a typo?

juj: Good point - the code there looks good, though the check there for `hasattr(self, 'activated_env')` is actually redundant, this check is also done inside `self.activated_environment()` function. Refactored this logic in the comment below to read more straightforward.

kripken:
Thanks!

Still not that familiar with this code overall, but the change looks good to me.

PR #168, opened 2018-08-16
Add --notty to force non-erasing progress bar

sbc100:
This is useful in CI environments that claim to be TTYs but
really don't have escape characters such as \r correctly.


juj:
Great!

PR #192, opened 2018-12-11
On Windows only read path data from registry if it's for a permanent …

iboB:
…change.

Otherwise don't nuke the user's current PATH environment variable.
Resolves juj#189


iboB:
Huh, why does this still say pending? I see it as complete here: https://travis-ci.org/juj/emsdk/builds/466688929

iboB:
@juj you should probably restart this build manually. Some bug in Travis is preventing it from finishing properly.

kripken:
Clicking on the travis build, it passed ok. Just the github UI has not updated, oddly. But the test seems fine.

PR #195, opened 2018-12-15
Enable "emsdk [install|activate] latest-upstream"

kripken:
This makes it possible to tell the emsdk to get "latest-upstream", which fetches the latest lkgr from there. This will probably be a common use pattern, I expect we may want to recommend users start trying out the wasm backend that way soon. This will also let us simplify this code: https://github.com/kripken/emscripten/blob/incoming/.circleci/config.yml#L334

Aside from adding the "latest-upstream" alias,  this PR has

 * A few minor cleanups in the emsdk code.
 * Minor restructuring of how we define the upstream stuff in the manifest: It seemed odd to have 3 things (clang, emscripten, binaryen) that are all coming from a single archive from the waterfall. Simpler to have just one - the archive is one big lump, there's no way to download just part of it.
 * Also add node 8.9.1 as a dependency of the upstream sdk, which makes things usable out of the box (node.js is the one thing not provided by the waterfall archive).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,6 @@ RUN cd /root/ \
  && /root/emsdk/emsdk activate latest \
  && source /root/emsdk/emsdk_env.sh --build=Release \
  && emcc hello_world.cpp \
- && rm ~/.emscripten \
  && /root/emsdk/emsdk update-tags \
  && /root/emsdk/emsdk install latest-upstream \
  && /root/emsdk/emsdk activate latest-upstream \
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,10 +9,7 @@ RUN cd /root/ \
  && echo "int main() {}" > hello_world.cpp \
  && apt-get update \
  && apt-get install -y python cmake build-essential openjdk-9-jre-headless \
- && /root/emsdk/emsdk install latest \
- && /root/emsdk/emsdk activate latest \
- && source /root/emsdk/emsdk_env.sh --build=Release \
- && emcc hello_world.cpp \
+ && /root/emsdk/emsdk update-tags \
  && /root/emsdk/emsdk install latest-upstream \
  && /root/emsdk/emsdk activate latest-upstream \
  && source /root/emsdk/emsdk_env.sh --build=Release \
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,10 +15,7 @@ RUN cd /root/ \
  && emcc hello_world.cpp \
  && /root/emsdk/emsdk update-tags \
  && /root/emsdk/emsdk install latest-upstream \
- && echo "activate!" \
  && /root/emsdk/emsdk activate latest-upstream \
- && echo "activated" \
  && source /root/emsdk/emsdk_env.sh --build=Release \
- && find -name "emcc.py" \
- && emcc hello_world.cpp \
+ && emcc hello_world.cpp
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,5 @@ RUN cd /root/ \
  && /root/emsdk/emsdk install latest-upstream \
  && /root/emsdk/emsdk activate latest-upstream \
  && source /root/emsdk/emsdk_env.sh --build=Release \
- && emcc hello_world.cpp -o a.bc \
- && export EMCC_DEBUG=1 \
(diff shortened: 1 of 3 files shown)


kripken:
cc @juj @jgravelle-google

emsdk_manifest.json
```diff
@@ -191,32 +191,13 @@
...
-  },
-  {
-    "id": "binaryen",
-    "version": "upstream-%waterfall-lkgr%",
-    "bitness": 64,
-    "linux_url": "https://storage.googleapis.com/wasm-llvm/builds/linux/%waterfall-lkgr%/wasm-binaries.tbz2",
-    "install_path": "upstream/%waterfall-lkgr%",
-    "activated_path": "%installation_dir%",
-    "activated_cfg": "BINARYEN_ROOT='%installation_dir%'",
-    "activated_env": "BINARYEN_ROOT=%installation_dir%"
```
juj: Are these package deletions intentional?

juj: Oh gotcha, noticed the comment about the lumped package - lgtm.

juj:
lgtm

Dockerfile
```diff
@@ -13,7 +13,6 @@ RUN cd /root/ \
  && /root/emsdk/emsdk activate latest \
  && source /root/emsdk/emsdk_env.sh --build=Release \
  && emcc hello_world.cpp \
- && rm ~/.emscripten \
```
juj: Is the intent of this Dockerfile to only have `latest-upstream` active at the end of this whole script, and nothing else? If so, then the line `&& /root/emsdk/emsdk activate latest \` looks redundant (and the `emsdk_env` line and compilation check should probably move down to after `activate latest-upstream` has been applied?)

kripken: This checks both paths - it compiles hello world with fastcomp, then with the wasm backend. So it's also checking switching between them.

emsdk
```diff
@@ -1697,10 +1707,13 @@ def download_waterfall_lkgr():
...
 def load_waterfall_lkgr():
   try:
-    data = json.loads(open(os.path.join('upstream', 'lkgr.json'), 'r').read())
+    text = open(sdk_path(os.path.join('upstream', 'lkgr.json')), 'r').read()
+    data = json.loads(text)
     lkgr = data['build']
     return [lkgr]
-  except:
+  except Exception as e:
+    print('Error parsing lkgr.json!')
```
jgravelle-google: Yay better debug messages

Although I think what I was thinking here was that a user may not have fetched the waterfall lkgr ever, say they upgraded emsdk from before it was included, and I think this code gets hit when invoking `emsdk list`?

Anyway this is harmless enough, and in the future will be the default, so this is still a good change. Was trying to dump my context of "why did I ever make this fail silently?"

emsdk
```diff
@@ -2210,6 +2223,8 @@ def main():
         sys.argv[i] = str(find_latest_nightly_32bit_sdk())
       elif sys.argv[i] == 'sdk-nightly-latest-64bit':
         sys.argv[i] = str(find_latest_nightly_64bit_sdk())
+      elif sys.argv[i] == 'latest-upstream' or sys.argv[i] == 'latest-clang-upstream':
```
jgravelle-google: Should we update readme.md to let people know this exists, or wait until that's less-beta? Should probably be done in a follow-up, if we want to document it reasonably well (e.g. why might one want to try the wasm backend today)

kripken: I'd prefer to wait until it's less beta - we have a chance to make a good first impression of the wasm backend path.

emsdk
juj: btw, what is lkgr short of?

juj: Hmm indeed it looks like `upstream/lkgr.json` might not exist? Perhaps `if not os.path.isfile(os.path.join('upstream', 'lkgr.json'): return []` ? (There is a `VERBOSE` variable top of the file if you want to do some logging only in a verbose/debug mode in this case)

kripken: I think it's "last known good revision".

Yes, it looks like the file might not exist if tags have not been updated. We could print "you should run update-tags" in that case, however, looks like that would be printed during the startup of the emsdk invocation that you run update-tags with, unless we added some more logic, not sure offhand how simple that would be.

PR #198, opened 2018-12-20
Deduplicate activated configs in the .emscripten file

### LINKED ISSUE #194: activate influenced by existing .emscripten?
I'm trying to get the upstream-waterfall option more usable. One issue I can into is if you
 * install a regular version.
 * activate it
 * install an upstream-waterfall version
 * activate that
then the generated .emscripten is not as I'd expect. It seems to contain things from the previous activation? As a workaround in my testing I am deleting the `.emscripten` in the middle, which seems to work, but is this expected behavior?

kripken:
While doing so, we:

* keep the latest activation (e.g., the user may have activated `latest` and then `latest-upstream`, then the upstream LLVM is what is desired).
* keep the order of keys fixed (so the relative order of lines in the `.emscripten` file is fixed)

This adds some assertions in the Dockerfile, to verify we have one `LLVM_ROOT` command, and it is the right one. Probably we should put tests in a nicer place (file, directory) - where would you prefer?

Fixes #194

cc @juj @jgravelle-google

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -1052,13 +1052,26 @@ def generate_dot_emscripten(active_tools):
   if embedded:
     cfg += "emsdk_path=os.path.dirname(os.environ.get('EM_CONFIG')).replace('\\\\', '/')\n"
 
+  # Different tools may provide the same activated configs; the latest to be
+  # activated is the relevant one.
+  final_activated_names = []
+  final_activated_values = {}
+
   for tool in active_tools:
     tool_cfg = tool.activated_config()
     if tool_cfg:
-      tool_cfg = tool_cfg.replace(';', '\n')
-      if 'SPIDERMONKEY_ENGINE=' in tool_cfg: has_spidermonkey = True
-      if 'NODE_JS=' in tool_cfg: has_node = True
-      cfg += tool_cfg + '\n'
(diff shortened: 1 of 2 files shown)


emsdk
```diff
@@ -1052,13 +1052,26 @@ def generate_dot_emscripten(active_tools):
...
+        if name not in final_activated_values:
+          final_activated_names.append(name)
+        final_activated_values[name] = value
+
+  for name in final_activated_names:
+    if name == 'SPIDERMONKEY_ENGINE':
+      has_spidermonkey = True
+    if name == 'NODE_JS':
+      has_node = True
+    cfg += name + ' = ' + final_activated_values[name] + '\n'
```
juj: A bit surprised that this is needed, but looks good, certainly won't harm. It looks like the variable `final_activated_names` is being used explicitly for the purpose of preserving an ordered list, would be good to add a comment about that - or perhaps rename `final_activated_names` -> `activated_keys_in_order` and `final_activated_values` -> `activated_key_values` or similar?

kripken: Sounds good, renamed as suggested.

juj:
Ops, sorry, missed that this was not merged. Looks good to go, although there is now a conflict in this PR.

PR #200, opened 2019-01-04
Add 'latest-fastcomp' sdk

kripken:
With this, we can do `emsdk install latest-fastcomp` and it installs fastcomp from the waterfall. That is, we then have 3 main sdks people might want to use:

 * `latest` which installs **fastcomp**-llvm (plus emscripten etc.) from the **mozilla** infrastructure. (fetches the last emscripten version there)
 * `latest-upstream` which installs **upstream**-llvm (plus emscripten etc.) from the **waterfall** infrastructure. (fetches the last known good revision (lkgr) there)
 * `latest-fastcomp` which installs **fastcomp**-llvm (plus emscripten etc.) from the **waterfall** infrastructure. (fetches the last known good revision (lkgr) there)

The first and last are currently somewhat overlapping in that both fetch a build of fastcomp. However, as we transition away from the mozilla infrastructure, we could just make `latest` an alias for `latest-fastcomp`. (And later, when we're ready to switch to the wasm backend by default, the alias could switch to `latest-upstream`.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,11 +28,19 @@ RUN cd /root/ \
  && emcc hello_world.cpp -s WASM=0 -o d.out.js \
  && echo "text executions" \
  && /root/emsdk/node/8.9.1_64bit/bin/node a.out.js &> o \
+ && ls -al o \
  && cat o \
+ && echo \
  && /root/emsdk/node/8.9.1_64bit/bin/node b.out.js &> o \
+ && ls -al o \
  && cat o \
+ && echo \
  && /root/emsdk/node/8.9.1_64bit/bin/node c.out.js &> o \
+ && ls -al o \
  && cat o \
+ && echo \
  && /root/emsdk/node/8.9.1_64bit/bin/node d.out.js &> o \
- && cat o
+ && ls -al o \
+ && cat o \
+ && echo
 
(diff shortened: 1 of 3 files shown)


kripken:
cc @juj @dschuff @jgravelle-google @sbc100

juj:
lgtm

kripken:
Some problems with this landing:

* @juj I meant to use the first commit as the message here, but you squashed and got the commit messages ("fix", "fix" etc.). I guess no big deal though.
* this broke the ec2 mozilla machines, http://ec2-54-213-252-128.us-west-2.compute.amazonaws.com:8112/#/builders/11/builds/659
```
pdating cef8e88..c4c6b03
Fast-forward
 bin/deploy_emscripten_llvm.py | 11 +----------
 1 file changed, 1 insertion(+), 10 deletions(-)
From https://github.com/juj/emsdk
   df648b4..b4de632  master     -> origin/master
error: Your local changes to the following files would be overwritten by merge:
	upstream/lkgr.json
Please, commit your changes or stash them before you can merge.
Aborting
```
Should we not update `upstream/lkgr.json`? Or is the bug on the ec2 machines? Not sure if we can fix things there though.

kripken:
Oh, it was actually #201 that broke that thing. I was confused by 2 PRs landing.

PR #201, opened 2019-01-06
Commit update-tags results

kripken:
@juj I'm not sure how this updating should happen - are we supposed to commit it to the repo every time there's a new tag? Alternatively, should `emsdk update` also run `emsdk update-tags` perhaps, to avoid that?


juj:
The tags have been committed from time to time for convenience to user, so they get the latest tag out of the box. `emsdk update` does also run `emsdk update-tags`, so they do get latest tags when updating (but still need to update). First time run of emsdk will do `emsdk update` to get the precompiled registry, so the download will automatically happen, basically updating to newer tags just makes the repository show more recent state from out of the box.

PR #205, opened 2019-01-08
Update emscripten and binaryen tags

kripken:
This does *not* update the lkgr.json file, since it broke the ec2 mozilla bots, see https://github.com/juj/emsdk/pull/200#issuecomment-452456523

(Is the bug on the bots, though?)


juj:
Oh, what happened there is that there was code in `emsdk` to download the `lkgr.json` file already before the update, so the bots would have a untracked copy of the file in the local filesystem. Then after the file was added to the repository git prevents one from updating to the newer branch when that would mean it would have to override the file.

Do you have a plan to recreate an EC2-like setup at Google for baking Emsdk builds? Or what is your thinking at the moment?

One way to get the current EC2 bots working may be to update emslave code so that their scripts do a forced git checkout that overrides such local file vs tracked file conflicts, and forces the file to be replaced. Although that might be of temporary help - still the current EC2 bots are a bit of a time bomb.

PR #207, opened 2019-01-09
Remove wasm.js reference

kripken:
We removed it in Binaryen #1858. The reference here broke source installs using the emsdk.

Fixes https://github.com/juj/emsdk/issues/206


kripken:
Merging quickly since this fixes serious current breakage (and includes a test for that, which verifies we fix it).

PR #213, opened 2019-02-05
sdk-wasm-master

juj:
Add SDK definitions for upstream wasm backend enabled LLVM, "sdk-wasm-master-32/64bit".

With this, one can get going with the upstream wasm backend with

```
git clone https://github.com/emscripten-core/emsdk.git
cd emsdk
emsdk install sdk-wasm-master-64bit # Compiles llvm-mirror/llvm master branch from source
emsdk activate sdk-wasm-master-64bit
source ./emsdk_env.sh
cd emscripten/incoming
em++ tests/hello_world.c -o a.js
node a.js # runs a.wasm produced with upstream wasm backend
```

I was a bit surprised we did not have something like this already, what is the workflow that you guys have been using? @kripken @dschuff @jgravelle-google ?

Compare the above with using the traditional fastcomp based backend, which is used with:

```
git clone https://github.com/emscripten-core/emsdk.git
cd emsdk
emsdk install sdk-incoming-64bit  # Compiles emscripten-fastcomp incoming branch from source
emsdk activate sdk-incoming-64bit
source ./emsdk_env.sh
cd emscripten/incoming
em++ tests/hello_world.c -o a.js
node a.js # runs a.wasm produced with fastcomp backend
```

so the two usages work well in parallel, one can then toggle between the two by flipping `emsdk activate sdk-wasm-master-64bit` vs `emsdk activate sdk-incoming-64bit`.

### DIFF
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -1078,9 +1078,21 @@
     "os": "win"
   },
   {
+    "version": "wasm-master",
+    "bitness": 32,
+    "uses": ["upstream-clang-master-32bit", "node-8.9.1-32bit", "python-2.7.13.1-32bit", "java-8.152-32bit", "emscripten-incoming-32bit", "binaryen-master-32bit"],
+    "os": "win"
+  },
+  {
     "version": "incoming",
     "bitness": 64,
     "uses": ["clang-incoming-64bit", "node-8.9.1-64bit", "python-2.7.13.1-64bit", "java-8.152-64bit", "emscripten-incoming-64bit", "binaryen-master-64bit"],
+    "os": "win"
+  },
+  {
+    "version": "wasm-master",
+    "bitness": 64,
+    "uses": ["upstream-clang-master-64bit", "node-8.9.1-64bit", "python-2.7.13.1-64bit", "java-8.152-64bit", "emscripten-incoming-64bit", "binaryen-master-64bit"],
     "os": "win"
   },
   {
@@ -1090,15 +1102,33 @@
     "os": "osx"
   },
   {
+    "version": "wasm-master",
+    "bitness": 64,
+    "uses": ["upstream-clang-master-64bit", "node-8.9.1-64bit", "emscripten-incoming-64bit", "binaryen-master-64bit"],
+    "os": "osx"
+  },
+  {
     "version": "incoming",
     "bitness": 32,
     "uses": ["clang-incoming-32bit", "node-8.9.1-32bit", "emscripten-incoming-32bit", "binaryen-master-32bit"],
     "os": "linux"
   },
   {
+    "version": "wasm-master",
+    "bitness": 32,
+    "uses": ["upstream-clang-master-32bit", "node-8.9.1-32bit", "emscripten-incoming-32bit", "binaryen-master-32bit"],
+    "os": "linux"
+  },
+  {
     "version": "incoming",
     "bitness": 64,
     "uses": ["clang-incoming-64bit", "node-8.9.1-64bit", "emscripten-incoming-64bit", "binaryen-master-64bit"],
+    "os": "linux"
+  },
+  {
+    "version": "wasm-master",
+    "bitness": 64,
+    "uses": ["upstream-clang-master-64bit", "node-8.9.1-64bit", "emscripten-incoming-64bit", "binaryen-master-64bit"],
     "os": "linux"
   },
   {


jgravelle-google:
Personally I've been using the wasm waterfall (https://github.com/WebAssembly/waterfall) to sync all the things for local development.
For users I've been recommending they use the `latest-upstream` prebuilt versions.

kripken:
`wasm-master` may be odd since it uses emscripten incoming, and not master?

Also, just `wasm` doesn't differentiate between asm2wasm and the upstream backend, so perhaps `upstream`?

This is basically a source build of `latest-upstream`, I believe, so perhaps something like `latest-upstream-src`, or does the emsdk have another way to differentiate a source from a binary build?

juj:
> `wasm-master` may be odd since it uses emscripten incoming, and not master?

`wasm-incoming` would be odd since it uses LLVM master, and not `incoming`.

> Also, just `wasm` doesn't differentiate between asm2wasm and the upstream backend, so perhaps `upstream`?

The thought was that this would be somehow "official" sounding going forward, I did not mind too much with this name that fastcomp+asm2wasm can also produce wasm.

> This is basically a source build of `latest-upstream`, I believe, so perhaps something like `latest-upstream-src`, or does the emsdk have another way to differentiate a source from a binary build?

For fastcomp emsdk used `sdk-x.yy.z` and `sdk-nightly-yyyymmdd` to represent binary builds, and `sdk-tag-x.yy.z` and `sdk-(branchname)` to represent source builds. (+ `-32bit`/`-64bit` suffix on each). In emsdk sdk targets always start with prefix `sdk-`, so e.g. `sdk-whatever-64bit`.

This target gets upstream wasm llvm git master branch + emscripten git incoming branch + builds 64-bit, so if `sdk-wasm-master-64bit` does not feel good, then perhaps one of

- `sdk-upstream-master-64bit`,
- `sdk-upstream-incoming-64bit`,
- `sdk-upstream-64bit`
- `sdk-llvm-master-64bit`
- `sdk-llvm-incoming-64bit`
- `sdk-llvm-64bit`
- `sdk-wasm-64bit`
- `sdk-llvm-master-emscripten-incoming-64bit`

I think any name would be fine for me, though perhaps probably not the last one, since that would be quite tedious to type. Having to type `sdk-latest-upstream-src-64bit` would be a bit tedious too.

For end users who install from precompiled build, emsdk has had the meta-target `latest`, i.e. `emsdk install latest` or `emsdk install latest-32bit` to get the latest precompiled tagged zip. Perhaps a corresponding `latest-wasm` could be introduced at some point. (or `latest` becomes wasm, and `latest-asmjs` would get fastcomp)

kripken:
I don't feel strongly. But "latest" implies master or incoming etc. depending on the repo, so it avoids hardcoding master or incoming in the text.

Maybe

 * `latest-upstream`, `latest-wasm`, etc. mean latest wasm backend using master/incoming/etc. from binary
 * `latest-upstream-src`, `latest-wasm-src` would mean to build from source

? I realize that is not consistent with past names, but just adding `-src` to mean "build from source" seems simple and not surprising?

juj:
Oh I realize this never landed.

Thinking more about the names, I find that it's perhaps best not to use the meta-name `latest` for wasm backend until we switch to wasm backend by default. That is, instead of having two meta-targets `latest` and `latest-wasm` or similar, or using a part of the name `latest` in the wasm SDK, we'd just keep the one `latest` meta-target, and when the switch occurs, that would just start pointing to the wasm-based LLVM backend instead of the fastcomp-based LLVM backend. So I still think using either the current name `sdk-wasm-master-64bit` (following LLVM repository branch name), or `sdk-wasm-incoming-64bit` (following Emscripten repository branch name) would be best here. What do you think @kripken?

juj:
> "latest" implies master or incoming etc. depending on the repo, so it avoids hardcoding master or incoming in the text.

Currently `latest` in emsdk is a meta-target that refers to the latest official stable release precompiled from a tag, so I would not like to confuse the SDK name with that.

kripken:
> Thinking more about the names, I find that it's perhaps best not to use the meta-name latest for wasm backend until we switch to wasm backend by default. 

Sounds good. But before we switch, we have `latest-upstream` already, and after, we'll need a way for them to get the non-wasm backend (for cases where they need the old one), I think?

juj:
Ah right, yeah - there's now `latest` and `latest-upstream`. I think in the future those will just be aliases to the same thing. I'll land this as-is now, merging `latest` and `latest-upstream` can then occur in the future.

PR #214, opened 2019-02-07
Support Fish shell (emsdk_env)

InBetweenNames:
Fish shell doesn't support the same syntax as most shells,
so add an emsdk_env.fish to support these users.
Instructions included in `emsdk_env.fish`.

### DIFF
--- a/emsdk_env.fish
+++ b/emsdk_env.fish
@@ -1,0 +1,17 @@
+#In your Fish configuration, add this line:
+#alias emsdk_setup ". /path/to/emsdk/emsdk_env.fish"
+#Now, when you want to use the SDK, run this alias first to set up
+#your environment.
+
+set -l script (status -f)
+set -l dir (dirname $script)
+
+pushd $dir > /dev/null
+
+./emsdk construct_env "$argv"
+. ./emsdk_set_env.sh
+
+set -e -l script
+set -e -l dir
+
+popd > /dev/null


kripken:
We simplified some things just now - is it still incompatible with fish?

If not, is there a way to simplify it more so it is compatible (a single script would be nice)?

InBetweenNames:
Unfortunately, Fish breaks compatibility with sh and hence has it's own syntax for nearly everything.  The only remotely portable way of doing it were if the `emsdk_env.sh` file contained only the following:

~~~
./emsdk construct_env "$@"
. ./emsdk_set_env.sh
~~~

Even then, Fish wants `$argv` instead of `$@`.  Not a great situation, I know.  I agree a single script would be nice -- I'm just wondering how we can achieve that.

kripken:
I see, thanks.

Seems like there's not much harm to adding this file here, except that we should probably only do that if we can test it.

InBetweenNames:
I would be willing to put forth the effort to implement such a test -- could you point me in the direction of what is usually tested for these env scripts?  I notice there's powershell, batch and SH already there, so I figure I could just take one of those tests and modify it accordingly for fish.

kripken:
The current tests are in `Dockerfile` - it basically gets the emsdk and verifies it can build hello world.

InBetweenNames:
Thanks -- it seems reasonable that testing hello world should suffice.  There's just one complication.
From `Dockerfile`:

~~~
...
...
~~~

It seems that `Dockerfile` itself uses Bash and just sources the `emsdk_env.sh`.  I'd have to write in an entirely separate test using `fish` with virtually the same content.  How are the Powershell and Batch  `emsdk_env` versions tested?  Perhaps I could get some inspiration from that.

InBetweenNames:
I agree it would be good to improve the testing framework to cover more than just Bash.  Since I'm not really affiliated with Emscripten or emsdk, I'm wondering how am I supposed to do that?  I understand you probably have many development policies that aren't in written form, and anything intrusive like that would no doubt step on some toes.  I really just wanted to commit this for convenience for myself and other Fish users -- I didn't expect to be getting into rewriting testing frameworks when I opened this PR.

kripken:
Yeah, maybe you're right - it's not easy to add testing of these things, and adding the file doesn't cause any harm. If we get confused bug reports from users, we may want to reconsider that, but for now lets merge this in. Thanks!

PR #223, opened 2019-02-24
Add the lkgr as a prefix to upstream downloads

kripken:
Without this, we would not download new versions, as the filename looked the same ("already downloaded"). We'd unzip those old contents to a directory with the new name, giving the impression we updated when we didn't :(

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -1395,7 +1395,8 @@ class Tool:
         success = git_clone_checkout_and_pull(url, self.installation_path(), self.git_branch)
       elif url.endswith('zip') or url.endswith('.tar') or url.endswith('.gz') or url.endswith('.xz') or url.endswith('.tbz2'):
         download_even_if_exists = (self.id == 'vs-tool')
-        success = download_and_unzip(url, self.installation_path(), download_even_if_exists, filename_prefix=self.zipfile_prefix if hasattr(self, 'zipfile_prefix') else '')
+        filename_prefix = self.zipfile_prefix if hasattr(self, 'zipfile_prefix') else ''
+        success = download_and_unzip(url, self.installation_path(), download_even_if_exists, filename_prefix=filename_prefix)
       else:
         dst_file = download_file(urljoin(emsdk_packages_url, self.download_url()), self.installation_path())
         if dst_file: success = True
--- a/emsdk
+++ b/emsdk
@@ -1395,7 +1395,7 @@ class Tool:
         success = git_clone_checkout_and_pull(url, self.installation_path(), self.git_branch)
       elif url.endswith('zip') or url.endswith('.tar') or url.endswith('.gz') or url.endswith('.xz') or url.endswith('.tbz2'):
         download_even_if_exists = (self.id == 'vs-tool')
-        filename_prefix = self.zipfile_prefix if hasattr(self, 'zipfile_prefix') else ''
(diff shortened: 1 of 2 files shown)


kripken:
(Just fixed the emsdk CI, which apparently was broken since we migrated to emscripten-core. Seems to start ok now.)

emsdk
```diff
@@ -1395,7 +1395,8 @@ class Tool:
         success = git_clone_checkout_and_pull(url, self.installation_path(), self.git_branch)
       elif url.endswith('zip') or url.endswith('.tar') or url.endswith('.gz') or url.endswith('.xz') or url.endswith('.tbz2'):
         download_even_if_exists = (self.id == 'vs-tool')
-        success = download_and_unzip(url, self.installation_path(), download_even_if_exists, filename_prefix=self.zipfile_prefix if hasattr(self, 'zipfile_prefix') else '')
+        filename_prefix = self.zipfile_prefix if hasattr(self, 'zipfile_prefix') else ''
+        success = download_and_unzip(url, self.installation_path(), download_even_if_exists=download_even_if_exists, filename_prefix=filename_prefix)
```
jgravelle-google: Nice cleanup

emsdk
```diff
@@ -1395,7 +1395,8 @@ class Tool:
         success = git_clone_checkout_and_pull(url, self.installation_path(), self.git_branch)
       elif url.endswith('zip') or url.endswith('.tar') or url.endswith('.gz') or url.endswith('.xz') or url.endswith('.tbz2'):
         download_even_if_exists = (self.id == 'vs-tool')
-        success = download_and_unzip(url, self.installation_path(), download_even_if_exists, filename_prefix=self.zipfile_prefix if hasattr(self, 'zipfile_prefix') else '')
+        filename_prefix = self.zipfile_prefix if hasattr(self, 'zipfile_prefix') else ''
```
sbc100: I think this can be more simple written as:
```
filename_prefix = getattr(self, 'zipfile_prefix', '')
```

jgravelle-google: I did not know `getattr` had a default field! TIL

I was thinking earlier that I'd personally prefer
```python
filename_prefix = ''
if hasattr(self, 'zipfile_prefix'):
  filename_prefix = self.zipfile_prefix
```
but that's just me not liking python's `B if A else C` way of reading an `A ? B : C` so I didn't mention it (also because it was essentially moving existing code)
But getattr default is better still, so yes please

kripken: Nice, thanks, updated.

sbc100:
Hmm, seems like we will end up with zip files called "1234-wasm-binaries.tbz2" which is a little confusing.

How about we rename this to `zipfile_name`.

The three existing uses can all be changed (e.g. `emscripten-e%tag`) and we can use `wasm-waterfall-binaries-%waterfall-lkgr%`?

Also, it looks like there should be 4th existing use since "emscripten" "64" seems to be missing this attribute.

kripken:
Sounds reasonable to me. @juj, what do you think?

sbc100:
I can do as a followup if you prefer?  I've been meaning to take a look at some of the emsdk stuff anyway.

kripken:
Followup sounds good. Then we can merge this now and unbreak people that might hit the bug.

PR #224, opened 2019-02-26
Fix some flake8 warnings in emsdk

sbc100:
Only some checks are enabled here in order to make this change
easier to review and less error prone.


sbc100:
@juj any other feedback?

sbc100:
Any other comments?  If not I'll land later today.

PR #220, opened 2019-02-15
Remove bashism in emsdk_env.sh in favor of portable alternative

nic-hartley:
Fixes #219 

I also replaced the backticks with `$()` because they're the modern POSIX standard and look moderately prettier. I'm fine changing it back.

PR #227, opened 2019-03-06
Fix bug introduced in #220

Coder-256:
It seems that some variables were mistakenly left unquoted, leading to paths containing spaces being treated incorrectly.

For example, the directory `/home/person/foo bar/test/` would cause `CURDIR=$(pwd)` to be expanded to `CURDIR=/home/person/foo bar/test/`. This means that the shell would attempt to run the command `bar/test/` with the environment variable `CURDIR` set to `/home/person/foo`. This is due to [word splitting](https://www.gnu.org/software/bash/manual/html_node/Word-Splitting.html). TL;DR the result of shell expansions should probably be quoted 99% of the time to be extra safe/compatible with spaces. (I say 99% because there are rare exceptions such as `"$@"` and running the contents of a variable)

### DIFF
--- a/emsdk_env.sh
+++ b/emsdk_env.sh
@@ -19,11 +19,11 @@ SRC="$BASH_SOURCE"
 if [ "$SRC" = "" ]; then
   SRC="$0"
 fi
-CURDIR=$(pwd)
-cd $(dirname "$SRC")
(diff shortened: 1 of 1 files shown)


kripken:
Thanks @Coder-256!

PR #238, opened 2019-05-20
Remove lkgr.json from git - the mozilla bots break if it's there

kripken:
E.g.

http://ec2-54-213-252-128.us-west-2.compute.amazonaws.com:8112/#/builders/11/builds/1096

The bots should probably work around this, but as we plan to replace them soon, let's do this for now which is the simplest solution.


kripken:
Merging to try to unbreak the CI.

PR #239, opened 2019-05-23
Fix a bunch of flake8 warnings

sbc100:
Now that we have much better testing in CI this kind of change
should be much easier to make with confidence.


sbc100:
Updated to be wider set of flake8 fixes

PR #243, opened 2019-05-23
Add an 'emscripten_version' property of an installed tool

kripken:
If set, it makes us emit an emscripten-version.txt file with those contexts. This lets the emsdk set the version in emscripten properly, see emscripten-core/emscripten#8660


kripken:
The test here can't pass until the emscripten part lands.

kripken:
Rethought this a little. It now:

 * Structures `emscripten-releases-tags.txt`  to mention the latest release, and then a list of all previous releases.
 * Adds a new "release" for 1.38.33. If this all works, we can let our users try this out!
 * If the manifest has `emscripten_releases_hash`, then we write out emscripten-version.txt with the proper version. We start with the hash, and check if there is a released version for it, if so we emit that, otherwise we emit the hash.

kripken:
Also includes a fix for zipfile handling on windows that @dschuff and I ran into while running this.

PR #247, opened 2019-05-24
Add a show-activated-paths option, which prints out the activated paths

kripken:
This came up as possibly useful for emsdk users that want to set up their own .emscripten file, so they need to know which dir the code is under. With the emscripten-releases system, currently each download gets its own directory, so it changes over time. (It's a new dir each time to make it easy to switch between builds back and forth, and also it lets us not unzip a download that was already downloaded.)

Example output:
```
$./emsdk show-activated-paths
/home/.../sdk/upstream/c7784483727437237472347/emscripten
/home/../sdk/node/../bin
```

By default the emsdk downloads node.js as well, so there are two activated tools, and so two paths.

Alternatively, perhaps we can fix this by reusing the same directory.  That might be better, come to think of it, as otherwise each update creates a new one and they aren't small, so if a user isn't careful over a year there can be a lot of disk usage that accumulates?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,8 +50,8 @@ RUN cd /root/ \
  && source /root/emsdk/emsdk_env.sh --build=Release \
  && emcc hello_world.cpp \
  && echo "test show-activated-paths" \
- && ./emsdk show-activated-paths | grep emscripten > a.tmp \
- && ./emsdk show-activated-paths tot-fastcomp | grep emscripten > b.tmp \
+ && /root/emsdk/emsdk show-activated-paths | grep emscripten > a.tmp \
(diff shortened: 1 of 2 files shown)


kripken:
We probably don't need this since #248 landed. However, maybe a tool like this is still useful, given the other stuff discussed in there like how to find the various paths? cc @floooh

floooh:
...see my reply here: https://github.com/emscripten-core/emsdk/pull/248#issuecomment-496856784

sbc100:
I suppose this is not needed at this point?

sbc100:
@kripken can you delete the branch if you agree?

PR #248, opened 2019-05-24
If an archive has not been downloaded, download it and assume its contents are new.

kripken:
This allows us to always unpack our builds into `upstream/` or `fastcomp/`, instead of creating a new directory each time. Since without this, we'd get told to download a new file, and the emsdk would see that where it would be unpacked had contents, and assumed that even though it wasn't downloaded it must be the same. So we'd silently skip it.

It's useful to always unpack into the same dir since it's easier for people that create their own .emscripten file (the location of emscripten etc. is all fixed under the emsdk), and also it avoids directories accumulating, which each take hundreds of megabytes, so over time it can get burdensome.

An alternative to #247.

cc @floooh

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -45,6 +45,8 @@ emscripten_releases_repo = 'https://chromium.googlesource.com/emscripten-release
 
 emscripten_releases_download_url_template = "https://storage.googleapis.com/webassembly/emscripten-releases-builds/%s/%s/wasm-binaries.%s"
 
+zips_subdir = 'zips/'
+
 # Enable this to do very verbose printing about the different steps that are being run. Useful for debugging.
 VERBOSE = bool(os.getenv('EMSDK_VERBOSE')) if os.getenv('EMSDK_VERBOSE') is not None else False
 TTY_OUTPUT = sys.stdout.isatty()
@@ -534,11 +536,7 @@ def get_content_length(download):
 
   return 0
 
-
-# On success, returns the filename on the disk pointing to the destination file that was produced
-# On failure, returns None.
-def download_file(url, dstpath, download_even_if_exists=False, filename_prefix=''):
-  if VERBOSE: print('download_file(url=' + url + ', dstpath=' + dstpath + ')')
+def get_download_target(url, dstpath, filename_prefix=''):
   file_name = filename_prefix + url.split('/')[-1]
   if path_points_to_directory(dstpath):
     file_name = os.path.join(dstpath, file_name)
@@ -547,6 +545,14 @@ def download_file(url, dstpath, download_even_if_exists=False, filename_prefix='
 
   # Treat all relative destination paths as relative to the SDK root directory, not the current working directory.
   file_name = sdk_path(file_name)
+
+  return file_name
+
+# On success, returns the filename on the disk pointing to the destination file that was produced
+# On failure, returns None.
+def download_file(url, dstpath, download_even_if_exists=False, filename_prefix=''):
+  if VERBOSE: print('download_file(url=' + url + ', dstpath=' + dstpath + ')')
+  file_name = get_download_target(url, dstpath, filename_prefix)
 
   if os.path.exists(file_name) and not download_even_if_exists:
     print("File '" + file_name + "' already downloaded, skipping.")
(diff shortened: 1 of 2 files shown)


kripken:
I lean towards this option myself, over #247.

floooh:
Some background info from my side, there's 2 problems:

- A cmake toolchain file needs to know where the emscripten toolchain is located.
- I want to have a "workspace-local" emscripten SDK installation, and several emscripten SDKs side by side on the same system (this works fine with the --embedded flag, once I know where emcc is.

Currently I'm "faking" it by hardwiring workspace-local path in my cmake toolchain file to a subdirectory "emscripten/incoming" in the emsdk directory. This breaks with the new "obfuscated" path.

I had a similar situation recently with macOS/iOS SDKs, cmake somehow lost the ability to find the active SDK's sysroot in recent macOS versions. What I'm doing there now is query the path to the active SDK myself in my cmake wrapper and give this to cmake as a config variable over the command line.

I'm going to clean up my emscripten SDK integration the same way now. I know where the emsdk is located, and from here on it's two options:

- parse the .emscripten file (in my case the right one is relative to the emsdk directory), and extract the paths from there, this should work already both for the "old" emsdk with predictable paths, and the "new" emsdk with the obfuscated paths, but it's a bit awkward (for me it's easy since I'm using python, but if other people want to do this with cmake scripting only it'll become quite messy)

- a better option would be to offer a mechanism to query the relevant emscripten SDK paths via the emsdk script (since emsdk should have all that knowledge), for instance ```./emsdk show-emsdk-path```, ```./emsdk show-emscripten-root```, ```./emsdk show-node-js``` etc... this would be similar to how xcrun works on Mac to query the active SDK (```xcrun --show-sdk-path```), the output should be an absolute path and nothing else, so that it's easy to process the output from scripts.

...but anyway, I think it's good to have "human readable" and "stable" paths in an emscripten SDK install, this will make life easier for people with exotic build system setup :)

floooh:
...one thing I noticed while working on my new emscripten SDK integration: the new .emscripten file no longer has an item which points to the directory where the emcc scripts are located (formerly there was an ```EMSCRIPTEN_ROOT``` item).

I need to tell cmake where the various toolchain binaries are (emcc, em++, emar and emranlib), is the idea now that I should use ```${LLVM_ROOT}/bin``` for this?

Also, I stumbled again over the fact that ```.emscripten``` is actually a python file which must be run through the python interpreter to make any sense (I guess this is required so that invoking emcc with --em-config continues to work)...

...but thinking about this I wonder whether such a dynamic evaluation is even required, because:

The --em-config arg already takes the name of a specific .emscripten file. This .emscripten file has been created either by an ```./emsdk activate ...``` for a "global" SDK installation or ```./emsdk activate --embedded ...``` for a "local" side-by-side installation.

So there's one .emscripten file per SDK installation anyway... why does the content of this emscripten file need to be dynamically evaluated with an incoming EM_CONFIG environment variable?

...it's a different topic, and probably deserved a different ticket, it just occurred to me for the first time that this is sort of strange ;)

floooh:
...another cmake-related note:

The cmake toolchain file in cmake/Modules/Platform expects a common "EMSCRIPTEN_ROOT_PATH" where both the compiler tools (emcc, em++, ...) and the system includes (under ${EMSCRIPTEN_ROOT_PATH}/system/include).

With the "new" .emscripten file it's hard to extract this location, the old EMSCRIPTEN_ROOT item is gone, and the LLVM_ROOT item points to a .../bin directory, which has the emcc, em++ etc tools, but not the system/include directory.

There's currently no item in .emscripten which points to this EMSCRIPTEN_ROOT_PATH location expected by the cmake toolchain file. The closest would be ```${BINARYEN_ROOT}/emscripten```, but this seems like a hack :)

Just wanted to mention this, I'll try to work around with some special cases for "old" vs "new" SDK directory structure.

floooh:
...ok I it generally working, but I think getting the paths for the cmake toolchain file by "parsing" the right .emscripten file is too much hassle, especially since the file needs to be evaluated by a python interpreter.

I could reduce the information that needs to be handed to cmake via command line options to 2 locations:

- the emsdk directory itself, because that's where the .emscripten file and .emscripten_cache directory is located when doing ```./emsdk activate --embedded xxx```, these are needed for the ```--em-config``` and ```--cache``` options passed to emcc/em++
- the emscripten SDK "root" directory, this is where the emcc, em++, ... scripts and the system/include directory is located, this is needed to tell cmake where the compiler and system headers are located

I'm not particularly happy with how the second path needs to be extracted from the .emscripten file, see the parse_config() and get_emscripten_root() functions here:

https://github.com/floooh/fips/blob/4d1d0872bf0f3c9378be5797111b3d1d402bb32a/mod/emsdk.py#L131-L161

...and this is even "easy mode" because I'm already using python there. This is where I think having an option for the emsdk-script to get the relevant paths would be better.

sbc100:
> ...ok I got it generally working, but I think getting the paths for the cmake toolchain file by "parsing" the right .emscripten file is too much hassle, especially since the file needs to be evaluated by a python interpreter.
> 
> I could reduce the information that needs to be handed to cmake via command line options to 2 locations:
> 
> * the emsdk directory itself, because that's where the .emscripten file and .emscripten_cache directory is located when doing `./emsdk activate --embedded xxx`, these are needed for the `--em-config` and `--cache` options passed to emcc/em++
> * the emscripten SDK "root" directory, this is where the emcc, em++, ... scripts and the system/include directory is located, this is needed to tell cmake where the compiler and system headers are located
> 
> I'm not particularly happy with how the second path needs to be extracted from the .emscripten file, see the parse_config() and get_emscripten_root() functions here:
> 
> https://github.com/floooh/fips/blob/4d1d0872bf0f3c9378be5797111b3d1d402bb32a/mod/emsdk.py#L131-L161
> 
> ...and this is even "easy mode" because I'm already using python there. This is where I think having an option for the emsdk-script to get the relevant paths would be better.

I've been wanting to come up with a better way for tools like cmake to find emscripten.  Thats really the only use for the EMSCRIPTEN_ROOT in the config file.    Those tools should not be parsing the emscripten config file... only emscripten should do that, and it already knows where it is.

There are a few options for how cmake can/should do this:
1) Find emcc in $PATH
2) Look for EMSCRIPTEN_ROOT or EMSCRIPTEN environment var?
3) Create a new ~/.emscripten_root file that just contains this value.

I think (1) and (2) make the most sense.  People who don't want emcc in $PATH can use (2).

kripken:
@floooh 

We can add EMSCRIPTEN_ROOT to the .emscripten file. I think the emsdk emits a pretty concise one atm. Please open an issue for that if it's still important, so we don't forget it.

The .emscripten file being a python file is useful so it can do things like check env variables, etc. But yeah, that does mean parsing is less trivial. We should probably move to a separate issue to discuss more in depth if we want to do something about that.

floooh:
Yes I agree. I'll play around a bit with the this merged PR in my build system integration, think a bit more about querying the paths and open a separate issue for this.

It'll be a few days though. I'm without fast internet until next week.

PR #249, opened 2019-05-24
Remove some waterfall code we forgot to remove earlier

kripken:


kripken:
This is no longer needed after more recent refactoring.

PR #250, opened 2019-05-29
 Switch over to emscripten-releases for 'latest'

kripken:
This makes `emsdk [install|activate] latest` get the latest fastcomp release from emscripten-releases, replacing our usage of the old mozilla build infrastructure.

This is only for "latest", so there should be no change to older releases.

This does remove non-release old stuff from the mozilla infrastructure, like "nightly" builds (that I don't think were even working?). After this PR, only old releases should be used from there.


kripken:
Context: we've asked that people test the emscripten-releases builds for 1.38.33, and fixed a few issues people found. There are no open showstoppers, so this is us moving to the next step, serving `latest` builds from the new infrastructure.

PR #251, opened 2019-05-29
Support getting specific releases from emscripten-releases, by their version

kripken:
This makes 

```
emsdk install 1.38.33
```

work (1.38.33 is from the new builders).

Also works with the old notation, `sdk-1.38.33-64bit`, and also supports `-(upstream|fastcomp)`.


Dockerfile
```diff
@@ -37,5 +37,16 @@ RUN cd /root/ \
...
+ && /root/emsdk/emsdk install sdk-1.38.31-64bit \
+ && /root/emsdk/emsdk activate tot-fastcomp \
+ && echo "test specific release (new, short name)" \
+ && /root/emsdk/emsdk install 1.38.33 \
+ && /root/emsdk/emsdk activate tot-fastcomp \
+ && python -c "import os ; assert 'fastcomp' in open(os.path.expanduser('~/.emscripten')).read()" \
+ && python -c "import os ; assert 'upstream' not in open(os.path.expanduser('~/.emscripten')).read()" \
+ && echo "test specific release (new, full name)" \
+ && /root/emsdk/emsdk install sdk-1.38.33-upstream-64bit \
+ && /root/emsdk/emsdk activate sdk-1.38.33-upstream-64bit \
```
sbc100: We should probably turn this into a separate script at some point..  this massive continuation is pretty ugly.

kripken: Yeah, agreed. Shell is really bad for this...

PR #253, opened 2019-05-30
Bundle java and python for emscripten-releases on windows, which matches the old behavior

kripken:


kripken:
They are currently on the mozilla S3. As part of moving the old binaries from there I'm moving these as well.

PR #258, opened 2019-06-02
Tag 1.38.34

kripken:
This should fix some version warnings people are currently seeing - we removed the fastcomp version warnings between repos check after the last version, in emscripten-core/emscripten#8664


kripken:
Merging quickly as we are getting multiple bug reports on this, #259

PR #261, opened 2019-06-03
Support ninja build tool

ekilmer:
~~Will only use `ninja` if it is on the path, otherwise uses Makefiles.~~

The user can choose which CMake generator to use by passing the `--generator` option, e.g. `--generator=Ninja` or `--generator="Unix Makefiles"` to `emsdk`.  

Default on *nix systems is `Unix Makefiles`

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -176,7 +176,8 @@ def vs_filewhere(installation_path, platform, file):
     return ''
 
 
-CMAKE_GENERATOR = 'Unix Makefiles'
+# Use ninja build tool if present
+CMAKE_GENERATOR = 'Ninja' if which('ninja') is not None else 'Unix Makefiles'
 if WINDOWS:
   # Detect which CMake generator to use when building on Windows
   if '--mingw' in sys.argv: CMAKE_GENERATOR = 'MinGW Makefiles'
@@ -870,7 +871,7 @@ def make_build(build_root, build_type, build_target_platform='x64'):
     else:
       make = ['mingw32-make', '-j' + str(CPU_CORES)]
   else:
-    make = ['make', '-j' + str(CPU_CORES)]
+    make = ['cmake', '--build', '.', '--', '-j' + str(CPU_CORES)]
 
   # Build
   try:
--- a/emsdk
+++ b/emsdk
@@ -176,8 +176,7 @@ def vs_filewhere(installation_path, platform, file):
     return ''
 
 
-# Use ninja build tool if present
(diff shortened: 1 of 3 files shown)


kripken:
Nice, thanks @ekilmer!

Let's wait a little before merging to see if anyone else has thoughts about this.

emsdk
```diff
@@ -174,7 +174,8 @@ def vs_filewhere(installation_path, platform, file):
     return ''
 
 
-CMAKE_GENERATOR = 'Unix Makefiles'
+# Use ninja build tool if present
+CMAKE_GENERATOR = 'Ninja' if which('ninja') is not None else 'Unix Makefiles'
```
trzecieu: Might be better to move the logic of selecting Generator to `else` after `if WINDOWS`.
Also might be worth to investigate if generator couldn't be passed to `./emsdk` with a flag i.e. `--generator=` then if given it will be selected unconditionally, otherwise auto-deduced

ekilmer: Thank you for the suggestion! I've rebased and included a new commit that hopefully addresses your comments.

test.py
```diff
@@ -132,7 +132,7 @@ def test_build(args, expected=None, unexpected=None):
 check_call('./emsdk activate sdk-tag-1.38.33-64bit')
 
 print('test binaryen source build')
-check_call('./emsdk install --build=Release binaryen-master-64bit')
+check_call('./emsdk install --build=Release --generator=Ninja binaryen-master-64bit')
```
kripken: It's good to add a test, thanks! But this removes the previous testing, was that intentional? I think we should still test the default configuration.

ekilmer: It was and it wasn't. It's a matter of preference I think to use either Makefile or Ninja as the build system in CI, since you control what is installed on the build server, and the resulting build artifacts/tests should be the same whether it's built using Makefiles or Ninja. 

I didn't want to change the default build system to use Ninja, since it's less commonly installed, but I can definitely change it back to use the default Makefiles. I will also remove `ninja-build` from the Dockerfile. I'll do this in a bit. Thanks!

ekilmer: Or perhaps using `--generator="Unix Makefiles"` would be better? Test both the `--generator` argument and still use the default Makefile build system?

kripken: Good idea, yeah, we can test the parameter using Makefiles.

kripken:
Looks good to me, thanks @ekilmer! If there are no other comments I'll merge this later today.

PR #264, opened 2019-06-04
Refactor test code into python

kripken:
Keep the basic workflow test (checking what a user would do to get started) in bash, but otherwise it's more convenient for most tests to be in python.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN cd /root/ \
  && apt-get update \
  && apt-get install -y python python3 cmake build-essential openjdk-9-jre-headless \
  && /root/emsdk/emsdk update-tags \
- && echo "test the standard workflow a user would do" \
+ && echo "test the standard workflow (as close as possible to how a user would do it, in the shell)" \
  && /root/emsdk/emsdk install latest \
  && /root/emsdk/emsdk activate latest \
  && source /root/emsdk/emsdk_env.sh --build=Release \
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,43 +10,14 @@ RUN cd /root/ \
  && apt-get update \
  && apt-get install -y python python3 cmake build-essential openjdk-9-jre-headless \
  && /root/emsdk/emsdk update-tags \
- && echo "test latest" \
+ && echo "test the standard workflow a user would do" \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -10,43 +10,14 @@ RUN cd /root/ \
...
- && echo "test latest" \
+ && echo "test the standard workflow a user would do" \
  && /root/emsdk/emsdk install latest \
  && /root/emsdk/emsdk activate latest \
- && python -c "import os ; assert 'fastcomp' in open(os.path.expanduser('~/.emscripten')).read()" \
- && python -c "import os ; assert 'upstream' not in open(os.path.expanduser('~/.emscripten')).read()" \
  && source /root/emsdk/emsdk_env.sh --build=Release \
  && emcc hello_world.cpp \
  && emcc hello_world.cpp -s WASM=0 \
  && emcc --clear-cache \
```
sbc100: Why not put everything above other that the apt-get commands in the test.py script too?

test.py
```diff
@@ -0,0 +1,56 @@
...
+assert 'fastcomp' in open(os.path.expanduser('~/.emscripten')).read()
+assert 'upstream' not in open(os.path.expanduser('~/.emscripten')).read()
+
+print('test specific release (new, full name)')
+check_call('./emsdk install sdk-1.38.33-upstream-64bit')
+check_call('./emsdk activate sdk-1.38.33-upstream-64bit')
+
+print('test binaryen source build')
+check_call('./emsdk install --build=Release binaryen-master-64bit')
+
```
sbc100: Remove trailing newline.

Dockerfile
kripken: I feel safer knowing we run the most basic commands like a user would - in the shell. Python is a layer on top, and maybe the path works differently, or escaping, or the environment, or something else... I changed the comment to reflect that.

sbc100: When I suggest this change it wasn't so much about converting to python but moving into a separate script that can be run outside the Dockfile (not least to avoid the horrible command continuation syntax used in Dockerfiles,).  

So just moving to `test.sh` would be fine with me.  But either way this is an improvement.

kripken: Oh, I see. Yeah, that would be an additional improvement, I agree.

PR #266, opened 2019-06-05
Move testing into a bash file on the side

kripken:
As suggested recently.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,7 +1,7 @@
 echo "test the standard workflow (as close as possible to how a user would do it, in the shell)"
-/root/emsdk/emsdk install latest
-/root/emsdk/emsdk activate latest
-source /root/emsdk/emsdk_env.sh --build=Release
+./emsdk install latest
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -0,0 +1,8 @@
+echo "test the standard workflow (as close as possible to how a user would do it, in the shell)"
+./emsdk install latest
+./emsdk activate latest
+source ./emsdk_env.sh --build=Release
+emcc hello_world.cpp
+emcc hello_world.cpp -s WASM=0
+emcc --clear-cache
+
```
sbc100: I hate to say it but why not just merge test.py back into this file :-/    I know you just ported it to python..

sbc100: Actually its probably quite easy to take the original unported shell code from the Dockerfile and dump it in here. so might not be too much effort.

kripken: I think the python will quickly become better as the tests get more sophisticated, e.g. #265

PR #271, opened 2019-06-08
Fix ./emsdk update of a *non*-git install

kripken:
The non-git case was not well documented and not tested, and it was broken since we switched the build infrastructure.

In the git case, an update is to do a git pull. In the non-git case, it downloads a zip and unpacks it. The old build infra apparently had builds created for this; instead, I made it download it directly from github. Perhaps we should consider creating builds for this as well eventually?

Also add a test for this, so we never break it again.

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -45,6 +45,8 @@ emscripten_releases_repo = 'https://chromium.googlesource.com/emscripten-release
 
 emscripten_releases_download_url_template = "https://storage.googleapis.com/webassembly/emscripten-releases-builds/%s/%s/wasm-binaries.%s"
 
+emsdk_zip_download_url = 'https://github.com/emscripten-core/emsdk/archive/master.zip'
+
 zips_subdir = 'zips/'
 
 # Enable this to do very verbose printing about the different steps that are being run. Useful for debugging.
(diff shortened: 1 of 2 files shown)


sbc100:
Do you know what was different with the builds of emsdk vs just downloading master.zip?  Presumably some kind of packaging was being done that we want to continue to do?

kripken:
I'm not sure, but I don't think there was a difference - the old zip downloads were used exactly as the new ones are, that is, they are unzipped over the emsdk directory, and so updating it. And then the emsdk is then expected to be run normally.

kripken:
Landing sooner as there have been user reports about this problem, but please review when you have the time.

juj:
> Do you know what was different with the builds of emsdk vs just downloading master.zip? Presumably some kind of packaging was being done that we want to continue to do?

Packaging of emsdk was done by this script: https://github.com/juj/emslave/blob/master/bin/deploy_emsdk.py . The main part is that it includes portable python on Windows. (Historically it would also do NSIS installer scripts for Windows, and install Visual Studio 2013 integration, but we dropped doing those at some point)

kripken:
Thanks @juj, I opened #272 to discuss this more.

I think portable python only matters for the first install, which is what I made the topic in that new issue. That is, for updating, what this PR is about, the user already has python (otherwise they couldn't run the emsdk to update), and so those downloads don't need python. In other words I don't think we need to change anything in this PR, but improving the first install experience for windows is still an open question.

PR #273, opened 2019-06-12
Default Python 3 for running emsdk

trzecieu:
On macOS it's pretty common to hit a problem
```
Error downloading URL 'https://github.com/kripken/emscripten/archive/1.38.25.tar.gz': <urlopen error [SSL: TLSV1_ALERT_PROTOCOL_VERSION] tlsv1 alert protocol version (_ssl.c:590)>
```
When an installation is started. I.e. `./emsdk install sdk-1.38.25-64bit` - this is caused in majority of cases by some problem of openssl and python2 - (some other https downloads work just fine, not sure why GitHub is special)

So this Pull Request will default Python to Python 3, for running `emsdk`

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 from __future__ import print_function
 
--- a/emsdk
+++ b/emsdk
@@ -1,4 +1,7 @@
-#!/usr/bin/env python3
+#!/bin/sh
+# Following line is responsible for selecting the best python for current host
+# This script can be called directly: `./emsdk` or indirectly: `python ./emsdk` `sh ./emsdk` or `bash ./emsdk`
+"exec" "`dirname $0`/python_selector" "$0" "$@"
 
 from __future__ import print_function
 
--- a/python_selector
+++ b/python_selector
@@ -1,0 +1,29 @@
+#!/usr/bin/env python
+
+# Copyright 2019 The Emscripten Authors.  All rights reserved.
+# Emscripten is available under two separate licenses, the MIT license and the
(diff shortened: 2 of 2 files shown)


kripken:
Thanks @trzecieu!

Is python3 always present on macOS? On linux it sometimes is not, which is why we have an indirection in emscripten, [python_selector](https://github.com/emscripten-core/emscripten/blob/incoming/tools/python_selector.py). Here it is used in emcc for example: https://github.com/emscripten-core/emscripten/blob/incoming/emcc that script should run in both python2 and 3. Then the selector can force a specific version (I think emscripten's doesn't do that yet though). Maybe that's the safest thing to do here as well?

trzecieu:
Good point with python selector, I've made some improvement to that. 
So basically this PR makes `emsdk` script able to be executed with following ways:
Explicitly define python executor:
* `python3 ./emsdk`: Will use `python3`
* `python2 ./emsdk`: Will use `python2`
* `python ./emsdk`: Will use system `python`

Will call `python_selector` script that will define the best choice, basically for Linux / macOS will test existence of `python3` then will fall-back to `python` or `sys.executable in some strange edge case
* `bash ./emsdk`:
* `sh ./emsdk`:
* `./emsdk`:

kripken:
Thanks @trzecieu, looks good!

Regarding testing, I think our current ones are enough after the recent improvements. We test `./emsdk` [here](https://github.com/emscripten-core/emsdk/blob/master/test.sh) and `python` and `python2` [here](https://github.com/emscripten-core/emsdk/blob/master/test.py). We don't explicitly test for `python3` because such a link doesn't exist on the CI, but `python` is version 3 there. Perhaps it might be good to eventually look into directly testing python3, and/or directly testing those `which()` commands.

PR #274, opened 2019-06-13
Clean up cleanup code + add tests

kripken:
The code to clean up downloaded zips right after they are used is odd: it looks at only some extensions (zip, tar, gz, but not bzip2, tbz2, xz - and we use some of all of those!), and also it's odd to delete the download immediately as then if the tool is installed right after it will be downloaded again. Other code in the emsdk works to avoid that.

This removes that code, so we leave downloaded zips in the zips/ folder, consistently.

Also adds a bunch of tests for the current behavior, to avoid regressions.

Background is that we need to figure out something for file downloads, see emscripten-core/emscripten#8791. This does not fix that, but the cleanup and tests should make it easier once we decide on a path.


juj:
> it looks at only some extensions (zip, tar, gz

Originally we only had packages with these suffixes, the bzip2 etc. formats have been added later. I think the reason I authored to look at extensions was due to some downloads being .exe or suffixless iirc, that needed different handling, or something like that.

> it's odd to delete the download immediately as then if the tool is installed right after it will be downloaded again.

The idea was that after a successful install, the source zip package is deleted to avoid cluttering disk space. Though I think the root problem around this is the check for "successful" - for most cases, testing for success has been to test for the existence of the output directory and that it contains files, which is probably not robust enough.

If a tool is attempted to be installed again right after it has already been installed, it should be a no-op, as the download+instlall is skipped as redundant since that tool is already installed.

Deleting the zip does mean that if one first installs a tool (downloading it), then uninstalls it, and then re-installs the same tool, then the re-install would come with a re-download of that zip, which I think would be expected and desired.

Not sure if it's good to leave the downloaded zips lying around - I don't particularly care if they are, though I wonder what is the main benefit of doing that? For the purpose of avoiding a redownload on a reinstall?

PR #275, opened 2019-06-13
python_selector checks if found python command is valid

trzecieu:
Addresses https://github.com/emscripten-core/emscripten/issues/8792

I've found what's the reason of the strange behaviour of last changes (https://github.com/emscripten-core/emsdk/pull/273). 

In order to reproduce this problem it's needed:
* To don't have python3 installed in system
* To have `pyenv` with installed at least one version from each family, like:
```
pyenv install 3.7.0
pyenv install 2.7.15
```
* To activate a version from python2
```
pyenv local 2.7.15
```

Then we have a strange situation that `python3` is available as a command, but it's just a mock for `pyenv`:
```bash
➜ pyenv versions
  system
* 2.7.15 (set by /home/trzeci/Projects/emsdk/.python-version)
  3.7.0

➜ which python3
~/.pyenv/shims/python3

➜ python3 --version
pyenv: python3: command not found

The `python3' command exists in these Python versions:
  3.7.0

➜ ./emsdk --help        
pyenv: python3: command not found

The `python3' command exists in these Python versions:
  3.7.0

```

As you see in an example above I didn't even call `emsdk` script, just tried to execute `python3 --version` and I've got exactly the same error. The problem of pyenv is that even if python3 isn't active it's available in the system PATH. 
Calling `./emsdk --help` get's the same error, as `python3` was used.

At the moment this entirely breaks logic in `python_selector` and in pyenv working command is indistinguishable from stub python command.

@kripken: A couple of ways to go from here:
* revert logic of python_selector and advertise mac users to use python3
* keep logic and advertise pyenv users to use python3
* This PR: extend logic of `python_selector` by checking if found python command is valid (for the instance by calling `python --version`), this will filter out false positive matches that coming from `pyenv`

After patch:
```bash
➜ pyenv versions        
  system
* 2.7.15 (set by /home/trzeci/Projects/emsdk/.python-version)
  3.7.0

➜ which python3 
~/.pyenv/shims/python3

➜ python3 --version
pyenv: python3: command not found

The `python3' command exists in these Python versions:
  3.7.0


➜ ./emsdk --help
 emsdk: Available commands:

   emsdk list [--old] [--uses]  - Lists all available SDKs and tools and their
                                  current installation status. With the --old
   ....

```
As you can see `python3 --version` command fails as it was, but that's expected behaviour. What's important is that `./emsdk --help` gets executed as `python_selector` recognizes a stub.

### DIFF
--- a/python_selector
+++ b/python_selector
@@ -16,7 +16,12 @@ def which(program):
   for path in [""] + os.environ.get("PATH", "").split(os.pathsep):
     exe_file = os.path.join(path, program)
     if os.path.isfile(exe_file) and os.access(exe_file, os.X_OK):
-      return exe_file
+      # in case of pyenv we might have `python3` command available, but it's just a stub
+      # https://github.com/emscripten-core/emscripten/issues/8792
+      p = subprocess.Popen([exe_file, "--version"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+      p.communicate()
+      if p.returncode == 0:
+        return exe_file
   return None
 
 # Look for the best choice for python, favours Python3 over Python2


python_selector
```diff
@@ -16,7 +16,12 @@ def which(program):
   for path in [""] + os.environ.get("PATH", "").split(os.pathsep):
     exe_file = os.path.join(path, program)
     if os.path.isfile(exe_file) and os.access(exe_file, os.X_OK):
-      return exe_file
+      # in case of pyenv we might have `python3` command available, but it's just a stub
+      # https://github.com/emscripten-core/emscripten/issues/8792
+      p = subprocess.Popen([exe_file, "--version"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
kripken: I'm not sure this is enough - when I check locally, `pyenv --version` works (prints the `pyenv 1.2.12`, and exits with `0`). Perhaps we can check for the output starting with `Python `? But maybe there's something more robust.

kripken: Oh, I guess it would actually error if pyenv didn't find python3 on my system?

trzecieu: In the `python_selector` we always check found python commands, not `pyenv` and good part is that python stubs created `pyenv` return 137 exit code in the problematic case. 

I'm more afraid of string parsing, as this might be more false positive results.
Just in case checked some major flavours of python: 
```bash
#cpython 
➜  python --version 
Python 3.7.0

#pypy3
➜ pypy3 --version
Python 3.5.3 (11af55503d5c, May 23 2019, 09:37:40)
[PyPy 7.0.0 with GCC 9.1.1 20190503 (Red Hat 9.1.1-1)]

#stackless
➜ python --version           
Python 3.5.4

#anaconda 
➜ python --version
Python 3.7.0
(anaconda3-5.3.1) 

#miniconda 
➜ python --version
Python 3.6.3 :: Anaconda, Inc.
(miniconda3-4.3.30) 
``` 

As you've mentioned Ubuntu, I did a quick test of calling python3 command on a fresh installation of 19.04: 
```bash
trzeci@ubuntu1904:~$ python

Command 'python' not found, but can be installed with:

sudo apt install python3         # version 3.7.3-1, or
sudo apt install python          # version 2.7.16-1
sudo apt install python-minimal  # version 2.7.16-1

You also have python3 installed, you can run 'python3' instead.

trzeci@ubuntu1904:~$ echo $?
127
trzeci@ubuntu1904:~$ which python
trzeci@ubuntu1904:~$ 

```
Hence it looks like that's system feature, not a stub command

kripken: Thanks, yeah, I agree. lgtm!

kripken:
And thanks @trzecieu for the quick fix!

python_selector
trzecieu: > Oh, I guess it would actually error if pyenv didn't find python3 on my system?

It's already guarded by: `if os.path.isfile(exe_file) and os.access(exe_file, os.X_OK):` hence it shouldn't error if `python3` command isn't available. 

Thanks for merging! 
// Cheers

sbc100:
I think if you put something called `python3` in your PATH then you should expect people to be able to try to use it.

This seems like a bug in pyenv that we shouldn't have to deal with here.   

I mean what if I symlink `/usr/bin/python3` to `/usr/bin/bash`.. `--version` will still work, but clearly we don't have to handle that case where python3 is broken.

Issue #72, opened 2017-04-28
incorrect node on non-x86/x86_64 systems


H-Plus-Time:
emsdk misidentifies the system architecture on ARM systems (I'm fairly certain that the arch checks extend as far as 32/64 bit checks), resulting in the incorrect version of node (and presumably python going by the source) being downloaded - if you replace the contents of node/<whatever_v_emsdk_downloads> with that of the equivalent arm version of node, everything works perfectly.

juj:
ARM hosts are currently not supported, but Linux/Windows/OS X with 32bit/64bit is assumed. Not sure how to develop a fix since I don't have a suitable ARM host to try on, but PRs with respect to this are welcome.

cheako:
QEMU has an emulator for ARM, plus u could pick up a rasbpi for under $50.

brion:
Raspberry pi doesn't have enough ram to build, I'm pretty sure. :) I've got a machine I'm going to do some testing on though, might or might not have any luck with it.

brion:
On a Lenovo C630 (Windows 10 on Snapdragon 850) running Ubuntu Bionic in Windows Subsystem for Linux, I was able to build via `emsdk install sdk-incoming-64bit`. It downloaded an x86-64 node binary, so I had to manually install the distro's node (`apt-get install nodejs`) and change the `NODE` setting in `~/.emscripten`.

Note the LLVM build took about 2 hours, on a machine with 8 cores and 8 GiB RAM.

Resulting build does appear to work, at least on a "hello world" .c file. :D

Things that need fixing:
[ ] distinguish arm and arm64 from x86 and x86-64 somehow, so it doesn't keep trying to install x86/x86-64 stuff
[ ] either download an arm/arm64 node, or require it to be installed separately
[ ] ideally, provide regular builds for arm64 (this may be tough to set up)

brion:
Ok I see that emsdk distinguishes only between 64-bit and non-64-bit:

```
def is_os_64bit(): # http://stackoverflow.com/questions/2208828/detect-64bit-os-windows-in-python
  return platform.machine().endswith('64')
```

So it'll see the "aarch64" and report back 64-bit, but doesn't distinguish arches in any other way. This can probably be modified around to return a normalized arch name; to keep back-compat could keep calling x86 `32bit` and x64 `64bit`, and add `arm32` and `arm64` or something alongside.

[...]

PR #277, opened 2019-06-15
Initial steps to Linux ARM support for emsdk

brion:
Provisional patch for stuff like https://github.com/emscripten-core/emsdk/issues/72 when working on ARM/ARM64 Linux. This could be extended a bit further to allow downloading Linux ARM 32-bit or 64-bit binaries under another key ('linux_arm_url' vs 'linux_url' for instance?) if they get added in future.

Tested lightly on aarch64 Linux, not yet tested on 32-bit or on x86/x64 Linux.

This treats Linux on ARM as a separate Unix-like OS for emsdk purposes, so it won't try to download any x86/x64 binary packages as it previously did.

Since there's no prebuilt node and the sdk metapackages depend on it, this requires manually issuing commands for the various sub-packages:

./emsdk install clang-incoming-64bit
./emsdk install binaryen-master-64bit
./emsdk install emscripten-incoming-64bit

Currently the error messages if you ask for something that needs a binary package aren't good, they just say stuff like 'tool not found', but they won't be advertised to you for install either.


emsdk
```diff
@@ -75,10 +75,19 @@ if platform.mac_ver()[0] != '':
   ENVPATH_SEPARATOR = ':'
 
 LINUX = False
+LINUX_ARM = False
```
juj: Perhaps instead of having a variable `LINUX_ARM`, we could have a variable `ARCH` that is one of `'x86', `'x64'`, '`arm'` or `'aarch64'`? that would be more future proof to avoid growing more variables. (In future a refactor might change OSX/WINDOWS/LINUX vars to an `OS='win'/'linux'/'mac'` string)

Then to abbreviate checks, could have helpers
```py
def arch_is_arm():
  return ARCH == 'arm' or ARCH == 'aarch64'
```
that way the amount of state will stay minimal/nonredudant, and e.g. it wouldn't be possible to mismatch `LINUX_ARM=True` with `LINUX=False`.

brion: Yeah that'd be better in the long run, I'll try that this weekend.

juj:
Looks good to me! There is some CI failure though, not sure if it's due to the CI or due to this PR, so I'll also defer to e.g.  @kripken to take a peek before landing.

brion:
I think the CI failure was from my inconsistent use of 'x64' vs 'x86_64'; hopefully should pass on this next run. :)

kripken:
Looks good to me too, thanks @brion!

CI looks red though?

brion:
Hmm, CI seems to have failed during a download?

```
 [---------------------------------------------------------Error downloading URL 'https://storage.googleapis.com/webassembly/emscripten-releases-builds/old/linux/emscripten-llvm-e1.38.31.tar.gz': [Errno 104] Connection reset by peer
```

kripken:
Might have been a temporary network error on CI, I restarted it.

kripken:
Looks good!

kripken:
What's a good commit message for the squash?

brion:
(Going to check tomorrow if this still applies cleanly and then I'll squash it)

brion:
Went ahead and squashed it while waiting at the airport; had to make a small update for a merge conflict.

Will test again tomorrow on my aarch64 and armv7 machines to confirm everything's still working as expected.

brion:
Confirmed it's working 32-bit on Raspberry Pi 4 running NOOBS raspbian variant (took about 2.5 hours to compile with active cooling) and 64-bit on Lenovo Yoga C630 under WSL Ubuntu (took about 1.5 hours to compile with passive cooling and some CPU throttling).

The correct node versions are downloaded, and I can build with it and run the resulting JS/wasm. (Tested sdk-incoming-32bit and sdk-incoming-64bit, so fastcomp.)

kripken:
Thanks @brion!

PR #285, opened 2019-06-25
Exclude `fastcomp` and `releases`

### LINKED ISSUE #284: Include `fastcomp` and `releases` in `.gitignore`
Including `fastcomp` and `releases` in the `.gitignore` file allows the `emsdk` repository to be used in-place as a submodule.

The `fastcomp` and `releases` directories are downloaded into the `emsdk` directory, which makes the submodule as `dirty`.

By adding them into the `.gitignore` file the submodule won't be marked `dirty` after use, making life a little easier.

Here is an example repository where I use this pattern: https://github.com/Ethsnarks/miximus-emscripten

HarryR:
Fixes #284

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,4 +26,5 @@ llvm-tags-32bit.txt
 llvm-tags-64bit.txt
 upstream
 emscripten-releases-tot.txt
-
+fastcomp
+releases


kripken:
Thanks @HarryR!

PR #291, opened 2019-06-26
Extra testing for library building

kripken:
Verify the latest things we've started to ship pre-built (libc, struct_info, optimizer) are in fact not built on the user's machine.


kripken:
(These tests could not land with that work, since it had to wait on tagging a new version, as we test it on the latest tag.)

PR #292, opened 2019-07-02
Use a wrapper script for running emsdk.py

sbc100:
This change moves the python code for emsdk into a file ending in .py.
This script is then run via emsdk.bat on windows or emsdk (a shell
script) on non-windows.

This avoid the #!/bin/sh at the top of the python script and the "exec"
hack on the first line that re-runs it under python.  Hopefully this
preserves the intent of #273 without jumping through so many hoops.


kripken:
cc @trzecieu

test.py
```diff
@@ -31,7 +32,7 @@ def failing_call_with_output(cmd, expected):
   assert expected in stdout, 'call did not have the right output'
 
 def hack_emsdk(marker, replacement):
-  src = open('emsdk').read()
+  src = open('emsdk.py').read()
```
kripken: why are the tests changed? none of the old usages are tested, so we lost coverage for the old `./emsdk` use pattern, which is what users are doing. best to test both (the official/intended, and also I assume some people will call the new `emsdk.py` directly because they find it).

test.py
```diff
@@ -87,8 +88,8 @@ def test_build(args, expected=None, unexpected=None):
 check_call('./emsdk update-tags')
 
 print('test latest-releases-upstream')
-check_call('python2 ./emsdk install latest-upstream')
-check_call('./emsdk activate latest-upstream')
+check_call('python2 ./emsdk.py install latest-upstream')
+check_call('python3 ./emsdk.py activate latest-upstream')
```
trzecieu: In addition I'd keep the old check `check_call('./emsdk activate latest-upstream')`

trzecieu:
Looks good, indeed it will be much cleaner. In the current form this script will hit issue addressed by #275. That is in case of pyenv, `python3` command might be available but it will be just a mock script.

test.py
sbc100: This test in particular is hacking the python source itself, so it needs to change.

test.py
sbc100: Reverted

PR #300, opened 2019-07-12
[windows] Replace python2-specific error handling with a string check that works in python3 as well

### LINKED ISSUE #298: `emsdk activate --global` fails with Python 3 on Windows due to different KeyError behaviour
Steps to reproduce:

1. Install Python 3 and add it to PATH:
```
> python --version
Python 3.7.3
```
2. Start as a usual user (not Administrator) at `C:\`
3. `git clone https://github.com/emscripten-core/emsdk.git` (HEAD is at 49652605051d9faca2dcb781918a028bdfbc2703 at the moment)
4. `cd emsdk`
5. `emsdk install latest-upstream`
6. Open an Administrator terminal
7. `cd C:\emsdk`
8. `emsdk activate --global latest-upstream`

Expected: successful activation.

Got:
```
Microsoft Windows [Version 10.0.18362.175]
(c) 2019 Microsoft Corporation. All rights reserved.

C:\Windows\system32>cd c:\emsdk

c:\emsdk>emsdk activate --global latest-upstream
Registering active Emscripten environment globally for all users.

Writing .emscripten configuration file to user home directory C:\Users\esuvorov/
Backing up old Emscripten configuration file in C:\Users\esuvorov\.emscripten.old
The Emscripten configuration file C:\Users\esuvorov\.emscripten has been rewritten with the following contents:

import os
LLVM_ROOT = 'c:/emsdk/upstream/bin'
BINARYEN_ROOT = 'c:/emsdk/upstream'
NODE_JS = 'c:/emsdk/node/8.9.1_64bit/bin/node.exe'
SPIDERMONKEY_ENGINE = ''
V8_ENGINE = ''
TEMP_DIR = 'C:/Users/esuvorov/AppData/Local/Temp'
COMPILER_ENGINE = NODE_JS
JS_ENGINES = [NODE_JS]

Adding directories to PATH:
PATH += c:\emsdk
PATH += c:\emsdk\upstream\emscripten
PATH += c:\emsdk\node\8.9.1_64bit\bin

EMSDK = c:/emsdk
EM_CONFIG = C:\Users\esuvorov\.emscripten
EMSDK_NODE = c:\emsdk\node\8.9.1_64bit\bin\node.exe

  File "c:\emsdk\\emsdk", line 286, in win_get_environment_variable
    return os.environ[key]
  File "C:\Users\esuvorov\AppData\Local\Programs\Python\Python37-32\lib\os.py", line 678, in __getitem__
    raise KeyError(key) from None
KeyError: 'EMSDK_NODE'

During handling of the above exception, another exception occurred:

  File "c:\emsdk\\emsdk", line 2697, in <module>
    sys.exit(main())
  File "c:\emsdk\\emsdk", line 2636, in main
    tools_to_activate = set_active_tools(tools_to_activate, permanently_activate=arg_global)
  File "c:\emsdk\\emsdk", line 2129, in set_active_tools
    tool.win_activate_env_vars(permanently_activate=True)
  File "c:\emsdk\\emsdk", line 1495, in win_activate_env_vars
    win_delete_environment_variable(key, False) # If there is an env var for the LOCAL USER with same name, it will hide the system var, so must remove that first.
  File "c:\emsdk\\emsdk", line 354, in win_delete_environment_variable
    win_set_environment_variable(key, None, system)
  File "c:\emsdk\\emsdk", line 316, in win_set_environment_variable
    previous_value = win_get_environment_variable(key, system)
  File "c:\emsdk\\emsdk", line 288, in win_get_environment_variable
    if e[0] != 2: # 'The system cannot find the file specified.'
TypeError: 'KeyError' object is not subscriptable

c:\emsdk>
```

Looks like the problem is in [`win_get_environment_variable`](https://github.com/emscripten-core/emsdk/blob/49652605051d9faca2dcb781918a028bdfbc2703/emsdk#L283-L288) at the very least. Python 3 does not have `pywin32` by default, so the first `try` fails, then `os.environ[key]` throws `KeyError` for non-existent variable, it is caught by the second `except` and everything fails when the code tries to access `[0]` of `KeyError`. It's ok for some reason in Python 2 (`[0]` simply returns the key of `KeyError`), but fails in Python 3, hence the error above.

kripken:
Fixes #298

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -251,7 +251,7 @@ def win_set_environment_variable_direct(key, value, system=True):
     win32api.RegSetValueEx(folder, key, 0, win32con.REG_EXPAND_SZ, value)
     if VERBOSE: print('Set key=' + key + ' with value ' + value + ' in registry.')
   except Exception as e:
-    if 'Access is denied.' in str(e):
+    if e.args[0] == 5 or e.args[2] == 'Access is denied.':
       print('Error! Failed to set the environment variable \'' + key + '\'! Setting environment variables permanently requires administrator access. Please rerun this command with administrative privileges. This can be done for example by holding down the Ctrl and Shift keys while opening a command prompt in start menu.')
       sys.exit(1)
     print('Failed to write environment variable ' + key + ':', file=sys.stderr)
@@ -285,7 +285,7 @@ def win_get_environment_variable(key, system=True):
       # need to be precise not to set these back to system registry, or expansion items would be lost.
       return os.environ[key]
   except Exception as e:
-    if 'The system cannot find the file specified.' not in str(e):
+    if e.args[0] != 2: # 'The system cannot find the file specified.'
       print('Failed to read environment variable ' + key + ':', file=sys.stderr)
       print(str(e), file=sys.stderr)
     try:
--- a/emsdk
+++ b/emsdk
@@ -251,7 +251,7 @@ def win_set_environment_variable_direct(key, value, system=True):
     win32api.RegSetValueEx(folder, key, 0, win32con.REG_EXPAND_SZ, value)
     if VERBOSE: print('Set key=' + key + ' with value ' + value + ' in registry.')
   except Exception as e:
-    if e[0] == 5 or e[2] == 'Access is denied.':
+    if 'Access is denied.' in str(e):
       print('Error! Failed to set the environment variable \'' + key + '\'! Setting environment variables permanently requires administrator access. Please rerun this command with administrative privileges. This can be done for example by holding down the Ctrl and Shift keys while opening a command prompt in start menu.')
       sys.exit(1)
     print('Failed to write environment variable ' + key + ':', file=sys.stderr)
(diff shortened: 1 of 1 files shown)


emsdk
```diff
@@ -251,7 +251,7 @@ def win_set_environment_variable_direct(key, value, system=True):
     win32api.RegSetValueEx(folder, key, 0, win32con.REG_EXPAND_SZ, value)
     if VERBOSE: print('Set key=' + key + ' with value ' + value + ' in registry.')
   except Exception as e:
-    if e[0] == 5 or e[2] == 'Access is denied.':
+    if 'Access is denied.' in str(e):
```
yeputons: That won't work on non-English Windows systems (e.g. Russian):

```
Failed to write environment variable PATH:
(5, 'RegOpenKeyEx', 'Отказано в доступе.')
```

Similarly below. I still think using `.args[0]` is a better fix.

kripken: Ah right, thanks - the old code was doing that so I just assumed it was ok, but of course it isn't! :)

Ok, I changed it all to just use the code, and just using `e.args`.

kripken:
review ping @jgravelle-google @sbc100

Issue #138, opened 2018-03-28
Calling "emsdk activate --global latest" hangs when setting environment variables.


timhutton:
Platform: Windows 10, Python 2.7.13

Without the --global option the command runs very quickly and succeeds. With --global it hangs here:

```
...
Setting environment variables:
EMSDK = c:/emscripten

```

**WORKAROUND:** ```emsdk activate latest```

(From [this commit](https://github.com/juj/emsdk/commit/2ed7a605a90ca72f750ffd04df736e67d4523ff4) it looks like this acts globally anyway.)

sbryfcz:
I have the same issue. Platform Windows 10. Python 3.6.3.

juj:
With `--global` the Windows UAC elevation prompt should appear to ask for permission to permanently add the PATHs and environment variables to registry. If that does not happen and you're getting a hang before that line, I wonder if [running Windows Command Prompt in Administrator mode](https://www.howtogeek.com/194041/how-to-open-the-command-prompt-as-administrator-in-windows-8.1/) might help avoid the issue?

sbryfcz:
The UAC prompt never appeared for me. And I tried in Adminstrator mode but nothing happened except it hanging.

d07RiV:
Same issue. It worked fine on my other machine, but here no matter what I do, it hangs before setting LLVM_ROOT, and none of the variables it claims to have set appear in global environment. Tried both cmd and powershell in administrator mode.

aplumly:
anyone find a fix for this?... apart from setting all the env variables by hand?

roman-petrov:
Today I spent a hour trying to investigate this issue on my machine. As I see the hang caused by `win32api.SendMessage(win32con.HWND_BROADCAST, win32con.WM_SETTINGCHANGE, 0, 'Environment')`
call in `emsdk` Python script.

Found some on StackOverflow https://stackoverflow.com/questions/1951658/sendmessagehwnd-broadcast-hangs

Still not sure if there is a simple workaround, but maybe removing this system call should work. Or we can use 'SendMessageTimeout` a timeout in this call as they suggested on `StackOverflow`.

I am new with `emsdk`  so I am not sure what is the right solution, need help...

roman-petrov:
Right solution might be to use `PostMessage` vs `SendMessage` in https://github.com/emscripten-core/emsdk/blob/master/emsdk#L263

Can anybody give an advice?...

PR #301, opened 2019-07-16
Fix 'SendMessage' hanging on Windows

roman-petrov:
Fixes https://github.com/emscripten-core/emsdk/issues/138

As I got from my research and [StackOverflow](https://stackoverflow.com/questions/1951658/sendmessagehwnd-broadcast-hangs), WinAPI `SendMessage` function doesn't return until all receivers have processed the message. So it might hang sometimes, this happens on my machine and some other users report such problem.

We can use `PostMessage` instead of `SendMessage`. Another solution is to use [SendMessageTImout](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-sendmessagetimeouta), but I am not sure what constant timeout value we should pass.

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -260,7 +260,7 @@ def win_set_environment_variable_direct(key, value, system=True):
 
   win32api.RegCloseKey(folder)
   os.environ['PATH'] = prev_path
-  win32api.SendMessage(win32con.HWND_BROADCAST, win32con.WM_SETTINGCHANGE, 0, 'Environment')
+  win32api.PostMessage(win32con.HWND_BROADCAST, win32con.WM_SETTINGCHANGE, 0, 'Environment')
 
 
 def win_get_environment_variable(key, system=True):


kripken:
What's the worst case if that message is sent asynchronously and not received before it is used? I'm not sure what would read those posted values.

roman-petrov:
> What's the worst case if that message is sent asynchronously and not received before it is used? I'm not sure what would read those posted values.

It's my first PR for `emscripten-core` so I am not sure why `SendMessage` was used instead of `PostMessage` so I am unable to answer your question...  I am not know if there is better solution for the issue and I am ready to merge this PR, please let me know if you can suggest any modifications.

kripken:
Sounds good @roman-petrov - this is probably fine to merge. To make sure, you tested it locally on windows and it works well?

roman-petrov:
> Sounds good @roman-petrov - this is probably fine to merge. To make sure, you tested it locally on windows and it works well?

@kripken , yes, I tested this workaround on my machine and the problem now seems to be gone. I not found any regression related to this fix, but I still have some problems activating`emsdk` globally
using `emsdk activate latest --global` command, like

image

Anyway, these issues seems to be not related to the issue we are discussing and might be something like https://github.com/emscripten-core/emsdk/issues/298.

So I am merging this PR now, thanks for all help.

kripken:
Great, merging.

roman-petrov:
thanks, @kripken !

PR #310, opened 2019-08-04
Fix issue with windows install

SagaraBattousai:
Windows version of python 3 returns platform.machine() in upper case for AMD64 which wasn't being checked for so added .lower().

Doesn't affect Linux python3 which already worked.

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -82,7 +82,7 @@ if not OSX and (platform.system() == 'Linux' or os.name == 'posix'):
 UNIX = (OSX or LINUX)
 
 ARCH = 'unknown'
-machine = platform.machine()
+machine = platform.machine().lower() #amd64 was not being recognised when returned as AMD64
 if machine.startswith('x64') or machine.startswith('amd64') or machine.startswith('x86_64'):
   ARCH = 'x86_64'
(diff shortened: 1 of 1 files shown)


emsdk
```diff
@@ -82,7 +82,7 @@ if not OSX and (platform.system() == 'Linux' or os.name == 'posix'):
 UNIX = (OSX or LINUX)
 
 ARCH = 'unknown'
-machine = platform.machine()
+machine = platform.machine().lower() #amd64 was not being recognised when returned as AMD64
```
kripken: please add a space after the `#`

also please reword the comment for future readers (current comment looks good for readers of the patch now), like maybe `# lower() handles 'AMD64' which can happen in some cases`

kripken:
We are getting more bug reports here, so let me merge this quickly now and I'll fix up the comment myself. Thanks again @SagaraBattousai!

SagaraBattousai:
> We are getting more bug reports here, so let me merge this quickly now and I'll fix up the comment myself. Thanks again @SagaraBattousai!

Sorry, I didn't see this, thanks for fixing the comment :)

Issue #312, opened 2019-08-06
emsdk install latest fails on Windows


fibbo:
I downloaded the latest Zip and when running `emsdk install latest` I get the following console output

```
E:\> cd emsdk-master

E:\emsdk-master> emsdk install latest
Warning: unknown machine architecture AMD64

Fetching emscripten-releases repository...
Cloning into 'E:/emsdk-master/releases'...
remote: Total 3119 (delta 932), reused 3119 (delta 932)
Receiving objects: 100% (3119/3119), 1.52 MiB | 2.47 MiB/s, done.
Resolving deltas: 100% (932/932), done.
Fetching latest changes to the branch 'master' for 'E:/emsdk-master/releases'...
Already up to date.
Successfully updated and checked out branch 'master' on repository 'E:/emsdk-master/releases'
Current repository version: "Tue, 6 Aug 2019 00:29:45 +0000 c911897c9caeff6857bb9aa557d31d15b11ab6df"
Fetching all precompiled tagged releases..
Downloading: E:/emsdk-master/llvm-tags-32bit.txt from https://s3.amazonaws.com/mozilla-games/emscripten/packages/llvm/tag/win_32bit/index.txt
Downloading: E:/emsdk-master/llvm-tags-64bit.txt from https://s3.amazonaws.com/mozilla-games/emscripten/packages/llvm/tag/win_64bit/index.txt, 2435 Bytes
Error: No tool or SDK found by name 'sdk-releases-fastcomp-7b4b328af02eafbc857b8ca1e3d9b12dddc56ef7-64bit'.

E:\emsdk-master>
```

Is this a problem with the installer or something on my side?

Edit: I tried it on a Linux VM (Mint) where it works without problems.

allrightzhu:
I recommend you get an exe file from this url:https://github.com/emscripten-core/emsdk,just after the exe file installed then you can use emsdk.

fibbo:
@allrightzhu Do you mean the old archived releases at the bottom? I need a current version.

allrightzhu:
> @allrightzhu Do you mean the old archived releases at the bottom? I need a current version.

Oh，I'am sorry，I tried the latest version and got the same problem with you. So I got an old archived releases at the bottom.

fibbo:
The issue is the platform detection. On my computer 

```
import platform
print(platform.machine())
```

returns `AMD64`

[...]

PR #314, opened 2019-08-06
Compare against lower-case platform string

fibbo:
At least on some platforms (my windows e.g.)

```
import platform
print(platform.machine())
```

returns `AMD64` with uppercase letters. This will then fail the platform check and the installation fails.

See https://github.com/emscripten-core/emsdk/issues/312

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -83,7 +83,7 @@ UNIX = (OSX or LINUX)
 
 ARCH = 'unknown'
 machine = platform.machine()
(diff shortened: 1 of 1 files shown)


kripken:
Thanks @fibbo! There is actually a PR already open for this, though, #310, which I landed just now, so I think we should close this one.

fibbo:
Awesome, thanks for the fix!

PR #316, opened 2019-08-06
List all archive suffixes properly, so we clean up all downloaded archives

kripken:
Implements what @juj suggested here: https://github.com/emscripten-core/emsdk/pull/274#issuecomment-502447670

After this, the behavior should consistently be: archives are downloaded, unpacked, then deleted immediately. Another install of the same thing will re-download.


juj:
lgtm!

PR #325, opened 2019-08-21
Fix file permissions when extracting zip files

sbc100:
This means that `./emsdk` works on UNIX system after emsdk self-updates
from a zip file.  Without this one would need to run `python ./emsdk`
which seems to be why the tests were doing it this way.

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -503,6 +503,10 @@ def unzip(source_filename, dest_dir, unpack_even_if_exists=False):
       # Now do the actual decompress.
       for member in zf.infolist():
         zf.extract(member, fix_potentially_long_windows_pathname(unzip_to_dir))
+        unix_attributes = member.external_attr >> 16
+        if unix_attributes:
+          target = os.path.join(unzip_to_dir, member.filename)
+          os.chmod(target, unix_attributes)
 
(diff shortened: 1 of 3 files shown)


emsdk
```diff
@@ -503,6 +503,10 @@ def unzip(source_filename, dest_dir, unpack_even_if_exists=False):
       # Now do the actual decompress.
       for member in zf.infolist():
         zf.extract(member, fix_potentially_long_windows_pathname(unzip_to_dir))
+        unix_attributes = member.external_attr >> 16
```
kripken: i'm not familiar with this - why `16`? is there a named constant for it? (seems not from the docs, but i'm unsure how you know what to do, and whether this is consistent between python versions?)

sbc100: Sorry, its not well documented.  I linked to the stack-overload discussion in the comment.

It does seems like a real shame the "extract" method doesn't already do this.  Sad.   I'm temped to say that we could just call "unzip", at least on unix systems, but this change at least fixes a real issue.

sbc100: Without this fix you can't run "./emsdk" after the update becasue its not executable anymore.  You are forced to run "python ./emsdk".

This change fixes that problem.

PR #341, opened 2019-08-29
reland: Update default node version to 10.16.3

sbc100:
Original change that got reverted: #331

See https://github.com/emscripten-core/emscripten/issues/9332


sbc100:
I had a typo in the original version of this cl.. that would have been found by the CI that we now have in place :)

PR #345, opened 2019-08-30
Temporarily disable some mac testing while we wait for a new emscripten release

sbc100:


sbc100:
No need for this thanks for #346

PR #349, opened 2019-09-03
Switch to python3 when installing python on windows installs.

sbc100:


sbc100:
Ugg.. found the strange python bug that seems to be blocking us: https://bugs.python.org/issue34841.

Looks like we can modify the python zip files in place and remove the .pth files..

sbc100:
Agreed. I'm just not sure where the most obvious place to document this stuff is.  The manifest itsefl if json so can't contain comments.   Maybe just the README?  Or some kind of new contributing document?

For the record all I did was run the following command on each zip file:

```
zip -d python-3.7.4-embed-<arch>.zip python37._pth
```

All this does is remove the .pth file.

sbc100:
I'm going to add some docs about updating node and python as a followup.

kripken:
We could support comments in the manifest file by stripping them out before parsing it as json. But a separate file might be fine too.

Issue #176, opened 2018-10-17
[macOS] urlopen error SSL: TLSV1_ALERT_PROTOCOL_VERSION


gubatron:
```
$ ./emsdk install latest
Done installing tool 'node-8.9.1-64bit'.
Installing tool 'emscripten-1.38.13'..
Error downloading URL 'https://github.com/kripken/emscripten/archive/1.38.13.tar.gz': <urlopen error [SSL: TLSV1_ALERT_PROTOCOL_VERSION] tlsv1 alert protocol version (_ssl.c:590)>
Installation failed!
```

Last commit:
```
commit 0d8576c0e8f5ee09a36120b9d44184b5da2f2e7a (HEAD -> master, origin/master, origin/HEAD)
Author: Jukka Jylänki <jujjyl@gmail.com>
Date:   Sun Aug 19 16:39:17 2018 +0300

    Micro-cleanup TTY_OUTPUT global declaration
```

```
uname -a
Darwin Angels-MacBook-Pro.local 18.0.0 Darwin Kernel Version 18.0.0: Wed Aug 22 20:13:40 PDT 2018; root:xnu-4903.201.2~1/RELEASE_X86_64 x86_64
```

```
Python 2.7.11 (v2.7.11:6d1b6a68f775, Dec  5 2015, 12:54:16) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
```

antonio-gomez:
Exactly same issue here.

gubatron:
ha, I tried doing a manual download but failed, will try again

gubatron:
that worked, I guess I was untaring the file but had to rename the folder to just the version number, thank you Antonio.

felipeplets:
Same here. @antonio-gomez solution worked for me.

Issue #140, opened 2018-04-24
"emsdk install latest" fails when downloading something due to SSL stuff


damu:
So I wanted to try Webassembly and followed the instructions at https://github.com/juj/emsdk.
First emsdk failed because I had no Python, so I installed Python (3.6.5).
"emsdk update" seemed to work (no output) but "emsdk install latest" failed, output:
```
C:\Users\damu\Downloads\emsdk-portable>emsdk install latest
Update complete, however skipped fetching the Emscripten tags, since git was not found.
If you want to compile one of the tagged releases from source, please install git by typing 'emsdk install git-1.9.4', o
r alternatively by installing it manually from http://git-scm.com/downloads . If you install git manually, remember to a
dd it to PATH.
If you are not looking to build Emscripten from source, you can safely ignore this message.
Installing SDK 'sdk-1.37.1-64bit'..
Installing tool 'clang-e1.37.1-64bit'..
The contents of file 'llvm/tag/win_64bit/emscripten-llvm-e1.37.1.zip' already exist in destination 'C:/Users/damu/Downl
oads/emsdk-portable/clang/e1.37.1_64bit', skipping.
Done installing tool 'clang-e1.37.1-64bit'.
Installing tool 'node-4.1.1-64bit'..
The contents of file 'node_4.1.1_64bit.zip' already exist in destination 'C:/Users/damu/Downloads/emsdk-portable/node/4
.1.1_64bit', skipping.
Done installing tool 'node-4.1.1-64bit'.
Installing tool 'python-2.7.5.3-64bit'..
The contents of file 'python_2.7.5.3_64bit.zip' already exist in destination 'C:/Users/damu/Downloads/emsdk-portable/py
thon/2.7.5.3_64bit', skipping.
Done installing tool 'python-2.7.5.3-64bit'.
Installing tool 'emscripten-1.37.1'..
Traceback (most recent call last):
  File "C:\Users\damu\Downloads\emsdk-portable\\emsdk", line 2149, in <module>
    sys.exit(main())
  File "C:\Users\damu\Downloads\emsdk-portable\\emsdk", line 2132, in main
    success = tool.install()
  File "C:\Users\damu\Downloads\emsdk-portable\\emsdk", line 1229, in install
    success = tool.install()
  File "C:\Users\damu\Downloads\emsdk-portable\\emsdk", line 1244, in install
    success = download_and_unzip(url, self.installation_path(), download_even_if_exists, filename_prefix=self.zipfile_pr
[...]

Issue #136, opened 2018-03-01
Ssl install error


Mcilie:
Fetching all precompiled Nightly versions..
Error downloading URL 'https://s3.amazonaws.com/mozilla-games/emscripten/packages/llvm/nightly/osx_32bit/index.txt': <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)>

When running ./emsdk install latest

Why doesnt this work?

kripken:
See https://github.com/kripken/emscripten/issues/6275 , details and work on a fix are mentioned there.

Mcilie:
Ok fixed. So because i have 3 python installs, My system uses the most awful one (python3.4) and so i just ran emsdk by running sudo python3.6 emsdk (command). I did not know emsdk was built in python lol :D

Issue #133, opened 2018-02-23
SSL error 1407742E during emsdk install --latest


Inreal:
Running on Windows 10. The commands and installation routine worked for the same installation package a few days ago. Now it gives the following SSL error:

```Installing SDK 'sdk-1.37.9-64bit'..
Installing tool 'clang-e1.37.9-64bit'..
The contents of file 'llvm/tag/win_64bit/emscripten-llvm-e1.37.9.zip' already exist in destination 'C:/Emscripten/clang/e1.37.9_64bit', skipping.
Done installing tool 'clang-e1.37.9-64bit'.
Installing tool 'node-4.1.1-64bit'..
The contents of file 'node_4.1.1_64bit.zip' already exist in destination 'C:/Emscripten/node/4.1.1_64bit', skipping.
Done installing tool 'node-4.1.1-64bit'.
Installing tool 'python-2.7.5.3-64bit'..
The contents of file 'python_2.7.5.3_64bit.zip' already exist in destination 'C:/Emscripten/python/2.7.5.3_64bit', skipping.
Done installing tool 'python-2.7.5.3-64bit'.
Installing tool 'java-7.45-64bit'..
The contents of file 'portable_jre_7_update_45_64bit.zip' already exist in destination 'C:/Emscripten/java/7.45_64bit', skipping.
Done installing tool 'java-7.45-64bit'.
Installing tool 'emscripten-1.37.9'..
Traceback (most recent call last):
  File "C:\Emscripten\\emsdk", line 2149, in <module>
    sys.exit(main())
  File "C:\Emscripten\\emsdk", line 2132, in main
    success = tool.install()
  File "C:\Emscripten\\emsdk", line 1229, in install
    success = tool.install()
  File "C:\Emscripten\\emsdk", line 1244, in install
    success = download_and_unzip(url, self.installation_path(), download_even_if_exists, filename_prefix=self.zipfile_prefix if hasattr(self, 'zipfile_prefix') else '')
  File "C:\Emscripten\\emsdk", line 850, in download_and_unzip
    dst_file = download_file(urljoin(emsdk_packages_url, zipfile), 'zips/', download_even_if_exists, filename_prefix)
  File "C:\Emscripten\\emsdk", line 416, in download_file
    u = urlopen(url)
  File "C:\Emscripten\python\2.7.5.3_64bit\lib\urllib2.py", line 127, in urlopen
    return _opener.open(url, data, timeout)
[...]

PR #354, opened 2019-09-17
Modified the code and added user notification about Python certificat…

Hillsie:
Modified the code and added user notification about Python certificates issue to reduce the number of future git issues being raised about the Python ... urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate  ... exception. This issue has resulted in continuous stream of issues with common  problems in: #133 #136 #140 #176 #6275 #6548 #6723 #9036 ...

A notification during installation would help the user resolve the issue without raising a new issue request.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -664,7 +664,9 @@ def download_file(url, dstpath, download_even_if_exists=False, filename_prefix='
         print(']')
         sys.stdout.flush()
   except Exception as e:
-    print("Error downloading URL '" + url + "': " + str(e))
+    print("Error: Downloading URL '" + url + "': " + str(e))
+    if "SSL: CERTIFICATE_VERIFY_FAILED" in str(e) or "urlopen error unknown url type: https" in str(e):
+      print("Warning: Possibly SSL/TLS issue. Update or install Python SSL root certificates (2048-bit or greater) supplied in Python folder or https://pypi.org/project/certifi/ and try again.")
     rmfile(file_name)
     return None
   except KeyboardInterrupt:


emsdk.py
```diff
@@ -665,6 +665,11 @@ def download_file(url, dstpath, download_even_if_exists=False, filename_prefix='
         sys.stdout.flush()
   except Exception as e:
     print("Error downloading URL '" + url + "': " + str(e))
+    # Modified code to handle common "...[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: ..." error
+    # Error results in continuous issues being raised for same problem: #133 #136 #140 #176 #6275 #6548 #6723 #9036  ...
+    r = "<urlopen error \[SSL: CERTIFICATE_VERIFY_FAILED\] certificate verify failed: unable to get local issuer certificate \(_ssl\.c:(\d+)\)>"
```
kripken: Perhaps this doesn't need to be so precise? Just searching for `SSL: CERTIFICATE_VERIFY_FAILED` may be good enough, and simpler. Can just do a python string check `pattern in str(e)` for that.

emsdk.py
```diff
@@ -665,6 +665,11 @@ def download_file(url, dstpath, download_even_if_exists=False, filename_prefix='
...
+    if re.match(r, str(e)):
+      print("Possible Python SSL issue. Download Python. Install certificates from Python installation folder or command line and try again")
```
kripken: I think the error message could be a little clearer. Please explain why downloading python can help - I assume because it's a newer version? (so we should suggest downloading the latest one)

Is the sentence with `Install [..]` in addition to the download (a second step), or another option? It's also not clear how to install those, perhaps there's a link we can point to?

Hillsie: For prosperity sake,  this is probably the reason the SSL issue is occuring on MacOS. https://bugs.python.org/issue28150#msg276516

emsdk.py
Hillsie: All done. See updatedcode

emsdk.py
Hillsie: Updated, see new commit

emsdk.py
```diff
@@ -665,6 +665,11 @@ def download_file(url, dstpath, download_even_if_exists=False, filename_prefix='
         sys.stdout.flush()
   except Exception as e:
     print("Error downloading URL '" + url + "': " + str(e))
+    # Modified code to inform solution for common "...[SSL: CERTIFICATE_VERIFY_FAILED] ..." error
+    # Error results in issues raised for similar problem: #133 #136 #140 #176 #6275 #6548 #6723 #9036  ...
+    r = "SSL: CERTIFICATE_VERIFY_FAILED"
+    if re.match(r, str(e)):
```
kripken: these two lines can be just `if "SSL: CERTIFICATE_VERIFY_FAILED" in str(e):`

Hillsie: @kripken, great hint, sure and much cleaner. Yes...been a little while since I've Pythoned with conviction. It can be such a lovely concise language. WIP as I've come across the same error on a Linux machine, so have the opportunity to work on a live error that hasn't been fixed.

Hillsie: @kripken The issue has a different message on linux. I have included a similar re for the message and updated the message.  I'll rebase based on changes from yesterday and squash the pull request.

Hillsie:
@kripken Updated and ready for review

emsdk.py
```diff
@@ -665,6 +665,11 @@ def download_file(url, dstpath, download_even_if_exists=False, filename_prefix='
         sys.stdout.flush()
   except Exception as e:
     print("Error downloading URL '" + url + "': " + str(e))
+    # Modified code to inform solution for common "...[SSL: CERTIFICATE_VERIFY_FAILED] ..." on Mac and
```
kripken: Can remove "modified code", as that makes more sense in a comment message, I think, than the text that remains in the code forever. It could start with "Hint at a solution for the common"

emsdk.py
```diff
@@ -665,6 +665,11 @@ def download_file(url, dstpath, download_even_if_exists=False, filename_prefix='
...
+    # "<urlopen error unknown url type: https>" on Linux
+    # Error results in issues raised for similar problem: #133 #136 #140 #176 #6275 #6548 #6723 #9036  ...
+    if "SSL: CERTIFICATE_VERIFY_FAILED" or "urlopen error unknown url type: https" in str(e):
```
kripken: This is wrong, I think: it will be interpreted as `if "a" or ("b" in c)`. I think we need `if "a" in c or "b" in c`.

Hillsie: Oh my, of course. Haven't resolved the Linux version, so might also change the link once I have it worked out on Linux.

Hillsie:
@kripken I think I have gotten to the bottom of what’s happening. It has taken some time and thinking through the problem.

Here’s my view, sorry a bit long, but necessary to understand the conclusion

## Background

The recommended installation on OS X/Linux/Unix is to use `./emsdk install latest` which runs a shell command and checks the python version and calls `emsdk.py` from the shell script.

[https://webassembly.org/getting-started/developers-guide/](https://webassembly.org/getting-started/developers-guide/)

## Issue - OS X/Linux/Unix

 During installation there is an install issue 

 `SSL: certificates_verify_failed` on Mac and  
`urlopen error unknown url type: https` on Linux/Unix

It’s  the storage locations rejecting the Python TLS certificate and Python error notifying that the storage locations of WebAssembly downloads do not accept 1024-bit TLS certificates.

## Current recommended solutions - Upgrade Certs  however,  nil impact

Current solutions recommended upgrading  the Python certificates to 2048-bit or greater

- On Mac a way to install the newer TLS is to download the latest Python version (currently  3.7.4) and run `open /Applications/python\3.x/Install\ Certificates.command`  in the terminal for x version

-  Another suggestion is to install certificates via `pip install certifi`, which upgrades the certificates for Mac/Linux/Unix

However, running `./emsdk install latest` results in the same error

## The shell script on OS X/ Linux/Unix `./emsdk`

There appears to be confusings issues with certificates and Python version being used. The complication is the shell script during the installation process calls a specific Python version.

- The shell script executes the shell command `which python3`.  What if it’s higher than 3, or you have skipped version 3, as in my case I have 3, 3.6 and 3.7 installed. In this case the shell script only uses python3 or python2 and certs are installed under a higher version and located in a different location.

## Some Work Arounds

After installing the new certs with `pip install certifi`

- Change the version of python used in the shell script  to the most currently installed version is one … for me. For instance, change this line to use python3.7 instead of python3
`python=$(which python3.7 2> /dev/null)` 
And the download works. Probably not a robust solution.

- `./emsdk.py install latest` is a second way. Relies on the Python shebang line. I think this is a more robust option.

- `python3.x ./emsdk.py install latest` means the user can do it with any version of python they choose



## Begs the question

Based on above,  why is a shell script required?

`#!/usr/bin/env python` is the python shebang line used in the `emdsk.py` that executes the latest installed version of python based on the symbolic links `/usr/bin/`

`./emsdk.py install latest` works and negates the need for a shell script.


## Over to you
IMHO, the installation process might be an obstacle to adoption.

I don’t know the history behind all the other installation scripts such as ` *.bat, *.sh, *.ps1, *.fish`. However, perhaps the Python shebang line should be trusted to work on each flavour of the installation and more emphasis placed on the Python scripting? Is this possible? Prolly needs a revisit.

I’ll raise a pull request for the change I’ve made, but the impact to the TLS rejection error and issues raised around it is prolly going to be limited as there seem to be a number of issue variables.

Hillsie:
@kripken Updated and ready for review.

emsdk.py
kripken: I think this has not yet been addressed?

kripken:
(I think there's still one review comment left to be addressed, see above.)

Thanks for the detailed writeup!

This is definitely not easy to fix. Not sure what's best here, but reading [this comment](https://github.com/emscripten-core/emscripten/issues/6275#issuecomment-535771643) I wonder if maybe we should have an `--ignore-cert` or something like that, to let users ignore this error (at their own risk, obviously)?

emsdk.py
Hillsie: > Can remove "modified code", as that makes more sense in a comment message, I think, than the text that remains in the code forever. It could start with "Hint at a solution for the common"

Removed in the latest pull request.

Hillsie:
@kripken I've updated the code and removed the comment. https://github.com/emscripten-core/emsdk/pull/354#discussion_r330315582 

Ready for review.

Hillsie:
> This is definitely not easy to fix. Not sure what's best here, but reading [this comment](https://github.com/emscripten-core/emscripten/issues/6275#issuecomment-535771643) I wonder if maybe we should have an `--ignore-cert` or something like that, to let users ignore this error (at their own risk, obviously)?

Not entirely sure if an ignore-cert  is the way to go. This is not exclusively a WebAssembly problem.  I guess we'd need to answer the question, why have TLS? The core issue is that the 1024-bit certificates are no longer trusted. The installation of 2048-bit upgrade Python has implemented, doesn't impact all locally installed versions, so the symbolic Python linked versions seem to pick the old TLS certificate.

Personally, I'd prefer to maintain confidence in the WebAssembly download and not bypass the certificate. This is probably important for the entire community. A good writeup would help resolve the confusion.

kripken:
Thanks @Hillsie, yeah, I think those are good points.

This should improve things for users a lot. If there's a good place to write more docs we can do that separately later.

ziocleto:
Hi, all of a sudden I had the same problem, I was doing some work in a Linux environment and I was hit by the lethal **SSL certificate madness**. I've followed every possible step of installing **certifi** from pip, pip3, making sure all the pythons have the correct 2048 root cert but nothing...

Still broked.

Until...

I've changed the script to include this magic line:

from:
`urlopen(url)`
to:
`urlopen(url, context=ssl.create_default_context(cafile=certifi.where()))`

Now I can run  

> ./emsdk install latest

again! Now is that magic `create_default_context` a proper fix? I have no idea what that is! :smile:

sbc100:
Can you tell me how you reproduced this on linux?  Which version linux and which version of python were you able to reproduce this on?

ziocleto:
> Can you tell me how you reproduced this on linux? Which version linux and which version of python were you able to reproduce this on?

Sure, I have been working on `Linux Mint 19.3, kernel 5.3`, this is the complete machine info: 

[linux SSL madness machine](https://termbin.com/nq22)

I'm not familiar with python but I've been doing all sorts of work in every possible computer language so I ended up with, strangely, and I don't know why, a python3.7 python3.6, python3.6m, python2.7, python2, python all installed at the same time. I've ran the `pip install certifi` for every python, pip and pip3. I've checked manually and the certificates are all there for every version. 

It was after a lot of poking around with all sort of other devs that I came up with the deaded SSL MADNESS, so unfortunately I don't have a dockerized example to reproduce it.

The thing I've noticed from that log in the link it's that it says "_unknown python3.6 version_".

Just to recap I had the error following step by step the instruction guidelines provided here:

[Emscripten install instructions](https://emscripten.org/docs/getting_started/downloads.html)

PR #357, opened 2019-09-21
Install python3 for mac

### LINKED ISSUE #356: Build broken on MacOS
See https://circleci.com/gh/emscripten-core/emsdk/287?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

@sbc100 [your PR](https://github.com/emscripten-core/emsdk/pull/353) where this started had a green build before landing, but became red after landing. I re-ran the build now and it's still red, so it doesn't look like a random error. But I'm not sure if it's related to that PR or not, what do you think?

kripken:
This appears to fix #356.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,6 +50,9 @@ jobs:
       - run:
           name: Install cmake
           command: brew install cmake
+      - run:
+          name: Install python3
+          command: brew install python3
       - run: ./test.sh
       - run: ./test.py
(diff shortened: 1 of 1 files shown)


sbc100:
I think https://github.com/WebAssembly/binaryen/pull/2349 is better for this case.  Since we do actually want to support OSX users who use the default python (for now)

kripken:
Sounds good.

PR #360, opened 2019-09-23
Remove some mozilla-games remnants

kripken:
Remove obsolete docs links.

Remove downloading of the old releases list, and bundle the file in the repo.

Also the message in update-tags was wrong: we need git to do anything there now, as update-tags gets emscripten-releases using git currently (which is ok as update-tags is run by users that installed with git; otherwise they installed without git and just update the entire emsdk normally).


kripken:
Mac error looks like the known issue that will be fixed by https://github.com/WebAssembly/binaryen/pull/2349

PR #362, opened 2019-09-30
Cleanup emsdk.py

sbc100:
- Remove some unneeded use of `global` keyword
- Use OrderedDict
- Add exit_with_error helper
- Remove SPIDERMONKY_ENGINE from emscripten config

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -3,7 +3,6 @@ ignore =
    E111, # Indentation is not a multiple of four
    E114, # Indentation is not a multiple of four (comment)
    E501, # Line too long
-   E261, # At least two spaces before inline comment
    E121, # Continuation line under-indented for hanging indent
    E722  # bare excepts
 exclude =
--- a/emsdk.py
+++ b/emsdk.py
@@ -7,6 +7,7 @@
 from __future__ import print_function
 
 import copy
+from collections import OrderedDict
 import errno
 import json
 import multiprocessing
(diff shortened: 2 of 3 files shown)


kripken:
Emscripten CI has broken ([ex1](https://circleci.com/gh/emscripten-core/emscripten/189042?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), [ex2](https://circleci.com/gh/emscripten-core/emscripten/189034?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)). The breakage looks unrelated to anything on emscripten, and may be due to the `.emscripten` file and how it is used in the test suite.

This PR landed around the relevant time, and has some changes to `.emscripten`, so it's my only guess atm as to what could have changed things. @sbc100 is that possible?

cc @jgravelle-google who may already be investigating the breakage.

kripken:
https://github.com/emscripten-core/emscripten/compare/fix2?expand=1 seems to confirm that this commit is the cause, as it uses the previous commit from the emsdk and gets past the `build*` stage error.

sbc100:
Hmm, something to do with JS_ENGINES.. investigating

sbc100:
Looks like it was removing the empty `V8_ENGINE = ''` from the config that was the problem.  I'm updating emscripten to deal with that better in the test code: https://github.com/emscripten-core/emscripten/pull/9597

Also, emscripten's circleci was depending on the `os` being imported in the config file, which I also removed in cases where it wasn't needed :(

PR #363, opened 2019-10-01
Add support for constructing a t?csh environment

nickie:
Now the (undocumented) `construct_env` command receives an optional
`--csh` argument.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -62,6 +62,8 @@ if os.name == 'nt' or (os.getenv('SYSTEMROOT') is not None and 'windows' in os.g
   WINDOWS = True
   ENVPATH_SEPARATOR = ';'
 
+SHELL = 'sh'  # this may change by the construct_env command
+
 MSYS = False
 if os.getenv('MSYSTEM'):
   MSYS = True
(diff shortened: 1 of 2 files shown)


kripken:
Thanks @nickie!

@sbc100 maybe you know about this stuff? (I don't understand why it's needed, but maybe that's just my lack of knowledge here...)

nickie:
> I don't understand why it's needed, but maybe that's just my lack of knowledge here...

Some people (rather few nowadays, admittedly) use [`tcsh`](https://en.wikipedia.org/wiki/Tcsh) as their shell in Linux, which is not POSIX compliant and has different syntax for defining environment variables.  These people should `source emsdk_env.csh` instead of `emsdk_env.sh`, as the latter will not work.  Btw, this will also work for people using [`csh`](https://en.wikipedia.org/wiki/C_shell), if there's any...

A similar support is already present for [`fish`](https://en.wikipedia.org/wiki/Friendly_interactive_shell).

emsdk.py
```diff
@@ -2761,6 +2767,10 @@ def print_tools(t):
 
     return 0
   elif cmd == 'construct_env':
+    if len(sys.argv) > 2 and sys.argv[2] == '--csh':
+      global SHELL
+      SHELL = 'csh'
+      del sys.argv[2]
```
sbc100: It looks like we have `extract_bool_arg` to do this kind of thing already.

nickie:
> Could we not use the same mechanism that POWERSHELL uses? i.e. use `EMSDK_CSH` environment variable to control this?

Right, I will make a new MR using `EMSDK_CSH` if you prefer.

emsdk.py
nickie: > we have `extract_bool_arg` to do this kind of thing already.

Yes, ok, I didn't see it --- I was not looking for it, to be honest.  The right way to do this is `argparse`, if you want my 2 cents...  :-)

PR #368, opened 2019-10-08
Added dockerfiles form trzeci/emscripten-docker

trzecieu:
As a continuation from: https://github.com/emscripten-core/emscripten/issues/8742

This PR provides adopted Dockerfile from https://github.com/trzecieu/emscripten-docker/blob/master/docker/trzeci/emscripten-fastcomp/Dockerfile which supports builds:
  * fastcomp
  * upstream

Features:
* An image can be build with fastcomp and upstream version of Emscripten SDK
* It applies some extra size optimizations (that might be removed in sake of clarity / stability)
* It creates conventional `1000:1000` user `emscripten`
* It creates extra entrypoint that mitigates running as non-root user (Some context: https://trzeci.eu/fixing-permission-of-files-created-from-docker/)
* It has simple-stupid tests performed during building

I understand that some of used solutions might not meet expectations, hence I'm open for suggestions how to improve it. 

CC: @kripken @sbc100


docker/README.md
```diff
@@ -0,0 +1,14 @@
+### Building
+
+This image requires to specify following build arguments:
```
sbc100: Perhaps add some more documentation here?   What is this image?  How does one use it?   How/when does it get updated?

docker/Dockerfile
```diff
@@ -0,0 +1,244 @@
...
+RUN	echo "## Installing CMake" \
+    &&	wget https://cmake.org/files/v3.14/cmake-3.14.3-Linux-x86_64.sh -q \
+    &&	mkdir /opt/cmake \
+    &&	printf "y\nn\n" | sh cmake-3.14.3-Linux-x86_64.sh --prefix=/opt/cmake > /dev/null \
+    &&		rm -fr cmake*.sh /opt/cmake/doc \
+    &&		rm -fr /opt/cmake/bin/cmake-gui \
+    &&		rm -fr /opt/cmake/bin/ccmake \
+    &&		rm -fr /opt/cmake/bin/cpack \
+    &&	ln -s /opt/cmake/bin/cmake /usr/local/bin/cmake \
+    &&	ln -s /opt/cmake/bin/ctest /usr/local/bin/ctest
```
sbc100: Why not just `apt-get install cmake`?

docker/Dockerfile
```diff
@@ -0,0 +1,244 @@
...
+COPY Dockerfile /emsdk_portable/dockerfiles/emscripten-core/emsdk/
+
+LABEL maintainer="kontakt@trzeci.eu" \
+      org.label-schema.name="emscripten" \
+      org.label-schema.description="The official container with Emscripten SDK" \
+      org.label-schema.url="https://emscripten.org" \
+      org.label-schema.vcs-url="https://github.com/emscripten-core/emsdk" \
+      org.label-schema.docker.dockerfile="/docker/Dockerfile"
+
+# ------------------------------------------------------------------------------
```
sbc100: Missing newline.

docker/entrypoint
```diff
@@ -0,0 +1,29 @@
...
+# Export this image specific Environment variables
+# Those variables are important to use dedicated folder for all cache and predefined config file
+export EMSDK=/emsdk_portable
+export EM_DATA=/emsdk_portable/.data
+export EM_CONFIG=/emsdk_portable/.emscripten
+export EM_CACHE=/emsdk_portable/.data/cache
+export EM_PORTS=/emsdk_portable/.data/ports
+
+# Activate Emscripten SDK
+. ${EMSDK}/emsdk_set_env.sh > /dev/null
```
sbc100: No need to pipe to dev/null, this script doesn't print anything.

docker/entrypoint
```diff
@@ -0,0 +1,29 @@
...
+# In case of running as root, use `umask` to reduce problem of file permission on host
+if [ "$(id -g)" = "0" ] && [ "$(id -u)" = "0" ] ;
+then
+    umask 0000
+fi
+
+# Export this image specific Environment variables
+# Those variables are important to use dedicated folder for all cache and predefined config file
+export EMSDK=/emsdk_portable
+export EM_DATA=/emsdk_portable/.data
```
sbc100: EM_DATA is not known to emsdk or emscripten, I don't think we need it here.

docker/entrypoint
```diff
@@ -0,0 +1,29 @@
...
+
+# In case of running as root, use `umask` to reduce problem of file permission on host
+if [ "$(id -g)" = "0" ] && [ "$(id -u)" = "0" ] ;
+then
+    umask 0000
+fi
+
+# Export this image specific Environment variables
+# Those variables are important to use dedicated folder for all cache and predefined config file
+export EMSDK=/emsdk_portable
```
sbc100: Isn't the job emsdk_set_env.sh below to set EMSDK correctly?   I think you can drop this line

docker/Dockerfile
```diff
@@ -0,0 +1,244 @@
...
+    && strip -s `which node` \
+    # Tests consume ~80MB disc space
+    && rm -fr ${EMSDK}/llvm/clang/emscripten/tests \
+    # strip out symbols from clang (~extra 50MB disc space)
+    && find ${EMSDK}/llvm/clang/bin -type f -exec strip -s {} + || true \
+    && find ${EMSDK}/llvm/clang/fastcomp/bin -type f -exec strip -s {} + || true \
+&&  echo "## Done"
+
+# Populate Emscripten SDK cache with libc++, to improve further compilation times.
+RUN echo "## Pre-populate cache" \
```
sbc100: The best way to populate the cache is to run "./embuilder".     Perhaps "./embuilder build ALL`, or `./embuilder build SYSTEM` (the formor might be too slow).

docker/test_dockerimage.sh
```diff
@@ -0,0 +1,18 @@
...
+
+which asm2wasm
+which llvm-ar
+which emsdk
+node --version
+npm --version
+python --version
+pip --version
+python3 --version
+pip3 --version
```
sbc100: Do we need python3, pip3, or pip?  There is no point if having 3 different version of python installed is there?

docker/Dockerfile
```diff
@@ -0,0 +1,244 @@
+FROM debian:stretch AS stage_build
```
sbc100: Why build in stretch and deploy in buster?

trzecieu: No reason here, at some moment I've just switched in the buster in stage_deploy as it was requested. Can be synchronize, but as long there is no dynamic linking we can use any distro as build_stage

docker/Dockerfile
trzecieu: Huh this is a trace from one of the very first implementation of this image (I believe 1.34.4).
Later I was experimenting with fat build (https://github.com/trzecieu/emscripten-docker/blob/master/docker/trzeci/emscripten/Dockerfile#L55) but that solution wasn't good - once is that it has increased build time significantly, another issue was build size that wasn't acceptable anymore. 

To just pre-populate the most frequently used libs I went that way. 

I've added `embuilder.py build SYSTEM` as it looks like a sweet spot.

docker/Dockerfile
trzecieu: To get newer cmake than the one is available for a base distribution: 
```
docker run --rm -it debian:buster bash -c "apt-get update -y && apt-cache policy cmake"
```
```
cmake:
  Installed: (none)
  Candidate: 3.13.4-1
  Version table:
     3.13.4-1 500
        500 http://deb.debian.org/debian buster/main amd64 Packages
```

docker/Dockerfile
trzecieu: done

docker/entrypoint
trzecieu: removed

docker/entrypoint
trzecieu: It is indeed - at some moment I've added that intentionally but I don't remember any more my motivation, will remove and re-test

docker/entrypoint
trzecieu: Removed

docker/test_dockerimage.sh
trzecieu: You mean two versions I believe, as there is no Python installed with Emscripten SDK. 
Motivation to have python3 comes from https://github.com/WebAssembly/binaryen/pull/2281 
About python2 - I don't have strong feelings. Maybe just that it takes a time for projects to migrate, and my intention was to provide a complete container that can be used in a general case. 

I can remove python2 in exchange of documentation how to extend this image.

docker/README.md
trzecieu: Not sure about answer for _How/when does it get updated?_, ideally it should be build and published as post build step after a release of Emscripten gets published. For that I will need some helping hand.

docker/Dockerfile
sbc100: Whats wrong with 3.13.4?  Isn't that a pretty recent version?   Its certainly recent enough to build all the projects related to emscripten.     Maybe this is also left over from when you were building against the older debian version?

docker/README.md
sbc100: For now maybe lets just include instructions on how to update the image?  And how to use it.

sbc100: Maybe a link to the image in the docker registry and how to update it?

docker/Dockerfile
trzecieu: Two major reasons: 
- back in time base image was Jessy, and there was no decently new cmake available there. 
- It was a contribution: https://github.com/trzecieu/emscripten-docker/pull/43 

For future reasons I thought it will be nice to preserve this privileged to be able to have the most recent version available. 

But no strong feelings about this, I see also value of having a simple apt package being installed.

sbc100: I think now that you are using buster it simpler just to use the packaged version.

docker/test_dockerimage.sh
trzecieu: Had to revert that. 
`em++` Still has a header `#!/usr/bin/env python` and I think making `python` pointing to `python3` is a bad way to mitigate this issue.

sbc100: OK.

docker/README.md
```diff
@@ -0,0 +1,83 @@
+### About this Dockerfile
+... to be added
```
sbc100: How about:

```
# Dockerfile for EMSDK 

This Dockerfile builds a self-contained version of emsdk that enables emscripten to be used without any
other installation on the host system.

It is pusblished at https://hub.docker.com/u/emscripten/emsdk.

```

docker/Dockerfile
```diff
@@ -0,0 +1,234 @@
...
+    \
+    &&  echo "## Done"
+
+# ------------------------------------------------------------------------------
+# -------------------------------- STAGE DEPLOY --------------------------------
+# ------------------------------------------------------------------------------
+
+FROM debian:buster-slim AS stage_deploy
+
+COPY --from=stage_build /emsdk_portable /emsdk_portable
```
sbc100: How about we drop the `_portable` and just call this `/emsdk`.. I'm not sure where the portable suffix comes from.

trzecieu: Good point. 
And it came from old time where emsd was distributed as a zip called `emsdk_portable` or `emsdk-portable` (I don't remember right now) and it stuck like that. 
Lots of code in this Dockerimage was transformed over 3y so, lots of legacy traces here.

Will make `/emsdk` :+1:

docker/README.md
trzecieu: Is https://hub.docker.com/u/emscripten the official profile of emscripten-core?

sbc100: As of just recently yes.  I think we just created it yesterday for the emscripten-ci image

sbc100: Not sure if we should call this image `emscripten/emsdk` or `emscripten/emscripten`?  I think maybe the former which this is where the dockerfile has ended up and the image does contain the emsdk?

kripken:
No comments from me, looks reasonable, thanks!

Would it be possible to test this? (as in, test that building this Docker image leads to a working environment?)

Sounds good to add to docker hub. Let me know your usernames there and I'll add you.

sbc100:
I'm sbc100 on docker hub

trzecieu:
I'm `trzeci` on docker hub

docker/README.md
trzecieu: Surprisingly hard question, `emscripten` will be more popular as the name has better recognition. 
If we'd go with a name `emsdk` it would be technically correct to have just `emscripten-core/emsdk` installed without specific emscripten sdk being pulled, where this image is profiled to deliver a concrete sdk version. 
My vote is for `emscripten`, but again. It's your call.

sbc100: OK, lets go with emscripten/emscripten to keep it simple.

docker/README.md
trzecieu: Added helper makefile + extended info how to build, push and tag this image

kripken:
Ok, added you two and a team and a repo, hopefully I've done it all correctly, let me know!

PR #369, opened 2019-10-09
Fix for default NODE_JS in config file

sbc100:
This was broken in #362


sbc100:
I verified locally, yes.

PR #370, opened 2019-10-09
Add support for constructing a t?csh environment

nickie:
This is a different approach, superseding #363, after the comments made by @sbc100 there.


kripken:
Thanks @nickie!

PR #373, opened 2019-10-16
Switch default backend from 'fastcomp' to 'upstream'

sbc100:
When users as for 'latest' or just '1.39.0' we now default to the upstream llvm backend.

For versions before 1.39.0 we continue to default to fastcomp.

Fixes: https://github.com/emscripten-core/emscripten/issues/5488


sbc100:
Can we land this before #373 ?

kripken:
Before we land this we should have docs probably, but I'm not sure where?

kripken:
Actually, maybe we could have the emsdk say something? Like show a message, "the default backend has changed" on first run after the change?

sbc100:
Good idea.  Added a bit notice.

kripken:
Looks good, thanks!

kripken:
In offline discussion we had the idea to maybe not always default to upstream, but just from 1.39.0 and onwards. The idea being that say at a version from last year, upstream's support was not that great yet, so it would be bad. Only at 1.39.0 are we confident the new backend is good enough for general use.

Let's discuss more and decide next week.

sbc100:
Updated so that we now only default for upstream for 'latest' and for 1.39.0 and above.

emsdk.py
```diff
@@ -2710,14 +2710,29 @@ def extract_bool_arg(name):
 
   releases_info = load_releases_info()['releases']
 
+  def report_upstream_by_default():
+    print('''\
+** NOTICE **: The default SDK changed from `fastcomp` to `upstream`.
+If you have problems, or wish to revert back to fastcomp for some other reason
+you can add `-fastcomp` to explitly install that fastcomp-based
```
kripken: type in `explicitly`

emsdk.py
```diff
@@ -2710,14 +2710,29 @@ def extract_bool_arg(name):
...
+SDK. .e.g ./emsdk install latest-fastcomp.
```
kripken: comma after `SDK` instead of period, and no period before `e.g.`

emsdk.py
```diff
@@ -2710,14 +2710,29 @@ def extract_bool_arg(name):
...
+''', file=sys.stderr)
+
   # Replace meta-packages with the real package names.
   if cmd in ('update', 'install', 'activate'):
     for i in range(2, len(sys.argv)):
       arg = sys.argv[i]
-      if arg in ('latest', 'sdk-latest', 'latest-64bit', 'sdk-latest-64bit', 'latest-fastcomp', 'latest-releases-fastcomp'):
+      if arg in ('latest', 'sdk-latest', 'latest-64bit', 'sdk-latest-64bit'):
+        # This is effectly the default SDK
+        report_upstream_by_default()
```
kripken: I think we can report this only if the version is high enough? Can maybe share code with the existing check later down.

sbc100: The version check make sense below, but here we know that we want to unconditionally select `upstream` right?  We never want latest to point to fastcomp ever again, do we?  And if we do we would revert this code.

aheejin:
🎉🎉🎉

PR #376, opened 2019-10-18
Add python and java to 'releases-upstream' for windows

sbc100:
These are part of `releases-fastcomp` but for some reason are not
included in the upstream equivalent.  I imagine they were simply
overlooked.

### DIFF
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -1679,7 +1679,7 @@
   {
     "version": "releases-upstream-%releases-tag%",
     "bitness": 64,
-    "uses": ["releases-upstream-%releases-tag%-64bit", "node-12.9.1-64bit"],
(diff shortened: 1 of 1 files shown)


kripken:
Hmm, I think this makes sense, however, in the current model this means we'll re-download python and java on each version update. We also do that for node, so this would be tripling or more that wasted overhead.

Maybe we can fix that re-download issue before landing this (or soon after, if this is urgent)? I think we may want to add an option to manifest entries to say "keep the zip around, and if it's around, don't re-download", which we'd use for node, python, java, but not for the waterfall builds (where we want to delete the archive after download, to avoid them accumulating over time).

sbc100:
Yes, this change just bring upsteam in line with fastcomp.   I think we should land this now.

As for caching tar files around, I think we should go back to doing that yes.  We can find some say the clean them up over time.  Perhaps just keep the last N, by filesytem date?  That's a separate issue though.

sbc100:
Can we land this before #373 ?

PR #390, opened 2019-11-21
Update README

sbc100:
- Remove references to non-existent `Archived Releases` section.
- Rename `Platform-Specific Notes` to `System Requirements`
- Put `SDK Maintenance` which is kind of like an FAQ at the end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-# Emscripten SDK
+Emscripten SDK
+==============
 
 [![CircleCI](https://circleci.com/gh/emscripten-core/emsdk/tree/master.svg?style=svg)](https://circleci.com/gh/emscripten-core/emsdk/tree/master)
 
@@ -14,8 +15,6 @@ You can also set up Emscripten from source, without the pre-built SDK, see
 
 To get started with Emscripten development, see the [Emscripten website
 documentation](https://emscripten.org/docs/getting_started/downloads.html).
-
-**Old Releases** are available in the **Archived Releases** section below.
 
 ## SDK Concepts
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,15 +39,53 @@ important concepts to help understanding the internals of the SDK:
   accessed through. Most operations are of the form `emsdk command`. To access
   the emsdk script, launch the Emscripten Command Prompt.
 
+## System Requirements
```
kripken: this section duplicates docs on our main site. how about linking to there, so there's a single source of truth?

sbc100: I see.  For building from source source the dependencies are listed in https://emscripten.org/docs/building_from_source/toolchain_what_is_needed.html.  For the the dependencies of the emsdk itself I think this is the only place we document those, no?

README.md
```diff
@@ -40,15 +39,49 @@ important concepts to help understanding the internals of the SDK:
...
+  install one now?" that will automatically download a Java runtime to the
+  system.
+
+### Linux
+
+- `python`: Version 2.7.0 or above.
+- `java`: For running closure compiler (optional)
+
+The emsdk pre-compiled binaries are built aginst Ubuntu/Xenial 16.04 LTS and
+therefore depend on system libraryes compatiable with versions of `glibc` and
```
kripken: typo in `libraries`

README.md
kripken: Yes, we could document the emsdk specific dependencies here, good point. But the only dependency there is python, I think? Can remove java below.

README.md
```diff
@@ -40,15 +39,49 @@ important concepts to help understanding the internals of the SDK:
   accessed through. Most operations are of the form `emsdk command`. To access
   the emsdk script, launch the Emscripten Command Prompt.
 
+## System Requirements
+
+Using the emsdk pre-compiled packages requires only the minimal set of
+dependenencies lists below.  When building from source a wider set of tools
```
kripken: typo in `dependencies`

README.md
kripken: But more generally, I don't think we need a specific section per OS? Just "the emsdk itself requires python, as described in the emscripten docs. The other requirements of emscripten itself are also in those docs."

sbc100: Sure I will do that.  But we do also depend on java, since emsdk doesn't handle install that.   Ideally emsdk users should not need to look at `building_from_source/toolchain_what_is_needed.html` since they are not building from source right?

sbc100: Another way of putting it:  We are document here any emscripten dependencies that emsdk doesn't install automatically.  At least that is how i think of it.

kripken: I think we do install java, don't we? It's in the emsdk_manifest file.

But to the second point, yeah, I think it makes sense to document here any deps emcc needs that the emsdk doesn't provide. I don't think there should be any, though, aside from python (which the emsdk doesn't provide as it needs it itself)?

kripken: Oh, looks like we install java on windows but nowhere else. So yeah, maybe mentioning that java is needed for closure is ok.

PR #394, opened 2019-11-25
Do not attempt to reinstall python if it already exists

juj:
Do not attempt to reinstall python if it already exists, since that python may be interpreting the current emsdk.py script that is being executed to do the installation.

At some point in the past the installation model changed to always redownload+unzip over tools that already have installed. That broke emsdk on Windows, as it would attempt to re-unzip python over to the directory where python was actively being executed, resulting in error

```
Downloading: C:/code/emsdk/zips/WinPython-64bit-2.7.13.1Zero.zip from https://storage.googleapis.com/webassembly/emscripten-releases-builds/deps/WinPython-64bit-2.7.13.1Zero.zip, 41285159 Bytes
Unpacking 'C:/code/emsdk/zips/WinPython-64bit-2.7.13.1Zero.zip' to 'C:/code/emsdk/python/2.7.13.1_64bit'
Unzipping file 'C:/code/emsdk/zips/WinPython-64bit-2.7.13.1Zero.zip' failed due to reason: [Error 5] Access is denied: '\\\\?\\C:\\code\\emsdk\\python\\2.7.13.1_64bit\\python-2.7.13.amd64\\DLLs\\_ctypes.pyd'
Installation failed!
```

Not sure of another way to resolve except to skip python from the "always re-download" behavior. (could check if the current python interpreter actually is the one that is being installed, but that might be a bit overkill and lead to odd discrepancies for users)

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1712,6 +1712,9 @@ class Tool(object):
       # We should not force reinstallation of python if it already exists, since that very python
       # may be interpreting the current emsdk.py script we are executing. On Windows this would
       # lead to a failure to uncompress the python zip file as the python executable files are in use.
+      # TODO: Refactor codebase to avoid needing this kind of special case check by being more
+      # careful about reinstallation, see https://github.com/emscripten-core/emsdk/pull/394#issuecomment-559386468
+      # for a scheme that would work.
       if self.id == 'python' and self.is_installed():
         print("Skipped installing " + self.name + ", already installed.")
         return True
--- a/emsdk.py
+++ b/emsdk.py
@@ -1709,6 +1709,12 @@ class Tool(object):
       print("Done installing SDK '" + str(self) + "'.")
       return True
     else:
+      # We should not force reinstallation of python if it already exists, since that very python
+      # may be interpreting the current emsdk.py script we are executing. On Windows this would
+      # lead to a failure to uncompress the python zip file as the python executable files are in use.
+      if self.id == 'python' and self.is_installed():
+        print("Skipped installing " + self.name + ", already installed.")
+        return True
       print("Installing tool '" + str(self) + "'..")
       url = self.download_url()
 


kripken:
Interesting, but not sure I understand this yet. On CI this isn't broken, and we haven't gotten user reports about breakage either, so maybe there is more to this?

With that said, I agree that sounds like a problem we should fix.

I think the reason we changed to always download things is that we decided that it was too bug-prone to keep around the existing code that checked whether a download was needed. I think that was reasonable, but yeah, looks like it can cause this bug, and also it is a little wasteful.

How about marking some things in the manifest file as "don't re-download", and that would include the python and node and java runtimes, basically? Then this would be more general than just hard-coding python there.

kripken:
Oh, I remember now, the issue was if we try to avoid re-downloading stuff, we'd end up with users accumulating old archives - and they are huge.

But for python, node, java, we don't constantly have new versions, and they are much smaller, so I think we can mark those as "don't re-download".

juj:
I think ideally everything should be "don't re-download", since if you have something installed already, why reinstall?

If the original problem was that installation was aborted in the middle, perhaps emsdk should always rollback a failed installation, i.e. delete the directory it attempted to unzip to (but only partially succeeded), and delete the partial downloaded zip if the download fails?

kripken:
Yes, proper automatic rollback seems necessary for such a system.

But another issue is the accumulating downloads that I mentioned. To avoid re-downloading, we must keep the old archive. So the user ends up with 1.39.0, 1, 2, 3, 4 etc. which each contain hundreds of MB of LLVM and clang binaries. So we'd need to start to manage some kind of automatic cleanup system, which when we discussed it, we decided to just not cache the binaries for now.

If we fix those issues, I think we can avoid re-downloading recent files. However, instead of that complexity, I think making the python, node, etc. downloads have a manifest property that says "keep archive" would be simpler (and avoid the accumulation issue as those are small, and update very rarely).

juj:
Hmm, why do we need to keep the old archive? Can’t it be deleted after installation (whether it succeeded or not)?

kripken:
I think the old algorithm did something like:

 * we want to download and unpack `archive.zip`
 * check if `archive.zip` exists
 * then also check if the directory it would be unpacked into exists

I think you're right that we could avoid saving the archive. But then we'd need a mechanism to know what is in the directory. That is, we unpack all `upstream` builds into `upstream/`, so the existence of the directory doesn't tell us if that's 1.X.Y or 1.Z.W.

(Another option is to unpack into a directory that has the version in the name. Users didn't like that because then they needed to constantly update the directory in their scripts, and also, it would bring us back to the issue of needing to clean up old directories that accumulate and take GB of disk space.)

juj:
> I think the old algorithm did something like:
> 
>     * we want to download and unpack `archive.zip`
> 
>     * check if `archive.zip` exists
> 
>     * then also check if the directory it would be unpacked into exists

That sounds about right. An issue arose if user did ctrl-c in middle of archive.zip download, or archive.zip uncompression, as emsdk did not have a way to roll back partial download/unzip. I think a way to resolve that would be to 
1. download the files under a temp name first, and only when the download completes, rename to the correct name. That signals a proper uncorrupted zip,
2. when unzipping, unzip over to a temp directory first. If that succeeds, rename the directory to the right name.

> we unpack all `upstream` builds into `upstream/`, so the existence of the directory doesn't tell us if that's 1.X.Y or 1.Z.W.

Perhaps there could be a file `upstream/version.txt` that emsdk would read to figure out which version already exists there?

kripken:
Yeah, a file that defines what is currently installed is a possible solution. I agree that that + the other things you mentioned (renaming the dir etc.) is a good plan.

juj:
Ok. I cannot work on that kind of refactor atm though, I am hoping that this could land in the meanwhile to unblock Windows users.

PR #400, opened 2019-12-05
Rename old fastcomp-based tools to a specific fastcomp- prefixed name

juj:
Rename old fastcomp-based tools to a specific fastcomp- prefixed name to clearly distinguish legacy toolchain


juj:
Putting this up for proposal - this will break a lot of legacy users, but I think it might be worth it, this way emsdk can explicitly highlight what is legacy and give a push to people to move over to wasm backend.

PR #403, opened 2019-12-07
Add Closure compiler to emsdk.

juj:
Also drop Java.

To land this, please upload http://clb.confined.space/dump/google-closure-compiler-20191111.zip to Google emsdk hosting. The zip was generated via instructions by @dcodeIO at https://github.com/emscripten-core/emscripten/pull/9962#issuecomment-562328316 .

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1446,6 +1446,8 @@ class Tool(object):
     if '%generator_prefix%' in str:
       str = str.replace('%generator_prefix%', cmake_generator_prefix())
     str = str.replace('%.exe%', '.exe' if WINDOWS else '')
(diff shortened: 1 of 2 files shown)


sbc100:
What do you think about instead just adding to the package.json file at the top level of emscripten?  Then we can just run `npm install` as a post install step and it will pull down whatever we need?  This puts emscripten in control of its own dependencies.

And for non-emsdk-users, we just say run `npm install` ..

juj:
That would work as well. I'll re-do the PR in that form tomorrow.

juj:
Superceded by #404.

PR #407, opened 2019-12-10
Restore caching of downloads

sbc100:
Download caching was removed in #324 and other PRs in order to avoid
unboundedly filling the user disk.  However this meant that no caching
at all was done and all archives were downloaded and extracted on every
`install` command.  Even things like node which never really changed
were downloaded *and* extracted each time.

This change re-instates archive caching, and enables skipping both
downloading and unpacking when possible.  The only time we don't want to
skip then unpacking phase is for `releases` SDKs since they always
unpack to the same directory, clobbering each other.

Unbounded cache growth is implemented via only keeping around the most
recent N archive in the cache directory.

Also, when unpacking and archive, remove any existing directory first.
Without this, files from one `releases` archive can end up lingering
within the tree of another.


juj:
Hmm, the intent was never to cache downloaded zip files to user's hard disk. The zip file is no longer needed after one installs a tool.

The original idea was that if one emsdk installs a tool, then reinstalling that tool is a no-op, since the tool is already installed, there is nothing to do.

But there were bugs about rollback cleanup, e.g. if user ctrl-c's in the middle of unzip, or download aborts in the middle of transfer, the installation process was left in bad state, and emsdk was unable to detect that.

Instead of caching, it would be better to improve the rollback cleanup (if download fails, delete any partial file ; if unzip fails, delete any partial unzipped directories) so that emdsk could reinstate the "emsdk install of an installed tool is a no-op" behavior?

sbc100:
Ok will try that approach.

PR #422, opened 2020-01-10
Remove EMSDK_NODE environment variable

sbc100:
There are no users of this env var.

### DIFF
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -192,8 +192,7 @@
     "windows_url": "node_4.1.1_32bit.zip",
     "linux_url": "node-v4.1.1-linux-x86.tar.gz",
     "activated_path": "%installation_dir%/bin",
-    "activated_cfg": "NODE_JS='%installation_dir%/bin/node%.exe%'",
-    "activated_env": "EMSDK_NODE=%installation_dir%/bin/node%.exe%"
+    "activated_cfg": "NODE_JS='%installation_dir%/bin/node%.exe%'"
   },
   {
     "id": "node",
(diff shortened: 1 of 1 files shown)


juj:
This is exposed for external users of EMSDK, not for EMSDK itself to consume, same goes for EMSDK_PYTHON. Not sure who asked for this, but I see that we use this in our Unity build system as well to locate the Node.js that is used.

sbc100:
Given that we add both node and python to the path wouldn't it be better to simply use the PATH?

I'd rather not have external tools depending on emsdk specificially becasue then those tools won't work for non-emsdk users.  For example I'm in discussion to with qtcreator on who to avoid depending on emsdk specifically, but instead depending on the PATH alone, which will work for all emscripten users.

Would you be OK with depending on the PATH instead?     Having multiple ways to specify things, IMHO, only complicates matters and leads to confusion.

juj:
> I'd rather not have external tools depending on emsdk specificially becasue then those tools won't work for non-emsdk users.

This is not a use case e.g. for Unity build system, which is always bootstrapped via emsdk.

> Would you be OK with depending on the PATH instead?

No. Depending on PATH is brittle because one cannot have control over what users have in their PATH, and it requires subprocess spawns that expand via shell, or manual path searches across the toolchain that take CMDSPEC ordering rules into account. Using EMSDK_NODE and EMSDK_PYTHON is a straightforward solution to the problem.

sbc100:
> > I'd rather not have external tools depending on emsdk specificially becasue then those tools won't work for non-emsdk users.
> 
> This is not a use case e.g. for Unity build system, which is always bootstrapped via emsdk.

I don't think we want to encourage people to build tools that work only with emsdk and not with a direct emscripten checkout?

> 
> > Would you be OK with depending on the PATH instead?
> 
> No. Depending on PATH is brittle because one cannot have control over what users have in their PATH,

emsdk already manipulates the users path though and injects the node path, so its *does* have control of the user path.   For better or worse.

> and it requires subprocess spawns that expand via shell, or manual path searches across the toolchain that take CMDSPEC ordering rules into account. Using EMSDK_NODE and EMSDK_PYTHON is a straightforward solution to the problem.

I don't think searching the PATH for a tool is complicated.  Sure it requires little more code, but  really its not much to ask.  But I won't push this too hard if you need to for unity.  I guess I will abandon this change for now.

I do worry about qtcreator though.. they already started depending on these things thinking that all emscripten users would provide them.   And there should be no need for the specific emsdk dependency when all the tools they need should be in the PATH.

PR #426, opened 2020-01-20
Pass `--no-audit` to `npm install`

sbc100:
Fixes https://github.com/emscripten-core/emscripten/issues/10246


sbc100:
I think this code only ever uses the node we ship which is always.

Specifically if find npm via:

```
node_path = os.path.join(node_tool.installation_path(), 'bin')                 
npm = os.path.join(node_path, 'npm' + ('.cmd' if WINDOWS else ''))  
```

Also, we only trigger the `emscripten_npm_install` post install step in the json file for recent versions. 

So AFAICT there should not be any chance of getting an old version of node here.

PR #431, opened 2020-01-28
Hide npm output by default

sbc100:
Fixes: https://github.com/emscripten-core/emscripten/issues/10291

Also fix issue where an old/incorrect version of node could be used during SDK install
since we were preferring activated versions of node.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1276,7 +1276,17 @@ def emscripten_npm_install(tool, directory):
   npm = os.path.join(node_path, 'npm' + ('.cmd' if WINDOWS else ''))
   env = os.environ.copy()
   env["PATH"] = node_path + os.pathsep + env["PATH"]
-  subprocess.check_call([npm, 'ci', '--production'], cwd=directory, env=env)
+  print('Running post-install step: npm ci ...')
+  try:
+    subprocess.check_output(
+        [npm, 'ci', '--production'],
+        cwd=directory, stderr=subprocess.STDOUT, env=env,
+        universal_newlines=True)
+  except subprocess.CalledProcessError as e:
+    print('Error running %s:\n%s' % (e.cmd, e.output))
+    return False
+
+  print('Done running: npm ci')
   return True
 
 
--- a/.flake8
+++ b/.flake8
@@ -9,6 +9,7 @@ exclude =
   ./gnu
   ./upstream
   ./fastcomp
+  ./fastcomp-clang
   ./releases
   ./clang
   ./emscripten


emsdk.py
```diff
@@ -1280,7 +1280,17 @@ def emscripten_npm_install(tool, directory):
   npm = os.path.join(node_path, 'npm' + ('.cmd' if WINDOWS else ''))
   env = os.environ.copy()
   env["PATH"] = node_path + os.pathsep + env["PATH"]
-  subprocess.check_call([npm, 'install', '--production', '--no-audit'], cwd=directory, env=env)
+  print('Running post-install step: npm install ...')
+  try:
+    subprocess.check_output(
+        [npm, 'install', '--production', '--no-audit'],
+        cwd=directory, stderr=subprocess.STDOUT, env=env,
+        universal_newlines=True)
```
kripken: the redirection might break colors in the shell, I think?

emsdk.py
```diff
@@ -1280,7 +1280,17 @@ def emscripten_npm_install(tool, directory):
...
+  except subprocess.CalledProcessError as e:
+    print('Error running %s:\n%s' % (e.cmd, e.output))
+    return False
```
kripken: before this PR the check_call threw, but now we will return false?

emsdk.py
```diff
@@ -1280,7 +1280,17 @@ def emscripten_npm_install(tool, directory):
...
+
+  print('Done npm install')
```
kripken: how about `npm install complete`, `Done running npm install` or `...success`?

emsdk.py
kripken: (if so, it makes me wonder if the redirection is worth it)

emsdk.py
sbc100: Its better to return false.  Makes for cleaner error messages.

emsdk.py
sbc100: It removes colors from the node output yes.   I'm totally fine with that give that most users should never see anything here.

It also disabled the nice progress bars and stuff like that.

emsdk.py
sbc100: Done.

sbc100:
OK with the latest version?

sbc100:
I split this change into two parts.  The npm ci part is now https://github.com/emscripten-core/emsdk/pull/433

emsdk.py
kripken: Was the thrown error caught higher up? Or is the false turned into an error or exit? (Otherwise it used to halt, but no longer does?)

kripken:
Looks like this contains part of the diff of another PR (the active tool stuff)?

emsdk.py
sbc100: Yeah the False return value is handled by the caller as an error.

PR #432, opened 2020-01-28
Extract `install_sdk` and `install_tool` from `install`. NFC

sbc100:
This change simply splits this function into its two logical parts.


sbc100:
View with whitespace disabled to see the actual content of the change.

Issue #434, opened 2020-01-29
No BINARYEN_ROOT in .emscripten for master branch SDKs


floooh:
It seems like an ```emsdk activate sdk-upstream-master-64bit``` doesn't write BINARYEN_ROOT to the .emscripten config file (happens on macOS, but I think it's not OS specific).

This in turn causes compilations issues later when trying to build with this SDK:

```
shared:ERROR: BINARYEN_ROOT is not defined in /Users/floh/projects/fips-sdks/emsdk/.emscripten
```

Also see this thread:

https://groups.google.com/forum/#!topic/emscripten-discuss/fvmSMLFIj5c

I did some python debugging, and this line here:

https://github.com/emscripten-core/emsdk/blob/deaa7e4bfd79246f02f785c8f2174d1e74ab6c00/emsdk.py#L2498

Returns the following "tools_to_activate":

- llvm-git-master-64bit
- node-12.9.1-64bit
- emscripten-master-64bit
- sdk-upstream-master-64bit

None of those have BINARYEN_ROOT in their "activated_cfg":

- llvm-git-master-64bit: **LLVM_ROOT='%installation_dir%/%fastcomp_build_bin_dir%**
- node-12.9.1-64bit: **NODE_JS='%installation_dir%/bin/node%.exe%'**
- emscripten-master-64bit: **EMSCRIPTEN_ROOT='%installation_dir%';EMSCRIPTEN_NATIVE_OPTIMIZER='%installation_dir%%generator_prefix%_64bit_optimizer/%cmake_build_type_on_win%optimizer%.exe%'**
- sdk-upstream-master-64bit: no "activated_cfg" member, because it's not a "tool"

I think some dependency got lost during the incoming -> master transition. I tried adding "binaryen-master-64bit" as dependency to the right SDK entry (since those had been removed in the update), but this gives a warning about binaryen not being installed:

```
Warning: The SDK/tool 'binaryen-master-64bit' cannot be activated since it is not installed! Skipping this tool...
```

(so I think the proper solution is to add BINARYEN_ROOT to ```activated_cfgs``` of one of the 3 tools, e.g. llvm-git or emscripten)

brion:
Possibly related, I'm unable to build `binaryen-master-64bit` via emsdk on macOS Catalina:

```
In file included from /Users/brion/src/emsdk/binaryen/master/src/cfg/Relooper.cpp:17:
[...]

PR #435, opened 2020-01-29
Re-add binaryen dep to upstream-master SDK builds

brion:
This should ensure that BINARYEN_PATH is added to the resulting
~/.emscripten file when building from source:

https://github.com/emscripten-core/emsdk/issues/434

Tested briefly on macOS with `emsdk install sdk-upstream-master-64bit`.


sbc100:
lgtm!  Thanks for looking into this!

PR #437, opened 2020-02-05
Make installation errors fatal

sbc100:
If `is_installed` doesn't return true after install than something
serious went wrong so make this into a warning.

Also, clean up couple of places to make the python code more
readable.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1397,7 +1397,8 @@ def get_required_path(active_tools):
   path_add = [to_native_path(emsdk_path())]
   for tool in active_tools:
     if hasattr(tool, 'activated_path'):
-      path_add += [to_native_path(tool.expand_vars(tool.activated_path))]
+      path = to_native_path(tool.expand_vars(tool.activated_path))
+      path_add.append(path)
   return path_add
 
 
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -2589,22 +2591,24 @@ def adjusted_path(tools_to_activate, log_additions=False, system_path_only=False
...
-  existing_emsdk_tools = [item for item in existing_path if to_unix_path(item).startswith(emsdk_root_path)]
-  new_emsdk_tools = [item for item in path_add if not normalized_contains(existing_emsdk_tools, item)]
+  existing_emsdk_tools = [i for i in existing_path if to_unix_path(i).startswith(emsdk_root_path)]
+  new_emsdk_tools = [i for i in path_add if not normalized_contains(existing_emsdk_tools, i)]
 
   # Existing non-emsdk tools
-  existing_path = [item for item in existing_path if not to_unix_path(item).startswith(emsdk_root_path)]
-  new_path = [item for item in path_add if not normalized_contains(existing_path, item)]
+  existing_path = [i for i in existing_path if not to_unix_path(i).startswith(emsdk_root_path)]
+
```
kripken: why a newline?

Issue #412, opened 2019-12-15
emsdk update does fail with 403 from AWS


diegocr:
```
$ emsdk update
Error downloading URL 'https://s3.amazonaws.com/mozilla-games/emscripten/packages/emsdk_windows_update.zip': HTTP Error 403: Forbidden
Fetching all tags from Emscripten Github repository...
Done. 181 tagged releases available, latest is 1.39.4.
Fetching all tags from Binaryen Github repository...
Done. 90 tagged Binaryen releases available, latest is 1.39.1.
Fetching all precompiled tagged releases..
Error downloading URL 'https://s3.amazonaws.com/mozilla-games/emscripten/packages/llvm/tag/win_32bit/index.txt': HTTP Error 403: Forbidden
Error downloading URL 'https://s3.amazonaws.com/mozilla-games/emscripten/packages/llvm/tag/win_64bit/index.txt': HTTP Error 403: Forbidden
```

kripken:
This looks like https://github.com/emscripten-core/emscripten/issues/5291#issuecomment-308078622 but I can't quite understand what the issue was there - @juj do you remember?

juj:
Is Mozilla's bucket even used anymore? @diegocr: Try `git pull`ing emsdk to latest?

kripken:
Oh, right, the URL is old here... sorry I missed that before.

@diegocr updating the emsdk should work, as @juj said. Only old versions of the emsdk will use those old URLs, which are no longer valid.

diegocr:
Thanks guys.

So, i have downloaded this repo files (_Download as ZIP_) and replaced those i had in `C: \emsdk`, then ran the update command:

```bash
$ emsdk update
Downloading: C:/emsdk/zips/master.zip from https://github.com/emscripten-core/emsdk/archive/master.zip
Unpacking 'C:/emsdk/zips/master.zip' to 'C:/emsdk'
Unzipping file 'C:/emsdk/zips/master.zip' failed due to reason: [Error 3] El sistema no puede encontrar la ruta especificada: '\\\\?\\C:unzip_temp\\'
```
^the Spanish message is about being unable to find the path where the zip should be extracted...

I am not very familiar with Python and whether `__file__` is expected to contain an UNC path or otherwise there is something funky from my side/OS, given:

https://github.com/emscripten-core/emsdk/blob/2546d5ab51e8c950e6301799efcdce64ef441667/emsdk.py#L141-L142

```bash
$ python
[...]

PR #438, opened 2020-02-05
Fix ZIP extraction on Windows.

diegocr:
Meant to resolve https://github.com/emscripten-core/emsdk/issues/412#issuecomment-566501047

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -555,7 +555,7 @@ def unzip(source_filename, dest_dir, unpack_even_if_exists=False):
 
       unzip_to_dir = dest_dir
       if common_subdir:
-        unzip_to_dir = os.path.join('/'.join(dest_dir.split('/')[:-1]), 'unzip_temp')
+        unzip_to_dir = os.path.join(os.path.dirname(dest_dir), 'unzip_temp')
 
       # Now do the actual decompress.
       for member in zf.infolist():


emsdk.py
```diff
@@ -522,9 +522,10 @@ def fix_potentially_long_windows_pathname(pathname):
 # race-free way to do it. This method removes the destination if it exists, so
 # the move always works
 def move_with_overwrite(src, dest):
+  # debug_print('move_with_overwrite(src=' + src + ', dest=' + dest + ')')
   if os.path.exists(dest):
     os.remove(dest)
-  os.rename(src, dest)
+  shutil.move(src, dest)
```
sbc100: Is this a separate issue?

emsdk.py
```diff
@@ -555,7 +556,10 @@ def unzip(source_filename, dest_dir, unpack_even_if_exists=False):
 
       unzip_to_dir = dest_dir
       if common_subdir:
-        unzip_to_dir = os.path.join('/'.join(dest_dir.split('/')[:-1]), 'unzip_temp')
+        if WINDOWS:
+          unzip_to_dir = os.path.join(os.getenv('TEMP'), 'unzip_temp')
+        else:
+          unzip_to_dir = os.path.join('/'.join(dest_dir.split('/')[:-1]), 'unzip_temp')
```
sbc100: What did `unzip_to_dir` looks like when it was causing issues for you?

diegocr: `\\?\C:unzip_temp\` which is inaccessible due the missing/stripped slash, as we were discussing in the issue ticket.

emsdk.py
diegocr: Given i did used `os.getenv('TEMP')` as the target where to extract the files under Windows, `os.rename()` may does fail, thus moving to `shutil.move()` to deal with the case where the target and destination are not in the same drive (if they are, it will invoke `os.rename()` anyway as you sure know)

emsdk.py
sbc100: But isn't the bug in `fix_potentially_long_windows_pathname`?

I mean at this point in the program `os.getenv('TEMP')` isn't in the `\\?\` form is it?

I mean if you print `os.path.join(os.getenv('TEMP'), 'unzip_temp') ` you don't see the problematic path do you?

sbc100: I would log the input and output of `fix_potentially_long_windows_pathname` because it seems like the bug must be in there no?

diegocr: That's what I thought initially, that the issue was from the path(s) constructed by `fix_potentially_long_windows_pathname`, but after doing a simple test as pointed in https://github.com/emscripten-core/emsdk/issues/412#issuecomment-582510880 i then realized what the actual/real problem was, the stripped out slash - and you helped me to reach this correct conclusion :-)

sbc100: So is the problem that you $TEMP env var doesn't have a slash?   Is it just "C:something"?   I guess fix_potentially_long_windows_pathname should be updated to handle that case?

diegocr: No, the problem was with the former `dest_dir.split('/')[:-1]` 

To show you, the below is what i get with this patch applied:

```diff
@@ -554,8 +555,15 @@ def unzip(source_filename, dest_dir, unpack_even_if_exists=False):
           break

       unzip_to_dir = dest_dir
+      print('dest_dir = ' + dest_dir)
+      print("dest_dir.split('/')[:-1] = " + ''.join(dest_dir.split('/')[:-1]))
+      print("'/'.join(dest_dir.split('/')[:-1]) = " + '/'.join(dest_dir.split('/')[:-1]))
+      sys.exit(1)
       if common_subdir:
-        unzip_to_dir = os.path.join('/'.join(dest_dir.split('/')[:-1]), 'unzip_temp')
+        if WINDOWS:
+          unzip_to_dir = os.path.join(os.getenv('TEMP'), 'unzip_temp')
+        else:
+          unzip_to_dir = os.path.join('/'.join(dest_dir.split('/')[:-1]), 'unzip_temp')
```

The output:

```bash
$ EMSDK_VERBOSE=1 emsdk update
download_and_unzip(zipfile=https://github.com/emscripten-core/emsdk/archive/master.zip, dest_dir=C:/emsdk)
download_file(url=https://github.com/emscripten-core/emsdk/archive/master.zip, dstpath=zips/)
mkdir_p(C:/emsdk/zips)
Downloading: C:/emsdk/zips/master.zip from https://github.com/emscripten-core/emsdk/archive/master.zip, 58641 Bytes
unzip(source_filename=C:/emsdk/zips/master.zip, dest_dir=C:/emsdk)
Unpacking 'C:/emsdk/zips/master.zip' to 'C:/emsdk'
mkdir_p(C:/emsdk)

dest_dir = C:/emsdk
dest_dir.split('/')[:-1] = C:
'/'.join(dest_dir.split('/')[:-1]) = C:
```

So, the original `dest_dir` was `C:/emsdk`, and after the call to `os.path.join(....)` then `unzip_to_dir` did turn `C:unzip_temp` which is not correct.

Later on, `fix_potentially_long_windows_pathname` converts it to `\\?\C:unzip_temp\` (so it is just correctly prepending the `\\?\` prefix but the problem is not on this)

There, i could also have fixed the issue by avoiding stripping the slash with `dest_dir.split('/')[:-1]`, but i thought more convenient just using $TEMP as the target where to extract the files.

diegocr: > \\?\C:unzip_temp\ which is inaccessible

I just realized i wrongly answered to your first question, sorry - hopefully my last post does clarify everything.

Next i need to grab some sleep 😅

diegocr: I was a little afraid of doing such kind of change that could affect non-Windows platforms, but given you seem confident about it i have just made your suggested change. 👍

PR #440, opened 2020-02-14
1.39.8

kripken:
Also fix Mac CI by installing newer python.


.circleci/config.yml
```diff
@@ -50,6 +50,9 @@ jobs:
       - run:
           name: Install cmake
           command: brew install cmake
+      - run:
+          name: Install python 3
+          command: brew install python3
```
sbc100: I would just combine these into one step maybe?

kripken: Is there a benefit one way or the other? (I'm not sure if there's overhead per "run".) Anyhow, let me land this to avoid the risk of conflicts, and let's think more on it separately.

sbc100: Just to keep the number of steps in the UI under to control, and to match the existing "Install debian packages" on linux.

PR #441, opened 2020-02-22
Python needs to be present before llvm-git-master

juj:
Python needs to be present before llvm-git-master so that -DPYTHON_EXECUTABLE= is passed properly when building llvm-git-master.

### DIFF
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -514,7 +514,7 @@
   {
     "version": "upstream-master",
     "bitness": 64,
-    "uses": ["llvm-git-master-64bit", "node-12.9.1-64bit", "python-3.7.4-64bit", "emscripten-master-64bit", "binaryen-master-64bit"],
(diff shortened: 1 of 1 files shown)


juj:
Hmm, there is an odd error

```
  File "test.py", line 151, in <module>
    assert os.path.exists(LIBC), 'activation supplies prebuilt libc'
AssertionError: activation supplies prebuilt libc
```
I wonder if this PR caused that? That looks unrelated from my impression?

sbc100:
That does looks strange.. I think @kripken wrote that code for shipping the pre-built libraries.  The assertion seems to contradict the message...

kripken:
It does look unrelated. Perhaps the current tot build was broken (libc not included for some reason) and re-running all tests fixes it?

@sbc100 by contradict do you mean the error message seems confusing?

kripken:
(I verified on the last tot build locally, and libc is there.)

sbc100:
Works for me too.  Maybe try re-running the tests?

sbc100:
Oh wait its the test that needs updating.

sbc100:
Fixed in #444

juj:
Closing, this has gone stale.

PR #444, opened 2020-02-26
Fix test after cache directory layout change in emscripten

sbc100:


kripken:
Looks like the only thing missing here might have been a merge of master...

PR #450, opened 2020-03-05
1.39.9

kripken:


kripken:
Looks like the prebuilt libc is not working as intended - was also broken on master 2 days ago. I remember @sbc100 just fixed that, or something related?

kripken:
Oh, there is a wip fix, but not done yet, https://github.com/emscripten-core/emsdk/pull/444

kripken:
I landed that fix, and now we should be good here.

PR #460, opened 2020-04-01
Avoid re-downloading files that are already downloaded

dnfield:
- Writes a `.emsdk_version` file to output directories using the name (which contains a version number and/or hash) once all installation steps have completed successfully
- If that file exists, skip downloading/installation hooks.

My primary use case for this is running `emsdk.py install latest` as part of another repository's setup steps - and I want to avoid developers needlessly downloading the same large tarballs every time they run it.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1885,11 +1885,11 @@ class Tool(object):
       print("Skipped installing " + self.name + ", already installed.")
       return True
 
-    version_hash = self.name
+    version_id = self.name
     version_file_path = os.path.join(self.installation_path(), '.emsdk_version')
     if os.path.isfile(version_file_path):
       with open(version_file_path, 'r') as version_file:
-        if version_hash == version_file.read():
+        if version_id == version_file.read():
           print("Skipped installing " + self.name + ", already installed.")
           return True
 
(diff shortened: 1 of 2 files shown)


emsdk.py
```diff
@@ -1885,6 +1885,15 @@ def install_tool(self):
       print("Skipped installing " + self.name + ", already installed.")
       return True
 
+    version_hash = self.name
+    try:
```
sbc100, line 1889: Can you just use `if os.path.exists()` here instead of try catch?

dnfield: Isn't that racey?  E.g. someone decides to try to clean up their folder while running this command and it then fails because the file existed when we checked but now it doesn't.

sbc100: It is a minor TOCTOU violation yes.  However, I don't mind crashing on people try to install/upgrade and SDK while that the same time running `rm -rf` on it.   I think its OK, maybe even better, to crash in such cases. 

We generally try to avoid try/catch block for this kind of thing in emscripten.   There are many of these kinds of TOCTOU (many that we can't protect against) already in the codebase.  e.g. checking the llvm version before calling in.

dnfield: Updated to use `isfile`.  Also did some minor cleanup of variable usage.

emsdk.py
```diff
@@ -1885,6 +1885,14 @@ def install_tool(self):
       print("Skipped installing " + self.name + ", already installed.")
       return True
 
+    version_hash = self.name
```
kripken, line 1888: this isn't necessarily a hash - for e.g. node it will be something like `node-VERSION`. How about `version_id`?

dnfield: Done

dnfield:
Added the test - I had to shift some var declarations around to make it work properly.

PR #466, opened 2020-04-09
Merge brew stages

kripken:
Suggested by @sbc100 a while ago I believe.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,11 +48,8 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Install cmake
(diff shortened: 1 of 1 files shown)


kripken:
Hmm, looks like there's a merge conflict here as the underlying code has changed a lot in the time meanwhile. Is this still useful do you think @sbc100 ?

sbc100:
I'll leave that up to you. maybe not worth your effort ?

kripken:
In that case probably not. I think I just opened it because you mentioned it might be nicer. But meanwhile it's been rewritten anyhow.

PR #468, opened 2020-04-12
Set user environment variables by default when giving `--global` in Windows

### LINKED ISSUE #189: emsdk construct_env on Windows nukes local paths
Windows-only repro: 
* open a console
* add something to the path: `> set path=%path%;something`
* `> emsdk_env`
* Your something is gone.

(conversely remove something and see it reappear)

The problem is that emsdk gets the path to which it appends from the registry so any local changes are lost. I was tempted to make a pull request which fixes this, but since specific effort is made for it to work like this, I decided to ask first.

The relevant code is here: [emsdk#L1935](https://github.com/juj/emsdk/blob/7a0e27441eda6cb0e3f1210e6837cae4b080ab4c/emsdk#L1935)

Why get the system path in such convoluted way? Why query the registry instead of just use the path environment variable?

### LINKED ISSUE #467: Way to set user environment variables
Currently `tool.win_activate_env_vars` always gets `permanently_activate=True`, which then leads to setting global environment variables. I need a way to set it for local user scope.

saschanaz:
Fixes #467

This makes `./emsdk activate --global` to set user environment variables by default instead of previous system-wide-only activation.

Also removes some magic that implicitly nukes user envvars when system ones are shadowed, as a system admin should understand what's going on when setting envvars. Thus fixes #189 and closes #192.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2584,7 +2584,7 @@ def to_msys_path(p):
 
 # Looks at the current PATH and adds and removes entries so that the PATH reflects
 # the set of given active tools.
-def adjusted_path(tools_to_activate, log_additions=False, system=False):
+def adjusted_path(tools_to_activate, system=False):
   # These directories should be added to PATH
   path_add = get_required_path(tools_to_activate)
   # These already exist.
--- a/emsdk.py
+++ b/emsdk.py
@@ -1800,18 +1800,13 @@ class Tool(object):
           return False
     return True
 
-  def win_activate_env_vars(self, permanently_activate):
+  def win_activate_env_vars(self, system):
     if WINDOWS:
       envs = self.activated_environment()
       for env in envs:
         key, value = parse_key_value(env)
 
-        if permanently_activate:
-          # If there is an env var for the LOCAL USER with same name, it will
-          # hide the system var, so must remove that first.
-          win_delete_environment_variable(key, False)
-
-        win_set_environment_variable(key, value, permanently_activate)
+        win_set_environment_variable(key, value, system)
 
   # If this tool can be installed on this system, this function returns True.
   # Otherwise, this function returns a string that describes the reason why this
@@ -2509,7 +2504,7 @@ def copy_pregenerated_cache(tools_to_activate):
 # Reconfigure .emscripten to choose the currently activated toolset, set PATH
 # and other environment variables.
 # Returns the full list of deduced tools that are now active.
-def set_active_tools(tools_to_activate, permanently_activate):
+def set_active_tools(tools_to_activate, permanently_activate, system):
   tools_to_activate = process_tool_list(tools_to_activate, log_errors=True)
 
   generate_dot_emscripten(tools_to_activate)
@@ -2533,13 +2528,13 @@ def set_active_tools(tools_to_activate, permanently_activate):
   if WINDOWS and permanently_activate:
     # Individual env. vars
     for tool in tools_to_activate:
-      tool.win_activate_env_vars(permanently_activate=True)
+      tool.win_activate_env_vars(system=system)
 
     # PATH variable
-    newpath, added_items = adjusted_path(tools_to_activate, system_path_only=True)
+    newpath, added_items = adjusted_path(tools_to_activate, system=system)
     # Are there any actual changes?
     if newpath != os.environ['PATH']:
-      win_set_environment_variable('PATH', newpath, system=True)
+      win_set_environment_variable('PATH', newpath, system=system)
 
   if len(tools_to_activate) > 0:
     tools = [x for x in tools_to_activate if not x.is_sdk]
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -2589,17 +2584,12 @@ def to_msys_path(p):
 
 # Looks at the current PATH and adds and removes entries so that the PATH reflects
 # the set of given active tools.
-def adjusted_path(tools_to_activate, log_additions=False, system_path_only=False):
+def adjusted_path(tools_to_activate, log_additions=False, system=False):
```
sbc100, line 2587: Looks like the `log_additions` param is never uses, or passed.  Unrelated to this PR but since your are changing that line maybe delete it?

sbc100:
This change looks great!

The only slight concern I have is changing the default.    Do you know that history here?  Why would setting system-wide have been desirable?  Under what circumstances would it still be desirable?  Do we need the option at all?   Maybe we simply remove the option and always set user settings?

@juj do you remember why this was done this way in the first place?

juj:
The intent of calling `emsdk activate` has been to only enable the env. vars for current/calling user prompt, not do any permanent modifications to the system.

Then, `emsdk activate --global` is intended to apply permanent modifications to "All users" section of the registry.

(The doc piece at emsdk --help looks like up to date with this respect:

```
    if WINDOWS:
      print('''
   emsdk activate [--global] [--embedded] [--build=type] [--vs2013/--vs2015/--vs2017] <tool/sdk>

                                - Activates the given tool or SDK in the
                                  environment of the current shell. If the
                                  --global option is passed, the registration
                                  is done globally to all users in the system
```

There has never existed an option to perform "Current user" only registry activation. If that is desirable, perhaps it would be good to have that as a new option `emsdk activate --local` or `emsdk activate --local_user`, instead of changing the existing `emsdk activate` from "current prompt only" to "permanent local user"?

saschanaz:
> If that is desirable, perhaps it would be good to have that as a new option `emsdk activate --local` or `emsdk activate --local_user`, instead of changing the existing `emsdk activate` from "current prompt only" to "permanent local user"?

Sorry for misleading description, but no, this doesn't change `emsdk activate` to set the permanent user global environment, instead it modifies `emsdk activate --global` to be user scope by default.

I think defaulting to user global environment is better as it does not require admin privilege.

juj:
> I think defaulting to user global environment with `--global` is better as it does not require admin privilege.

It feels like changing `--global` from meaning global to user local would be counterintuitive? A new option `--local` would be a better name?

saschanaz:
IMO it's global user variables and global system variables, not "local", which sounds like non-permanent local shell variable.

sbc100:
I also feel like changing the meaning of an existing flag I a little confusing/misleading.    How about simply removing the old `--global` flag which is ambiguous to something more specific such as `--persistent` or `--registry`? 

Any users of the old `--global` flag could be asked to be more specific and switch to either `--persistent`  or `--persistent + --system`.

saschanaz:
`--persistent` sounds good to me, as `--global` can be an alias for `--persistent --system` then.

juj:
`--persistent` sounds good, although I would recommend instead of `--persistent --global` to have `--persistent_global`, since the flags are not orthogonal: passing `--global` on its own would not be meaningful.

saschanaz:
So does `--persistent` and `--persistent-global` sound good? Those should be mutually exclusive.

sbc100:
I like `--persistent-system` and you could even keep the old `--global` flag as an alias to that or while/

sbc100:
Any update on this?  I think it would be good change overall.

saschanaz:
I'll try get some time this weekend.

sbc100:
Any update?

aminya (changes requested):
This is a very good change. What is blocking this from merging? 

Can someone with write access update this branch?

emsdk.py
```diff
@@ -2842,6 +2832,7 @@ def extract_bool_arg(name):
   arg_old = extract_bool_arg('--old')
   arg_uses = extract_bool_arg('--uses')
   arg_global = extract_bool_arg('--global')
+  arg_system = extract_bool_arg('--system')
```
aminya, line 2835: If `arg_system` is provided bu `arg_global` is not, it should automatically set that.
```suggestion
  arg_system = extract_bool_arg('--system')
  if arg_system
  	arg_global=True
```

aminya (changes requested):
If there is anywhere to document the meanings of these options, we should.

```
global: set the environment variables permanently.
system: set the environment variables for all users of the system.
```

sbc100:
`emsdk --help` is the main place we document this stuff.

saschanaz:
I have to admit that I'm not actively using emsdk nowadays. I still intend to work on it but if anyone wants to take this over, please do.

sbc100:
No problem.  Hopefully someone can pick this up and get it landed.

aminya:
@sbc100 this is almost ready. If you have write-access why don't you just finish it?

sbc100:
I will get around to it at some point sure, but I've got a lot of other things I'm working so if anyone else want to pick this up in the mean time feel free.

aminya:
Well, this is a critical bug on Windows, and we should fix it as soon as possible.

Issue #470, opened 2020-04-15
32bit support gone?


j4ys0n:
i'm trying to recompile doom with some modifications. that needs to be compiled on a 32bit os. but it looks like 32bit support was dropped from emsdk?

```
root@doom:~/emsdk$ ./emsdk install latest
Error: 'sdk-releases-upstream-e13b86d4dbd9a986525ef27d4ad8157949b9bc3a-64bit' is only provided for 64-bit OSes.
root@doom:~/emsdk$ ./emsdk install sdk-master-32bit
Error: No tool or SDK found by name 'sdk-master-32bit'.
```

sbc100:
Can you try with #472 and with the following command:

`./emsdk install sdk-upstream-master-32bit`

j4ys0n:
that may have fixed it!

i'm running into some issues with cmake version now, but i'll work on resolving those and report back.

thanks!

sbc100:
Looks like one problem is that we don't currently provide any node-12.9.1-32bit.

sbc100:
It looks like upstream node no longer supply 32-bit packages: https://nodejs.org/download/release/v12.9.1/

guest271314:
Just tried instructions at https://emscripten.org/docs/getting_started/downloads.html getting error ```
./emsdk install latest
Error: 'sdk-releases-upstream-574ad04affb82cc36a32dd89b2a87bea4fb30eba-64bit' is only provided for 64-bit OSes.
```
is there a solution?

PR #471, opened 2020-04-15
Re-enable building master from source for linux 32-bit

sbc100:
Doing this for windows is trickier as we would also need to provide
the python binaries for win32.

Fixes: #470


sbc100:
Verified that this workd.

PR #472, opened 2020-04-15
Use embedded configuration by default

sbc100:
I can't see any reason why the embedded mode isn't always preferable.
The emsdk is designed to be used as-is, and per-user customizations
don't make sense.

I'm hoping if this sicks we can just remove this options completely.

This is part of a wider plan to remove the use of the user's HOME
directory completely:
  https://github.com/emscripten-core/emscripten/issues/9543

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2760,15 +2760,15 @@ def main():
 
     if WINDOWS:
       print('''
-   emsdk activate [--global] [--embedded] [--build=type] [--vs2017/--vs2019] <tool/sdk>
+   emsdk activate [--global] [--[no-]embedded] [--build=type] [--vs2017/--vs2019] <tool/sdk>
 
                                 - Activates the given tool or SDK in the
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -54,7 +56,9 @@ jobs:
           name: Install python 3
           command: brew install python3
       - run: ./test.sh
-      - run: ./test.py
+      - run: |
+         - source emsdk_env.sh
```
kripken, line 60: is this change required because of this PR?

sbc100: I should be able to revert that, now that https://github.com/emscripten-core/emscripten/pull/10935 has landed.

Technically I think the tests are wrong here because they should run the context in whic `emsdk_env` has been run.   i.e. the tests were not following the instructions here.

sbc100:
I'm going to write the mailing list about this change now.. and hopefully is now good to go?

PR #483, opened 2020-04-29
Add support for one custom forked copy of Emscripten repositories

juj:
We are automating emsdk builds for our internal QA process. For that use, it is advantageous to enable emsdk to point to a custom set of forks that one can install and activate via emsdk.

This PR adds support for users to have one custom clone per emsdk installation, by specifying a set of custom URLs to their own GitHub forks.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -182,6 +182,32 @@ if POWERSHELL:
 else:
   EMSDK_SET_ENV = os.path.join(emsdk_path(), 'emsdk_set_env.bat')
 
+
+# Parses https://github.com/emscripten-core/emscripten/tree/d6aced8 to a pair (https://github.com/emscripten-core/emscripten, d6aced8)
+def parse_github_url_and_refspec(url):
+  if not url:
+    return ('', '')
+
+  if url.endswith(('/tree/', '/tree', '/commit/', '/commit')):
+    raise Exception('Malformed git URL and refspec ' + url + '!')
+
+  if '/tree/' in url:
+    if url.endswith('/'):
+      raise Exception('Malformed git URL and refspec ' + url + '!')
+    return url.split('/tree/')
+  elif '/commit/' in url:
+    if url.endswith('/'):
+      raise Exception('Malformed git URL and refspec ' + url + '!')
+    return url.split('/commit/')
+  else:
+    return (url, 'main')  # Assume the default branch is main in the absence of a refspec
+
+
+emscripten_config_directory = os.path.expanduser("~/")
+# If .emscripten exists, we are configuring as embedded inside the emsdk directory.
+if os.path.exists(os.path.join(emsdk_path(), '.emscripten')):
+  emscripten_config_directory = emsdk_path()
+
 ARCHIVE_SUFFIXES = ('zip', '.tar', '.gz', '.xz', '.tbz2', '.bz2')
 
 
@@ -498,11 +524,11 @@ def num_files_in_directory(path):
   return len([name for name in os.listdir(path) if os.path.exists(os.path.join(path, name))])
 
 
-def run(cmd, cwd=None):
+def run(cmd, cwd=None, quiet=False):
   debug_print('run(cmd=' + str(cmd) + ', cwd=' + str(cwd) + ')')
   process = subprocess.Popen(cmd, cwd=cwd, env=os.environ.copy())
   process.communicate()
-  if process.returncode != 0:
+  if process.returncode != 0 and not quiet:
     errlog(str(cmd) + ' failed with error code ' + str(process.returncode) + '!')
   return process.returncode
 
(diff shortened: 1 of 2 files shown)


kripken:
This seems like a large change, and I spent some time trying to think of a way to make it simpler.

Could we allow an optional manifest file to be merged in to the main one, possibly overriding entries there? Then custom setups could change repo locations and branches etc. that way.

juj:
> Could we allow an optional manifest file to be merged in to the main one, possibly overriding entries there? Then custom setups could change repo locations and branches etc. that way.

An issue with that is that there are a number of moving/internal parts there, that we would have to lock down in time the JSON entries that we will have, so that the overridden JSON entries that people use will stay compatible. It will be a bit hidden what people write in their JSONs, so we may break "unknown" uses.

Having a cmdline option helps ensure the API interface as a cmdline interface, and we'll know what we need to keep supporting.

In CI environments it is easier to do `emsdk install sdk-forked-upstream-64bit --forked-emscripten-url https://github.com/myfork/emscripten.git` than to programmatically create an override JSON that would carry all the changes starting from https://github.com/emscripten-core/emsdk/pull/483/files#diff-a1b4faa6091de6a8facd6ea8c3ffabd4R32.

Also, it is easier to instruct users to type in `emsdk install sdk-forked-upstream-64bit --forked-emscripten-url https://github.com/myfork/emscripten.git` to use a custom emscripten fork, rather than again ask them to create a new file that adds in https://github.com/emscripten-core/emsdk/pull/483/files#diff-a1b4faa6091de6a8facd6ea8c3ffabd4R32 on their own.

We could look at doing some kind of file format that is not a "full" manifest like emsdk_manifest.json is, but then the file would probably contain the same info that the cmdline proposed here does, and the question is whether specifying that in a file has benefit over the cmdline options?

kripken:
> An issue with that is that there are a number of moving/internal parts there, that we would have to lock down in time the JSON entries that we will have, so that the overridden JSON entries that people use will stay compatible. It will be a bit hidden what people write in their JSONs, so we may break "unknown" uses.

Oh, I may not understand your use case fully. Don't you provide the JSON files for your users? (Who are "people" in your last sentence?)

> Also, it is easier to instruct users to type in emsdk install sdk-forked-upstream-64bit --forked-emscripten-url https://github.com/myfork/emscripten.git to use a custom emscripten fork, rather than again ask them to create a new file that adds in https://github.com/emscripten-core/emsdk/pull/483/files#diff-a1b4faa6091de6a8facd6ea8c3ffabd4R32 on their own.

That's a good point. Yeah, commandline flags for users does make the most sense for that.

Can we perhaps add generic commandline flags for local manifest modification? That is,

```
emsdk install sdk-upstream-64bit --localmod emscripten url https://github.com/myfork/emscripten.git
```

I'm not tied to the particular syntax, but the idea is the user says "apply a local mode of the `emscripten` entry's `url` to be ...".

juj:
> Don't you provide the JSON files for your users?

We do, but it is our CI that is awkward.. I could programmatically generate the extra JSON file for the CI if needed, but not sure I want to spec such an extra file format interface.

By people I was referring to general Emscripten developers who might create their own JSONs. We don't document the file format of the JSONs, and a lot of that has been considered technical/internal, so it would be a large change to diligently turn that to a public facing file.

> Can we perhaps add generic commandline flags

The generic command line flags kind of have the identical problem as the extra JSON files? Since that would force the JSON syntax out in the public, and people would have to learn the meaning of the file, and it would likewise "freeze" the file structure from being able to change - some internal change might cause a breakage in the flags that someone is passing.

In other words, if we opened such generic support for command line flags, what possible alterations should be expected to be reasonable? Would that require freezing all the existing JSON fields in place, so that someone's use of overriding an entry for their own purpose will not break on a refactor of the JSON file?

kripken:
> The generic command line flags kind of have the identical problem as the extra JSON files? Since that would force the JSON syntax out in the public, and people would have to learn the meaning of the file, and it would likewise "freeze" the file structure from being able to change - some internal change might cause a breakage in the flags that someone is passing.

I think I understand this concern in general. But the main use cases seem fine to me. Concretely, I mean

```
emsdk install sdk-upstream-64bit --localmod emscripten url https://github.com/myfork/emscripten.git
emsdk install sdk-upstream-64bit --localmod binaryen url https://github.com/myfork/binaryen.git
emsdk install sdk-upstream-64bit --localmod llvm-git url https://github.com/myfork/binaryen.git
```

Those don't seem like they expose any internal API in a bad way. The user just mentions "emscripten, binaryen, llvm", and the "url" field. They don't need to learn any internal details of the JSON. (The one odd thing maybe is "llvm-git" has "-git" as a suffix, but maybe we should change that to just "llvm" anyhow?)

I do agree that such a feature *might* be used to do more things, maybe some of them dangerous. But I think the benefit of doing it that way is it's a much smaller change to emsdk.py (all it would do is, right after loading the manifest file, apply the localmods). And the more general capabilities, while potentially dangerous if misused, likely allow more useful things for people that are careful.

But maybe this is a philosophical difference. I don't want to waste people's time here if no one else is concerned about the extra complexity here. @sbc100 what do you think?

sbc100:
Obviously we would prefer not to have to support forked versions but it sounds like @juj you have decided that it really is necessary in your case.

Can you say if you think this will be a permanent thing?   Do you plan on maintain your own patches would you idea be the patches migrate over time into upstream emscripten?   I would hope the later.

I don't like the extra complexity this adds, but its sounds like something you really need at your end, and forking emsdk itself would be sad too.

Do you need to support forks for all 4 repos or would it be enough to support just emscripten?  It would reduce the size of this patch a fair bit if just do this for emscripten if possible.

juj:
Closing this in favor of a simpler variant.

PR #492, opened 2020-05-07
test.py should require that the SDK be activated

sbc100:
It was working previously by accident because the tests were
using the defaults such as $HOME/.emscripten.  However usage of the
emsdk should really only happen in the emsdk environment where
EM_CONFIG dictates the config location (not just using the default).

This change is needed as precursor to making `--embedded` the default
(in this case the config doesn't live in `$HOME/.emscripten`).

### DIFF
--- a/scripts/test.py
+++ b/scripts/test.py
@@ -9,6 +9,9 @@ import tempfile
 WINDOWS = sys.platform.startswith('win')
 MACOS = sys.platform == 'darwin'
 
+assert 'EM_CONFIG' in os.environ, "emsdk should be activated before running this script"
+
+emconfig = os.environ['EM_CONFIG']
 upstream_emcc = os.path.join('upstream', 'emscripten', 'emcc')
 fastcomp_emcc = os.path.join('fastcomp', 'emscripten', 'emcc')
 emsdk = './emsdk'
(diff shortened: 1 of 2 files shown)


emsdk.py
```diff
@@ -81,6 +78,20 @@
...
+POWERSHELL = bool(os.getenv('EMSDK_POWERSHELL'))
+CSH = bool(os.getenv('EMSDK_CSH'))
+BASH = bool(os.getenv('EMSDK_BASH'))
+CMD = bool(os.getenv('EMSDK_CMD'))
+if not CSH and not POWERSHELL and not BASH and not CMD:
+  # Fall back to default of `cmd` on windows and `bash` otherwise
+  if WINDOWS and not MSYS:
+    CMD = True
+  else:
+    BASH = True
```
kripken, line 92: this change looks unrelated to testing, so the PR is doing more than that? Or is this NFC somehow?

scripts/test.py
```diff
@@ -9,6 +9,9 @@
 WINDOWS = sys.platform.startswith('win')
 MACOS = sys.platform == 'darwin'
 
+assert 'EM_CONFIG' in os.environ, "emsdk should be activated before running this script"
```
kripken, line 12: don't we test activation here? must we do it before as well?

kripken: (and does an activation before not change the tests for activation in this script?)

sbc100: Activating and sdk is not enough; that only creates the config files.

One also needs to run `source emsdk_env.sh` to activate it within the current shell.

emsdk.py
sbc100: Sorry that is a separate change that I'm landing first.

sbc100:
Rebased on top of correct change.

scripts/test.py
kripken: Oh, I think the terminology confused me here. `emsdk activate` does stuff like copy precompiled libc, etc. I assumed that the `activate` you meant.

sbc100: It is a little confusing isn't it.   I guess the job of `activate` is to decide which tools are active.. but unless you run `emsdk_env` when the emsdk itself as a whole is not active if your current shell.

Perhaps we should remind users who run `./emsdk activate` in shells without EMSDK that they still have one more step to actually activate in the current shell.

PR #495, opened 2020-05-07
Add CACHE setting to config file (for embedded users)

sbc100:
See https://github.com/emscripten-core/emscripten/pull/11091

PR #498, opened 2020-05-08
Fix for parsing existing config in --embedded mode after #495

### LINKED ISSUE #497: construct_env with activate --embedded won't list paths to node and upstream/emscripten
While creating docker image with emscripten, I got following error:
```sh
/bin/bash: em++: command not found
```
This appears to be because emsdk_set_env.sh (emitted with `./emsdk construct_env`) does not add path to node and upstream/emscripten.
```sh
# emsdk_set_env.sh
export PATH="/emsdk_portable:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
export EMSDK="/emsdk_portable"
```
and content of .emscripten file is:
```txt
import os
emsdk_path = os.path.dirname(os.environ.get('EM_CONFIG')).replace('\\', '/')
CACHE = emsdk_path + '/.emscripten_cache'
NODE_JS = emsdk_path + '/node/12.9.1_64bit/bin/node'
LLVM_ROOT = emsdk_path + '/upstream/bin'
BINARYEN_ROOT = emsdk_path + '/upstream'
EMSCRIPTEN_ROOT = emsdk_path + '/upstream/emscripten'
```
While changes in [emsdk.py:1486](https://github.com/emscripten-core/emsdk/blob/master/emsdk.py#L1486), `emsdk_path + '` is still expected to be `' + emsdk_path + '` in [emsdk.py:1742](https://github.com/emscripten-core/emsdk/blob/master/emsdk.py#L1742)

sbc100:
Ooops, I guess even the most innocent-seeming change can have
unintended consequences. Switching to embedded by default will at
least make this path more tested in the future.

Fixed #497

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1736,8 +1736,9 @@ class Tool(object):
         debug_print(str(self) + ' is not active, because key="' + key + '" does not exist in .emscripten')
         return False
 
-      # If running in embedded mode, all paths are stored dynamically relative to the emsdk root, so normalize those first.
-      dot_emscripten_key = dot_emscripten[key].replace("' + emsdk_path + '", emsdk_path())
+      # If running in embedded mode, all paths are stored dynamically relative
+      # to the emsdk root, so normalize those first.
+      dot_emscripten_key = dot_emscripten[key].replace("emsdk_path + '", "'" + emsdk_path())
       if dot_emscripten_key != value:
         debug_print(str(self) + ' is not active, because key="' + key + '" has value "' + dot_emscripten_key + '" but should have value "' + value + '"')
         return False


emsdk.py
```diff
@@ -1736,8 +1736,9 @@ def is_active(self):
         debug_print(str(self) + ' is not active, because key="' + key + '" does not exist in .emscripten')
         return False
 
-      # If running in embedded mode, all paths are stored dynamically relative to the emsdk root, so normalize those first.
-      dot_emscripten_key = dot_emscripten[key].replace("' + emsdk_path + '", emsdk_path())
+      # If running in embedded mode, all paths are stored dynamically relative
+      # to the emsdk root, so normalize those first.
+      dot_emscripten_key = dot_emscripten[key].replace("emsdk_path + '", "'" + emsdk_path())
```
kripken, line 1741: this still looks weird to me even with the explanation from before... it's shifting a `'` from the right to the left. That seems really surprising.

Can't this just replace `' emsdk_path '` (note the space before and after) with the path?

sbc100: No, we also want to replace the ` +=` part.

Old:

```
FOO = emsdk_path + "/some/path"
```

New:

```
FOO = "/the/actual/path/some/path"
```


I can probably come up with a cleaner way to do this overall, but for now this fixes the immediate breakage and I don't think its any nicer before of after this change, right?

kripken: I see better now, thanks!

PR #499, opened 2020-05-11
Always use node 12.9.1

kripken:


sbc100:
I'm not sure this is needed.  All these tag are for legacy versions. 

Any sdk with a modern emscripten already uses the new node.  So updated the node dependency in esccripten ToT is not dependent on doing this. 

Is it best to just leave these old version be maybe?

kripken:
What's the benefit of leaving them? It seems safest to me to know that the emsdk always uses a new node, which has no downsides I am aware of.

In theory if there is some platform on which new node can't run, it is still possible to use an older emsdk. But I think carrying around old version support forever isn't great.

sbc100:
I guess I'm wondering if the SDKs you are modifying here are historical or current?    If they are historical its probably best to leave set in stone.

If they are current and people are still using them install current tagged versions then yes we should update.  But in that case we should update the python version too.   

Are they current? Can you use those tags to install recent SDKs?

kripken:
I'm not totally sure of the status of "nightly" and other builds here.

But if they are legacy, then we don't support them, and should remove them, IMO - let people use old unsupported emsdks to install old unsupported things.

IOW I'd rather get a bug report on this change, and then remove that legacy build entirely, than keep "support" for all legacy builds.

sbc100:
Yes,  I think we should look into removing them rather than changing them.

There are basically two possibilities:

1.  These are legacy and never use emscripten ToT.  
2.  These are current and can be used to install SDKs that include recent emscripten.

If (2) is true we should land this change.

If (1) is true we should not land this change.. and perhaps consider removing them completely.

kripken:
> These are legacy and never use emscripten ToT.

As best I can tell that is correct. They all use earlier versions than 1.37.X.

But I'm not sure I agree with you. I'd like to land this, as it shouldn't break those things - it's orthogonal to them. Separately, we can consider removing these legacy modes. That's not urgent and maybe I'm missing a reason to not do so. But there is an immediate benefit to having a guaranteed new node for all our users, that we can use newer node features in the JS compiler.

sbc100:
But the old`1.37.X` versions of emscripten are known to work and tested with the old node version and not with the new node version. 

Why retroactively change an existing, shipping SDK?  I don't see the benefit of doing this.   It doesn't unblock anything does it?

sbc100:
Ditto the old python version.

kripken:
Well, I think I've explained why I disagree, but your point of view is reasonable too, so this seems not worth debating more. I don't think it blocks anything. Closing.

sbc100:
Sorry I didn't mean to come across as obstinate.  I just want to full understand how these things are used before modifying/removing them.

kripken:
No worries @sbc100 , as I said I appreciate your opinion here.

PR #500, opened 2020-05-12
Remove legacy builds

kripken:
"Legacy" here meaning anything using only < 1.38.0, and anything
using 'nightly' mozilla builds (which have not been done for years).

This also lets us remove node 4.1.1 (but not 8.9.1).


juj:
lgtm

PR #501, opened 2020-05-12
Update default emscripten cache directory

sbc100:
Now that emscripten uses an in-tree cache directory there is no
need to copy the pre-built libraries around.
    
See: https://github.com/WebAssembly/waterfall/pull/644
    
Also for the sake of older versions of emscripten explictly set
EM_CACHE environment variable to point to the in-tree cache
directory.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2433,6 +2433,7 @@ def copy_pregenerated_cache(tools_to_activate):
       em_cache_dir = os.path.join(config['EMSCRIPTEN_ROOT'], 'cache')
       break
   else:
+    debug_print('Not copying pregenerated libaries (no EMSCRIPTEN_ROOT found)')
     return
 
   # If we found an EMSCRIPTEN_ROOT look for any tools that include
--- a/emsdk.py
+++ b/emsdk.py
@@ -1455,7 +1455,6 @@ def generate_dot_emscripten(active_tools):
   if embedded:
     cfg += 'import os\n'
     cfg += "emsdk_path = os.path.dirname(os.environ.get('EM_CONFIG')).replace('\\\\', '/')\n"
-    cfg += "CACHE = '%s'\n" % sdk_path('.emscripten_cache')
 
   # Different tools may provide the same activated configs; the latest to be
   # activated is the relevant one.
@@ -1490,11 +1489,9 @@ JS_ENGINES = [NODE_JS]
   with open(dot_emscripten_path(), "w") as text_file:
     text_file.write(cfg)
 
-  # Clear old cached emscripten content.
+  # Clear old emscripten content.
   try:
-    remove_tree(os.path.join(emscripten_config_directory, ".emscripten_cache"))
     os.remove(os.path.join(emscripten_config_directory, ".emscripten_sanity"))
-    os.remove(os.path.join(emscripten_config_directory, ".emscripten_cache__last_clear"))
   except:
     pass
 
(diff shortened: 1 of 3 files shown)


emsdk.py
```diff
@@ -2448,24 +2447,25 @@ def run_emcc(tools_to_activate):
...
 def copy_pregenerated_cache(tools_to_activate):
   for tool in tools_to_activate:
     pregenerated_cache = getattr(tool, 'pregenerated_cache', None)
     if not pregenerated_cache:
       continue
+    config = {}
+    for specific_cfg in tool.activated_config().split(';'):
+      name, value = specific_cfg.split('=')
+      config[name] = value.strip("'")
+    cache_root = os.path.join(config['EMSCRIPTEN_ROOT'], 'cache')
```
kripken, line 2461: I am slightly worried to not see a utility function for this (getting a key from the config in an activated config) - is there maybe another way to do this? or why has this never been needed before?

regardless it would be nice to put this in a utility function, so the code is easier to read i think.

sbc100: I found a way nicer solution which is just to remove this function :)

emsdk.py
```diff
@@ -2628,15 +2597,14 @@ def construct_env(tools_to_activate, permanent):
...
-    em_cache_dir = sdk_path('.emscripten_cache')
-    if to_unix_path(os.environ.get('EM_CACHE', '')) != to_unix_path(em_cache_dir):
-      env_vars_to_add += [('EM_CACHE', em_cache_dir)]
-    mkdir_p(em_cache_dir)
 
   for tool in tools_to_activate:
+    config = tool.activated_config()
+    if 'EMSCRIPTEN_ROOT' in config:
+      # For older emscripten versions that don't use this default we export
+      # EM_CACHE.
```
kripken, line 2605: don't all configs have `EMSCRIPTEN_ROOT`?

but regardless I'm confused here. if that's in the config, does that mean this config "doesn't use this default" - ? I'm not sure how to understand that phrase and connect it to this code.

sbc100: Regarding the first point.  No EMSCRIPTEN_ROOT doesn't always exists. For example if you just install a single tool:

```
./emsdk install node-12.9.1-64bit,
./emsdk activate node-12.9.1-64bit,
./emsdk_env.sh
```

To answer the second point:

New emscripten: Uses in-tree cache by default
Old emscripten: Needs to be told to use in-tree case via EM_CACHE environment variable.

In neither case is that CACHE or EM_CACHE part of the activated_config.

kripken: I see, thanks!

sbc100:
This change is more focused now, and its compatible with SDK build both before and after the pre-built libraries were moved.

scripts/test.py
```diff
@@ -158,12 +157,10 @@ def run_emsdk(cmd):
 
 print('test tot-upstream')
 run_emsdk('install tot-upstream')
-assert not os.path.exists(LIBC)
```
kripken, line 161: why can't we check this any more?

sbc100: Sorry I should have explained.  This sequence of assert was checking the "activate" re-installs libraries in the cache.

However, in the now model there is no "install into cache step".   The cache is just part of the tar archive.. which means if you want to restore the cache after a `--clear-cache` you need to for a re-install from the tar archive (or just let it get built on demand).

PR #502, opened 2020-05-13
Refactor such that Tool.activated_config returns a map. NFC

sbc100:
Split out from #501


sbc100:
The test failure here is unrealted.  Its due to the fact the current tot build don't include the libraries in the expected location anymore.  It will be fixed by #501.

PR #503, opened 2020-05-14
1.39.16

sbc100:


scripts/test.py
```diff
@@ -10,6 +10,11 @@
 MACOS = sys.platform == 'darwin'
 
 assert 'EM_CONFIG' in os.environ, "emsdk should be activated before running this script"
+# Remove the EM_CACHE environment variable.  It interferes with testing since 
+# it would otherwise be fixed for the duration of the script and we expect
+# "emsdk activate" to be able swtich between SDKs during the running of this
+# script.
+del os.environ['EM_CACHE']
```
kripken, line 17: doesn't `activate` change the value of that env var, so it would switch properly? or is that set from somewhere else?

scripts/test.py
```diff
@@ -10,6 +10,11 @@
 MACOS = sys.platform == 'darwin'
 
 assert 'EM_CONFIG' in os.environ, "emsdk should be activated before running this script"
+# Remove the EM_CACHE environment variable.  It interferes with testing since 
+# it would otherwise be fixed for the duration of the script and we expect
+# "emsdk activate" to be able swtich between SDKs during the running of this
```
kripken, line 15: typo in `switch`

scripts/test.py
```diff
@@ -95,31 +98,27 @@ def hack_emsdk(marker, replacement):
...
-
-  test_build('', expected=first_time_system_libs,
-             unexpected=unexpected_system_libs)
-  test_build(' -O2', unexpected=unexpected_system_libs + first_time_system_libs)
-  test_build(' -s WASM=0', unexpected=unexpected_system_libs + first_time_system_libs)
-  test_build(' -O2 -s WASM=0', unexpected=unexpected_system_libs + first_time_system_libs)
+  # The emsdk ships all system libraries so we don't expect to see any
+  # cache population unless we explicly --clear-cache.
+  test_build('', expected=False)
+  check_call(emcc + ' --clear-cache')
```
kripken, line 118: this is maybe excessive - CI here will wait on building all of libc. maybe just delete say `libdlmalloc`?

scripts/test.py
sbc100: Its it not possible for a child process to modify the environment of the parent process.  So the is no way activate can do this.  This is why we have the `source ./emsdk_env.sh` thing... the source in bash takes case or modifying the running shell.

Within this python script the environment cannot be changed by calling a subprocess.

scripts/test.py
sbc100: The problem is that each version of SDK now has it own cache... so I'd need to figure out where each libc is.  For the upstream SDK I know this because of `EM_CACHE`, but for fastcomp I'd need to derive the cache directory somehow.

scripts/test.py
kripken: I see, thanks.

scripts/test.py
kripken: Ok, how about removing fastcomp cache testing then?

sbc100: Sure.. although building libc only takes less than a minute, no?    Does it it matter on the emsdk CI?

kripken: Not a big deal, yeah. But it's nice if emsdk CI is fast as we want to avoid anything landing on emscripten while we tag. lgtm either way though.

emscripten-releases-tags.txt
```diff
@@ -1,6 +1,7 @@
 {
-  "latest": "1.39.15",
+  "latest": "1.39.16",
   "releases": {
+    "1.39.16": "701922c0d4a47ad6bc39bd72f083813a71be4ea7",
```
kripken, line 4: more stuff landed in emscripten meanwhile, so we need to update this

PR #504, opened 2020-05-17
Add a .gitattributes file that forces unix shell scripts to have unix file endings

aatifsyed:
Errors running files for [quickstart](https://webassembly.org/getting-started/developers-guide/) on unix before this fix:
```bash
aatif@PC6353:~/lab/emsdk$ ./emsdk
: not found ./emsdk: 
: not found ./emsdk: 
: not found: ./emsdk: 
./emsdk: 19: ./emsdk: Syntax error: end of file unexpected (expecting "then")
```
I recommend a once-and-for-all overhaul of text files to use LF, and a .gitattributes that sets text files to `text=auto`, but this change gets the repository newbie-friendly faster.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
+# Shell scripts should be converted to native line endings on checkout
+/emsdk                text eol=lf
+/emsdk_env.sh         text eol=lf


sbc100:
Can I ask how you ended up if that situation?   i.e. How is it that emsdk was checked out with the wrong line endings?

aatifsyed:
Good question!
I'm not sure, I was trying to run emsdk, and kept running into this problem
after cloning the repo.
Can't reproduce on a new OS image - maybe my git config was incorrect? I
think the `* text=auto` line would fix incorrect config, but maybe this is
just a freak accident

sbc100:
I could imagine that if you used git for windows to do the checkout it might result in windows line endings?

Issue #477, opened 2020-04-21
More than one version installed (per backend)?


fibbo:
Is it not possible anymore to keep more than one version of a backend? Let's say I have 1.39.12 (upstream) installed and run `emsdk install 1.39.13`. The old version is then gone. Do I have to manage versions myself?

kripken:
We removed this after users reported problems with disk space - each version takes hundreds of MB, so if we keep all the versions forever eventually that becomes a problem.

It would be good to fix this if you or someone else has time. One option is to keep versions but issue a warning when the disk space gets high, with instructions for how to clear it. Another option might be to save the last N versions but no more.

sbc100:
I'd like to see this behaviour restored if possible.

flab03:
So, when installing a new version, the old version is silently erased, but the situation is actually worse than this. Every version from 1.38.33 onwards is marked INSTALLED. Running `emsdk activate <old_version>` doesn't fail. So in effect emsdk is lying to me when it says that it activated an old version when it didn't. This is extremely counterintuitive and I lost a lot of time figuring this out, especially since README.md says

> You can toggle between different tools and SDK versions by running emsdk activate <tool/sdk name>. 

If the new policy is to keep only one version, then at least make sure that activating a version other than the installed version fails.

kripken:
Sorry about that @flab03 , thanks for the bug report! I opened #505 which I think will fix the reporting.

I also verified that with that patch, trying to activate a non-installed tool will skip it and warn:
```
Warning: The SDK/tool 'releases-upstream-40f3caabcef7b52bdde63d3883462414d7a25bec-64bit' cannot be activated since it is not installed! Skipping this tool...
```

torarnv:
I think this is still an issue. I was under the impression, based on the distinction between `install` and `activate`, that the emsdk repo/tool could be used to mange multiple parallel installations, but in practice this does not seem to be the case.

The tool however happily allows me to install `tot-upstream` and `3.1.25` and activate them both, without actually changing anything in `upstream/emscripten`. 

And `list` shows both versions installed.

sbc100:
Its is still the case that you can only install one pre-build sdk at a time.  Each installation will replace the previous one that lives in the `upstream` directory.

I'd like find a way to do parallel installations of SDK versions (like you can for different versions of node, python and other tools in emsdk).   But currently this is not possible.

MoshiKoi:
[...]

PR #505, opened 2020-05-20
Properly report what is installed in "emsdk list"

kripken:
This expands on #460 which added a file in each install dir
with the version, that lets us know not to re-download it. That
didn't integrate with `is_installed`, so it wasn't reported properly,
which this PR fixes.

With this we can remove a hack for python, as we can simply
not install anything if it's already installed (see line 1859).

Also add the "INSTALLED" indicators on the main listings in
"emsdk list". The output can now look like this:
```
To install/activate it, use one of:
         latest                  [default (llvm) backend]
         latest-fastcomp         [legacy (fastcomp) backend]

Those are equivalent to installing/activating the following:
         1.39.16             INSTALLED
         1.39.16-fastcomp    

All recent (non-legacy) installable versions are:
         1.39.9    
         1.39.8    
         1.39.7    
         1.39.6    
         1.39.5    
         1.39.4    
         1.39.3    
         1.39.2    
         1.39.16    INSTALLED
         1.39.15    
         1.39.14    
         1.39.13    
         1.39.12    
         1.39.11    
         1.39.10    
         1.39.1    
```

Bug was reported in https://github.com/emscripten-core/emsdk/issues/477#issuecomment-631152317

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1677,7 +1677,7 @@ class Tool(object):
     return False
 
   def update_installed_version(self):
-    with open(self.get_version_file_path(), 'wr') as version_file:
+    with open(self.get_version_file_path(), 'w') as version_file:
       version_file.write(self.name)
     return None
 
@@ -2088,6 +2088,10 @@ def get_emscripten_releases_tot():
       continue
     return release
   return ''
+
+
+def get_release_hash(arg, releases_info):
+  return releases_info.get(arg, None) or releases_info.get('sdk-' + arg + '-64bit')
 
 
 # Finds the best-matching python tool for use.
@@ -2887,7 +2891,7 @@ def main():
           arg = arg.replace('-fastcomp', '')
           backend = 'fastcomp'
         arg = arg.replace('sdk-', '').replace('-64bit', '').replace('tag-', '')
-        release_hash = releases_info.get(arg, None) or releases_info.get('sdk-' + arg + '-64bit')
+        release_hash = get_release_hash(arg, releases_info)
         if release_hash:
           if backend is None:
             if version_key(arg) >= (1, 39, 0):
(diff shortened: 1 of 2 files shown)


emsdk.py
```diff
@@ -1707,7 +1727,7 @@ def each_path_exists(pathlist):
       else:
         raise Exception('Unknown custom_is_installed_script directive "' + self.custom_is_installed_script + '"!')
 
-    return content_exists
+    return content_exists and (assume_is_installed_version or self.is_installed_version())
```
sbc100, line 1730: I wonder we can just do `if not content_exist: return False` up on line 1715,

I guess it depends on what those `custom_is_installed_script` things do because they looks like they are currently independent of `content_exists`,

I looks me like if you move the content_exists calculation and the vs-tool handing below the custom_is_installed_script thing, then you can just return early when content_exists is false.

emsdk.py
```diff
@@ -1912,10 +1918,9 @@ def install_tool(self):
 
     # Sanity check that the installation succeeded, and if so, remove unneeded
     # leftover installation files.
-    if self.is_installed():
+    if self.is_installed(assume_is_installed_version=True):
```
sbc100, line 1921: How about calling this argument "skip_version_check" ?

emsdk.py
sbc100: The comment
```
#vs-tool is a special tool since all versions must be installed to the
# same dir, so dir name will not differentiate the version.
```

Makes me thing we can take this further perhaps?  Maybe a followup?

kripken: Maybe possible in a follow. Like you, I'm not sure what those actually are.

emsdk.py
kripken: Good idea, renamed.

PR #509, opened 2020-05-28
Improve behaviour regarding "help"

orestisfl:


emsdk.py
```diff
@@ -2798,7 +2800,7 @@ def main():
        'activate' commands and the invocation of 'emsdk_env', or otherwise
        these commands will default to operating on the default build type
        which in and RelWithDebInfo.''')
-    return 1
+    return exitcode
```
sbc100, line 2803: How about instead we behave like something like ls:

```
$ ls --oijdgf
ls: unrecognized option '--oijdgf'
Try 'ls --help' for more information.
``

i.e. handle the missing command first and return 1.  Then handle the `--help` argument and return 0.  But they don't need to be in the same block.

Comment (orestisfl) [reply emsdk.py:2803]: I am not sure what you are suggesting here, emsdk behaves like ls:

```
./emsdk --oijdgf
Unknown command '--oijdgf' given! Type 'emsdk help' to get a list of commands.
```

exit code 1 (btw I think code 2 is more usual for wrong arguments if you want I can also change this)

Comment (orestisfl) [reply emsdk.py:2803]: And `ls --help --oijdgf` & `./emsdk --help --oijdgf` also behave the same: they show the help and exit with 0

Comment (sbc100) [reply emsdk.py:2803]: Can you refactor to be more like:

if <no_args>:
   print "Missing command; Type 'emsdk help' to get a list of commands."
   exit 1

if <help>:
   print help message
   exit 0
```

Avoids the exitcode variable and I think I prefer not to print the big help message unless the user actually asks for it.


I'm not aware of an exit code convention that says code 2 means bad argument.. is that a thing?

orestisfl: This occurred to me again recently and AFAICT there is no wildly accepted convention. 2 meaning bad argument happens in a lot of cli tools (ls, diff, grep, curl, wget) but many others exit with 1.

sbc100:
Great!

PR #510, opened 2020-05-29
Remove support for non-embedded mode

sbc100:
See https://github.com/emscripten-core/emscripten/issues/9543


sbc100:
Yup.. its been the default for a while now.

Issue #372, opened 2019-10-11
Followup to #368 - Emscripten Docker Image 


trzecieu:
Hi, I can I'm a part of organization on docker. Thanks.

Looks like there are 2 projects: `emscripten/emscripten-ci` and `emscripten/emsdk` - I thought that we will go under `emscripten/emscripten` name. 

If everything is fine let me know so that I can start building and pushing images 1.38.40... (fastcomp + upstream flavors). 

Here is my small story to complete:
~- [ ] build and publish all current versions~
~- [ ] create dockerhub project description~
~- [ ] update [trzeci/emscripten](https://hub.docker.com/r/trzeci/emscripten/) descriptions to advertise using the official image~
~- [x] hook building image CI (starting with own Jenkins)~
update: changed as https://github.com/emscripten-core/emsdk/issues/372#issuecomment-639986878

I see that hooking building Docker Image with my current infrastructure as a temporary solution. Ideally this should be hooked after the official build is published. 

Issue created for visibility after #368, let me know what you think about plan above and if I have a green light. 
CC: @sbc100, @kripken 

Best regards,
Piotr

kripken:
> Looks like there are 2 projects: emscripten/emscripten-ci and emscripten/emsdk - I thought that we will go under emscripten/emscripten name.

Oh, I thought @sbc100 suggested "emscripten/emsdk" for this image - maybe I misunderstood? I don't have a preference either way, but emsdk does make sense I think.

sbc100:
I did agree that `emscripten/emscripten` was ok, but I'm ok with `emscripten/emsdk` too.  in some say it does make more sense.

sbc100:
Regarding CI, please stick to cricleci so that we only have one CI system.

sbc100:
Otherwise sounds great!  Thanks for working on this.

Squareys:
@sbc100 @trzecieu Hey guys, I highly appreciate your efforts of moving @trzecieu's Dockerfiles to an official image!
What's the status on this?

sbc100:
I think the status is that we'd love to see it move forward but its been stalled for a few months.  If yourself or @trzecieu wants to get this going agin it would be more than welcome!

sbc100:
Also, lets drop that fastcomp image in just ship llvm upstream.

trzecieu:
Hi @Squareys, @sbc100 : Thank you for poking and being interested in help, and in general what's the status. I appreciate this. 
[...]

PR #513, opened 2020-06-03
Build docker image on Circle CI

Squareys:
As per https://github.com/emscripten-core/emscripten/issues/11078 and https://github.com/emscripten-core/emsdk/issues/372.

@trzecieu @sbc100 
Very very rough draft for building the docker image on Circle CI.

**TODO**
 - [x] Build docker image
 - [x] Properly specify the tag(s) for docker hub
 - [x] Trigger build only on ~~`emscripten-releases-tags.txt` changes.~~ tags
    - [x] Double check that this is _really_ just triggering on tags.
 - [x] What's the process of getting the docker hub deploy key? They should be specified through secret variables in Circle CI and probably could be done as a final step by one of the project maintainers.
 - ~~Build various variations of the image? (i.e. I remember emscripten-slim and a ubuntu based version in trzeci's original images.)~~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,6 +90,15 @@ jobs:
             source emsdk_env.sh
             python scripts/test.py
 
+  build-docker-image:
+    machine: true
+    steps:
+      - checkout
+      - run: cd docker && make version=${CIRCLE_TAG}-upstream
+      - run: |
+          docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
+          docker push emscripten/emsdk:${CIRCLE_TAG}-upstream
+
 workflows:
   flake8:
     jobs:
@@ -103,3 +112,11 @@ workflows:
   test-windows:
     jobs:
       - test-windows
+  build-docker-image:
+    jobs:
+      - build-docker-image:
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /.*/


.circleci/config.yml
```diff
@@ -90,6 +90,15 @@ jobs:
             source emsdk_env.sh
             python scripts/test.py
 
+  build-docker-image:
+    machine: true
+    steps:
+      - checkout
+      - run: docker build -t emscripten-core/emsdk:$CIRCLE_BRANCH docker/
+      - run: |
+          echo "$DOCKER_PASS" | docker login --username $DOCKER_USER --password-stdin
```
sbc100, line 99: Is this the recommended way to authenticate the push ?  I wonder if we can pass login credentials directly to the push command somehow?

Squareys: This is from the CircleCI docs, yes. It is definitely possible to pass the password directly:

```
docker login -u "$DOCKER_USER" -p "$DOCKER_PASS" ...
```

sbc100: Pattern from the CircleCI docs, seems fine

trzecieu: Looks odd, but also I've found that the recommended - more secure way for login.

.circleci/config.yml
```diff
@@ -90,6 +90,15 @@ jobs:
             source emsdk_env.sh
             python scripts/test.py
 
+  build-docker-image:
+    machine: true
+    steps:
+      - checkout
+      - run: docker build -t emscripten-core/emsdk:$CIRCLE_BRANCH docker/
```
trzecieu, line 97: We do have docker.io organization called [emscripten/emsdk](https://hub.docker.com/repository/docker/emscripten/emsdk/general). 
I don't know who manages this organisation, looks like @haraldreingruber has access to another project [emscripten/emscripten-ci](https://hub.docker.com/repository/docker/emscripten/emscripten-ci/general) - maybe he knows more :)

sbc100: I seem to remember deciding on `emscripten/emsdk` for this image, right?  Or am I remembering wrong?

trzecieu: Most likely you're right. And looking at the name of this repo it makes sense. 
In this case there should be a following action on https://hub.docker.com/orgs/emscripten

Squareys:
>  Trigger build only on emscripten-releases-tags.txt changes.

@sbc100 It looks like whenever this file changes, emsdk is tagged. Could we instead make this job trigger on tags instead? That would be very straight forward to achieve.

.circleci/config.yml
Squareys: > Error: Cannot perform an interactive login from a non TTY device

Looks like `--password-stdin` is not accepted here anyways? 🤔

Squareys:
From my side this is pretty much ready, what's left is:

 - Ensure this is running only on tags (I set up a filter which should make the job ignored for all branches, but included for all tags). The `$CIRCLE_TAG` variable is only set for tags, hence it fails for this PR.
 - Set `DOCKER_PASS` and `DOCKER_USER` secret variables in the Circle CI settings after merge.

So, this is ready to be merged from my side.

.circleci/config.yml
```diff
@@ -90,9 +90,19 @@ jobs:
...
+      - run: cd docker && make version=${CIRCLE_TAG}-upstream
+      - run: |
+          docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
+          docker push emscripten/emsdk:${CIRCLE_TAG}-upstream
+
 workflows:
+  version: 2
   flake8:
-    jobs:
+    jobs:ignore: /.*/
```
sbc100, line 105: What does this do?

.circleci/config.yml
```diff
@@ -90,9 +90,19 @@ jobs:
             source emsdk_env.sh
             python scripts/test.py
 
+  build-docker-image:
+    machine: true
```
sbc100, line 94: Why this?

.circleci/config.yml
```diff
@@ -90,9 +90,19 @@ jobs:
...
+    steps:
+      - checkout
+      - run: cd docker && make version=${CIRCLE_TAG}-upstream
+      - run: |
```
sbc100, line 98: Does this run at the top level, or within the `docker` subdirectory?   (i.e. does the `cd` in the previous run apply to this run too?

.circleci/config.yml
Squareys: This uses a docker machine executor which is required to build docker images.
https://circleci.com/docs/2.0/examples-intro/#linux-with-machine

.circleci/config.yml
Squareys: Sorry, that was a pretty sloppy mistake on my behalf

.circleci/config.yml
Squareys: I am pretty sure that docker push and login do not require the Docker file but use the tag to reference the build docker image stored in the local registry. Right, @trzecieu?

trzecieu: That's right, push and login can be done from any place.

.circleci/config.yml
haraldreingruber: The org (https://hub.docker.com/orgs/emscripten) was created by @kripken, right?
It's odd, but I don't see any other members...

Squareys:
@sbc100 I fixed or commented on the issues you found. Tell me if there is anything else I can do.

sbc100:
lgtm.

What are the next steps?  It looks like the CI is currently failing... and I guess we need to figure out how to inject the credentials?

Squareys:
Injecting the credentials should be done after merge by someone who has access to the Circle CI backend and the docker hub namespace (@kripken? See https://github.com/emscripten-core/emsdk/pull/513#discussion_r435738602).

The CI is failing because it is not running for a tag, hence `CIRCLE_TAG` variable is missing aswell. Once the PR is merged, it will only run on tags.

sbc100:
> The CI is failing because it is not running for a tag, hence `CIRCLE_TAG` variable is missing aswell. Once the PR is merged, it will only run on tags.

It seems a little odd to me that this would run on this PR at all.. if its only supposed to run on tags.. why is it running here?

PR #514, opened 2020-06-04
Add host network for building image

trzecieu:
Without `--network host` It's impossible to build images (might be some recent addition for docker engine, as I don't recall it was a case before)

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -11,7 +11,7 @@ ifndef version
 endif
 
 build: .TEST
-	docker build --build-arg=EMSCRIPTEN_VERSION=${version} --tag emscripten/emsdk:${version} .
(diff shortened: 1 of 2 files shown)


kripken:
The docker docs say this flag only works on linux. Is that the only place this is used?

sbc100:
We shouldn't be running docker commands anywhere except on linux.

kripken:
sgtm then.

trzecieu:
@kripken, it's supported on windows as well in my version: 
```
PS C:\Users\piotr> docker build --help | grep net
      --network string          Set the networking mode for the RUN
PS C:\Users\piotr> docker --version
Docker version 19.03.8, build afacb8b
```

Thx for merging !

Issue #4, opened 2014-04-25
SDK installer may break software depending on PATH environment variable


oyshole:
It seems when doing a full install using emsdk-1.16.0-full-64bit.exe the PATH environment variable on Windows is changed in ways that might break working software in some ways.

In my scramble to detect what happened i have not recorded the exact values before/during/after the installation, but the scenario was as follows;

After installations all previous paths in the PATH variable had their backslashes turned to slashes (\ turned to /). Following these paths, all Emscripten paths were added using slashes, and then all Emscripten paths were repeated, this time with backslashes.

At least on my system, this broke several other pieces of software, probably using older libraries not supporting slashes on Windows etc.

System config basics: Windows 7 64bit SP1, Visual Studio 2013, emsdk-1.16.0-full-64bit.exe.

PR #515, opened 2020-06-04
Fix Path,EMSDK and EM_CACHE not being set permanently when using --global

kudaba:
Going through the initial setup of emscripten the first time, the first thing I wanted to do was set it up globally to use in visual studio and found that some of the environment variables are not getting applied properly. There are still additional issues with environment variables however since EMSDK_NODE and EMSDK_PYTHON  and PATH will get installed system wide while EM_CACHE and EMSDK will only get installed to user.

FYI: I also ran into some reported issues during my initial setup that didn't block me in the end:
* pywin32 was missing, so I needed to pip install it, maybe add a not in the windows setup documentation.
* ``ERROR: The system was unable to find the specified registry key or value.`` which are not actual issues but should probably be suppressed (trying to delete registry key that doesn't exist). Happens on every activate call.
* ``(1159, 'PostMessage', 'The message can be used only with synchronous operations.')`` Only happens when system registry keys are not set, it doesn't cause them to fail though.

Edit:
So, a big update here. While the code remains to be able to set user environment variables in windows it really only runs in two modes: current session locals or system wide environment variables. I removed the code that was trying to delete the variables and causing errors since it doesn't really do anything.

I removed setx and the option to permanently activate from construct_env, but it's still used as a fallback to set registry variables where it already had the 1024 check in place.

I fixed the PostMessage error that was happening by using SendMessageWithTimeout instead. It only sends the message if a registry value actually changed. I also confirmed that the message is being received by other applications.

The one caveat that still exists is that you need to run emsdk activate before running emsdk activate --global so that it will use the correct python version.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -334,7 +334,7 @@ def win_get_environment_variable(key, system=True, fallback=True):
         folder = winreg.OpenKey(winreg.HKEY_CURRENT_USER, 'Environment')
       value = str(winreg.QueryValueEx(folder, key)[0])
     except Exception:
-      # PyWin32 is not available - read via os.environ. This has the drawback
+      # If reading registry fails for some reason - read via os.environ. This has the drawback
       # that expansion items such as %PROGRAMFILES% will have been expanded, so
       # need to be precise not to set these back to system registry, or
       # expansion items would be lost.
@@ -346,14 +346,11 @@ def win_get_environment_variable(key, system=True, fallback=True):
         folder.Close()
 
   except Exception as e:
+    # this catch is if both the registry key threw an exception and the key is not in os.environ
     if e.args[0] != 2:
       # 'The system cannot find the file specified.'
       errlog('Failed to read environment variable ' + key + ':')
       errlog(str(e))
-    try:
-      folder.Close()
-    except Exception:
-      pass
     return None
   return value
 
@@ -2581,7 +2578,6 @@ def construct_env_with_vars(env_vars_to_add):
         else:
           assert False
 
-    errlog('')
   return env_string
 
 
--- a/emsdk.py
+++ b/emsdk.py
@@ -299,6 +299,7 @@ def remove_tree(d):
 
 
 def win_set_environment_variable_direct(key, value, system=True):
+  folder = None
   try:
     if system:
       # Read globally from ALL USERS section.
@@ -308,23 +309,22 @@ def win_set_environment_variable_direct(key, value, system=True):
       folder = winreg.OpenKeyEx(winreg.HKEY_CURRENT_USER, 'Environment', 0, winreg.KEY_ALL_ACCESS)
     winreg.SetValueEx(folder, key, 0, winreg.REG_EXPAND_SZ, value)
     debug_print('Set key=' + key + ' with value ' + value + ' in registry.')
+    return True
   except Exception as e:
     # 'Access is denied.'
-    if e.args[0] == 5:
+    if e.args[3] == 5:
       exit_with_error('Error! Failed to set the environment variable \'' + key + '\'! Setting environment variables permanently requires administrator access. Please rerun this command with administrative privileges. This can be done for example by holding down the Ctrl and Shift keys while opening a command prompt in start menu.')
     errlog('Failed to write environment variable ' + key + ':')
     errlog(str(e))
-    folder.Close()
-    return None
-
-  folder.Close()
-  HWND_BROADCAST = ctypes.wintypes.HWND(0xFFFF)  # win32con.HWND_BROADCAST == 65535
-  WM_SETTINGCHANGE = 0x001A  # win32con.WM_SETTINGCHANGE == 26
-  ctypes.windll.user32.SendMessageA(HWND_BROADCAST, WM_SETTINGCHANGE, 0, 'Environment')
+    return False
+  finally:
+    if folder is not None:
+      folder.Close()
 
 
-def win_get_environment_variable(key, system=True):
+def win_get_environment_variable(key, system=True, fallback=True):
   try:
+    folder = None
     try:
       if system:
         # Read globally from ALL USERS section.
@@ -338,7 +338,13 @@ def win_get_environment_variable(key, system=True):
       # that expansion items such as %PROGRAMFILES% will have been expanded, so
       # need to be precise not to set these back to system registry, or
       # expansion items would be lost.
-      return os.environ[key]
+      if fallback:
+        return os.environ[key]
+      return None
+    finally:
+      if folder is not None:
+        folder.Close()
+
   except Exception as e:
     if e.args[0] != 2:
       # 'The system cannot find the file specified.'
@@ -349,7 +355,6 @@ def win_get_environment_variable(key, system=True):
     except Exception:
       pass
     return None
-  folder.Close()
   return value
 
 
@@ -367,11 +372,11 @@ def win_get_active_environment_variable(key):
 
 def win_set_environment_variable(key, value, system=True):
   debug_print('set ' + str(key) + '=' + str(value) + ', in system=' + str(system), file=sys.stderr)
-  previous_value = win_get_environment_variable(key, system)
+  previous_value = win_get_environment_variable(key, system, fallback=False)
   if previous_value == value:
     debug_print('  no need to set, since same value already exists.')
     # No need to elevate UAC for nothing to set the same value, skip.
-    return
+    return False
 
   if not value:
     try:
@@ -382,13 +387,12 @@ def win_set_environment_variable(key, value, system=True):
       debug_print(str(cmd))
       value = subprocess.call(cmd, stdout=subprocess.PIPE)
     except Exception:
-      return
-    return
+      return False
+    return True
 
   try:
-    if system:
-      win_set_environment_variable_direct(key, value, system)
-      return
+    if win_set_environment_variable_direct(key, value, system):
+      return True
     # Escape % signs so that we don't expand references to environment variables.
     value = value.replace('%', '^%')
     if len(value) >= 1024:
@@ -398,15 +402,53 @@ def win_set_environment_variable(key, value, system=True):
     retcode = subprocess.call(cmd, stdout=subprocess.PIPE)
     if retcode != 0:
       errlog('ERROR! Failed to set environment variable ' + key + '=' + value + '. You may need to set it manually.')
+    else:
+      return True
   except Exception as e:
     errlog('ERROR! Failed to set environment variable ' + key + '=' + value + ':')
     errlog(str(e))
     errlog('You may need to set it manually.')
 
+  return False
+
+
+def win_set_environment_variables(env_vars_to_add, system):
+  if not env_vars_to_add:
+    return
+
+  changed = False
+
+  for key, value in env_vars_to_add:
+    if win_set_environment_variable(key, value, system):
+      if not changed:
+        changed = True
+        print('Setting global environment variables:')
+
+      print(key + ' = ' + value)
+
+  if not changed:
+    print('Global environment variables up to date')
+    return
+
+  # if changes were made then we need to notify other processes
+  try:
+    HWND_BROADCAST = ctypes.wintypes.HWND(0xFFFF)  # win32con.HWND_BROADCAST == 65535
+    WM_SETTINGCHANGE = 0x001A  # win32con.WM_SETTINGCHANGE == 26
+    SMTO_BLOCK = 0x0001  # win32con.SMTO_BLOCK == 1
+    ctypes.windll.user32.SendMessageTimeoutA(
+      HWND_BROADCAST,    # hWnd: notify everyone
+      WM_SETTINGCHANGE,  # Msg: registry changed
+      0,                 # wParam: Must be 0 when setting changed is sent by users
+      'Environment',     # lParam: Specifically environment variables changed
+      SMTO_BLOCK,        # fuFlags: Wait for message to be sent or timeout
+      100)               # uTimeout: 100ms
+  except Exception as e:
+    errlog('SendMessageTimeout failed with error: ' + str(e))
+
 
 def win_delete_environment_variable(key, system=True):
   debug_print('win_delete_environment_variable(key=' + key + ', system=' + str(system) + ')')
-  win_set_environment_variable(key, None, system)
+  return win_set_environment_variable(key, None, system)
 
 
 # Returns the absolute pathname to the given path inside the Emscripten SDK.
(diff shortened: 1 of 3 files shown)


sbc100:
Regarding pywin32.  This should now be included in the version of python that is downloaded by emsdk.  At least as of #487..  can you confirm?

sbc100:
Seems like a good change to me.   @juj?

emsdk.py
```diff
@@ -2486,7 +2486,7 @@ def set_active_tools(tools_to_activate, permanently_activate):
 
   # Construct a .bat script that will be invoked to set env. vars and PATH
   if WINDOWS:
-    env_string = construct_env(tools_to_activate, False)
+    env_string = construct_env(tools_to_activate, permanently_activate)
```
juj, line 2489: Hmm, there are two issues with this:

1. `construct_env` is intended to create the `emsdk_env.bat` script that one can use to enter the Emsdk-enabled command prompt locally. It should never globally set environment variables.
2. Not a fault of this PR, but `construct_env` should actually remove the permanently_activate=true support. In current emsdk, it is always called with permanently_activate=False, to avoid a catastrophic bug where `%PROGRAMFILES%` etc. wildcards were getting expanded in PATH. (and also https://github.com/emscripten-core/emsdk/issues/4). The rule of thumb is that one should never be using the `SETX` command to set system PATH.

The right location to do this would be later in the code block starting from the comment
`# Apply environment variables to global all users section.`, looks like that is missing the relevant registrations.

sbc100:
This change is a lot bigger now :(   Can it be simplified or broken down into smaller chunks?   I'm having trouble groking all the moving parts there.

emsdk.py
kudaba: So, a big update here. While the code remains to be able to set user environment variables in windows it really only runs in two modes: current session locals or system wide environment variables. I removed the code that was trying to delete the variables and causing errors since it doesn't really do anything.

I removed setx and the option to permanently activate from construct_env, but it's still used as a fallback to set registry variables where it already had the 1024 check in place.

I fixed the PostMessage error that was happening by using SendMessageWithTimeout instead. It only sends the message if a registry value actually changed. I also confirmed that the message is being received by other applications.

The one caveat that still exists is that you need to run ``emsdk activate`` before running ``emsdk activate --global`` so that it will use the correct python version.

emsdk.py
```diff
@@ -3023,7 +3016,7 @@ def print_tools(t):
       outfile = sys.argv[2]
     tools_to_activate = currently_active_tools()
     tools_to_activate = process_tool_list(tools_to_activate, log_errors=True)
-    env_string = construct_env(tools_to_activate, len(sys.argv) >= 3 and 'perm' in sys.argv[2])
```
sbc100, line 3026: I also noticed this strange line.  I think we can maybe split this out as a separate change to simplify this one?   As far as I know this "perm" argument is not documented or used, right @juj?

kudaba: That and I removed the secondary argument for that function.

sbc100: I split that part out into separate change #525, in the hope that it will simplify this one.    If you'd rather do that yourself (to avoid re-basing on top of my change) please go ahead.

sbc100:
If there are part of this change that are cleanups or re-structuring only perhaps you could submit those separately?

Also, perhaps the PR description could be update to reflect exactly what this change is supposed to do?  Or is it up-to-date already?

kudaba:
Updated the PR with the info of all the changes. I could go through an restructure it, unfortunately I won't have the time to get back to this for a while.

sbc100:
How does this relate to https://github.com/emscripten-core/emsdk/pull/468  ?

kudaba:
There will be merge conflicts but they should be able to co-exists.

aminya:
Can we get this rebased? This will probably fix the windows tests.

emsdk.py
```diff
@@ -2584,10 +2591,9 @@ def construct_env(tools_to_activate):
         else:
           assert False
 
-      errlog(key + ' = ' + value)
+    errlog('')
```
sbc100, line 2596: Is this added an extra empty line to the output?

emsdk.py
```diff
@@ -2584,10 +2591,9 @@ def construct_env(tools_to_activate):
...
   return env_string
 
-
 def error_on_missing_tool(name):
```
sbc100: Doesn't flake8 complain about this if you don't have two lines between each function?

emsdk.py
```diff
@@ -2533,39 +2532,47 @@ def construct_env(tools_to_activate):
...
+
+  return env_vars_to_add
+
+
+def construct_env(tools_to_activate):
+  env_vars_to_add = get_env_vars_to_add(tools_to_activate)
 
-  if env_vars:
+  env_string = ''
+  if len(env_vars_to_add) > 0:
```
sbc100, line 2560: Just `if env_vars_to_add:` should be fine here.

emsdk.py
```diff
@@ -2421,20 +2417,32 @@ def set_active_tools(tools_to_activate, permanently_activate, system):
...
+
+            print(key + ' = ' + value)
+
+        if not changed:
+          print('Global environment variables up to date')
+        else:      
+          try:
+            import win32gui
+            import win32con
+            win32gui.SendMessageTimeout(win32con.HWND_BROADCAST, win32con.WM_SETTINGCHANGE, 0, 'Environment', win32con.SMTO_BLOCK, 100)
```
sbc100, line 2444: Split the params onto separate lines to avoid the very long line here?

emsdk.py
```diff
@@ -2421,20 +2418,32 @@ def set_active_tools(tools_to_activate, permanently_activate, system):
...
+      write_set_env_script(env_string)
+    else:
+      env_vars_to_add = get_env_vars_to_add(tools_to_activate)
 
-  # Apply environment variables to global all users section.
-  if WINDOWS and permanently_activate:
-    # Individual env. vars
-    for tool in tools_to_activate:
-      tool.win_activate_env_vars(system=system)
+      if len(env_vars_to_add) > 0:
```
sbc100, line 2427: Just `if env_vars_to_add:` should be fine here.

emsdk.py
```diff
@@ -2421,20 +2418,32 @@ def set_active_tools(tools_to_activate, permanently_activate, system):
...
+
+        if not changed:
+          print('Global environment variables up to date')
+        else:
+          try:
+            import win32gui
+            import win32con
+            win32gui.SendMessageTimeout(win32con.HWND_BROADCAST, win32con.WM_SETTINGCHANGE, 0, 'Environment', win32con.SMTO_BLOCK, 100)
+          except Exception as e:
+            errlog('SendMessageTimeout failed with error: ' + str(e))
```
sbc100, line 2446: Might be nice to keep this entire block as a seperate function.   Then you could avoid a lot of the nesting by having early returns after `if not changed:` and after `if env_vars_to_add:`?

emsdk.py
kudaba: yeah, just to make formatting a bit nicer

emsdk.py
kudaba: I force pushed the formatting fixes from flake

aminya (approved):
I tested this. It works! 

I recommend enabling the tests:
https://github.com/kudaba/emsdk/blob/ca1e15fb6a66c4d62e2a72a27cbff371b4e2a09f/.circleci/config.yml#L95
```diff
          command: |
-            try {
               scripts/test_permanent.ps1
-            } catch {
-             Write-Host "`n Error Message: [$($_.Exception.Message)"] -ForegroundColor Red
-              Write-Host "`n Line Number: "$_.InvocationInfo.ScriptLineNumber -ForegroundColor Red
-            }
```

aminya:
I am happy to say that I tested this in combination with #635, and all the `scripts/test_permanent.ps1` tests finally pass! 

We should definitely enable them as I showed in https://github.com/emscripten-core/emsdk/pull/515#pullrequestreview-503497305

After #635 is merged there is a small conflict that needs to be fixed. I removed the following lines in that PR, so after accepting your changes just remove this line:
```
prev_path = os.environ['PATH']
```

kudaba:
Sorry for all the force pushes, it took me a while to get local testing working.

I fixed up the exception handling with the registry and had to fix up the permanent test script a bit as well since it assumed emsdk was being installed to a folder called emsdk.

I did notice that the registry handling is a lot slower now, but since it's generally a one time thing it's probably not anything worth worrying about.

aminya:
Awesome! 

I did not notice any performance difference. Anyways, I think stability is more important. 😃

aminya:
@sbc100 Can we roll this in?

emsdk.py
sbc100: But why do you say this is nicer?   Right no I see the following output on my machine:

```
sbc.sbc-glaptop (~/dev/wasm/emsdk) ((53c184e...)) $ . ./emsdk_env.sh 
Adding directories to PATH:
PATH += /home/sbc/dev/wasm/emsdk
PATH += /home/sbc/dev/wasm/emsdk/upstream/emscripten
PATH += /home/sbc/dev/wasm/emsdk/node/12.18.1_64bit/bin

EMSDK = /home/sbc/dev/wasm/emsdk
EM_CONFIG = /home/sbc/dev/wasm/emsdk/.emscripten
EM_CACHE = /home/sbc/dev/wasm/emsdk/upstream/emscripten/cache
EMSDK_NODE = /home/sbc/dev/wasm/emsdk/node/12.18.1_64bit/bin/node
sbc.sbc-glaptop (~/dev/wasm/emsdk) ((53c184e...)) $
```

To me an extra empty line at the end would look pointless/messy.. but maybe that is just me?

sbc100 (commented):
I think we still have the problem that we will clobber any local changes that the user has in their PATH with a fresh copy from the registry... but that is separate issue I think.

emsdk.py
```diff
@@ -338,7 +338,13 @@ def win_get_environment_variable(key, system=True):
...
-      return os.environ[key]
+      if fallback:
+        return os.environ[key]
+      return None
+    finally:
+      if folder is not None:
+        folder.Close()
+
   except Exception as e:
     if e.args[0] != 2:
```
sbc100, line 349: I'd love to know under what circumstances these two excpt blocks can fire?   Do we really need two nested try blocks here?  Again, not really related to this PR.

sbc100 (approved):

emsdk.py
```diff
@@ -398,15 +402,53 @@ def win_set_environment_variable(key, value, system=True):
...
+    WM_SETTINGCHANGE = 0x001A  # win32con.WM_SETTINGCHANGE == 26
+    SMTO_BLOCK = 0x0001  # win32con.SMTO_BLOCK == 1
+    ctypes.windll.user32.SendMessageTimeoutA(
+      HWND_BROADCAST,    # hWnd: notify everyone
+      WM_SETTINGCHANGE,  # Msg: registry changed
+      0,                 # wParam: Must be 0 when setting changed is sent by users
+      'Environment',     # lParam: Specifically environment variables changed
+      SMTO_BLOCK,        # fuFlags: Wait for message to be sent or timeout
+      100)               # uTimeout: 100ms
+  except Exception as e:
```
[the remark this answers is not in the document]
aminya: @kudaba has added comments for these. The original code did not have any description
https://github.com/emscripten-core/emsdk/pull/515/files#diff-7ecf1f48f2952ff372972476842c8882L321

emsdk.py
```diff
@@ -338,7 +338,13 @@ def win_get_environment_variable(key, system=True):
       # that expansion items such as %PROGRAMFILES% will have been expanded, so
```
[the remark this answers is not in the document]
aminya: `winreg.OpenKey` is not "noexcept", so it may throw. The documentation says the type of error would be OSError.
https://docs.python.org/3/library/winreg.html#winreg.OpenKey

emsdk.py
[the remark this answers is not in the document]
sbc100: I mean more specific in the type of exceptions we catch here?  Can it be just OSError perhaps?

emsdk.py
kudaba: Oh, I can clean that up, I think it was there to try to catch exceptions from opening the registry key. I almost did the same in set variable but didn't notice that the get was already doubled up.

emsdk.py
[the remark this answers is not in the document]
aminya: I don't it hurts to catch all the exceptions. We are not concerned with the performance of this program that runs on a powerful desktop processor. The stability is more important, IMO.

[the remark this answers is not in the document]
kudaba: I'm not sure how error handling or exceptions are working with the interop between python and win32 so this is just me being a bit extra cautious.

kudaba:
> I think we still have the problem that we will clobber any local changes that the user has in their PATH with a fresh copy from the registry... but that is separate issue I think.

Actually I think this will be fine. It adds emsdk paths to the local path environment variable and the registry setting separately so it should be additive in both cases. I had to do that since just setting the registry setting was not actually changing the currently set environment variable.

sbc100:
> > I think we still have the problem that we will clobber any local changes that the user has in their PATH with a fresh copy from the registry... but that is separate issue I think.
> 
> Actually I think this will be fine. It adds emsdk paths to the local path environment variable and the registry setting separately so it should be additive in both cases. I had to do that since just setting the registry setting was not actually changing the currently set environment variable.

The problem is not about addings to the current env.   I believe the problem is that we always take the registry value as the *initial* PATH to add to (we've add the concept of fallback here but I think in practice it won't fire).   This means that you always end up with "registry value + extras" rather than "current shell value + extras".  

Basically we need to calculate two different paths which can be different:

1. What path should be added the registery.
2. What path should be set in the current shell

Again, its not something we need to fix as part of this PR, I guess its bug we have always had.

emsdk.py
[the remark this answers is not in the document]
sbc100: Catching specific exceptions is not about performance but about code quality and maintainability.   In general we want to minimize the size of the try block and be as specific as possible in the exceptions we catch.  

From pep8:  `When catching exceptions, mention specific exceptions whenever possible instead of using a bare except: clause:`

One problem with catching everything is that it can hide all kinds of programmer errors. e.g. simple things like accessing out of bounds which the except block is almost never supposed to handle.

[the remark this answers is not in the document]
sbc100: Anyway, unrelated to this PR so we don't need to block on this.

sbc100:
I think this change is ready to land now.

Can you craft as suitable message for the squashed commit?    I would use the PR description but I'm not sure which parts are still relevant?

kudaba:
The first commit has the most accurate messaging I think:

Fix environment variables to be consistently set when using --system or --permanent
Switch PostMessage with SendMessageWithTimeout and only send one if environment variables actually changed
Enable and fix permanent unit tests

Or did you mean you want me to squash the review feeback commit back down into the original (I had been doing it this whole time until the last bit of feedback).

sbc100:
Thanks!

PR #523, opened 2020-06-14
Simplify docker image in several ways

sbc100:
- `--embeddded` mode is the now the default
- embedded cache and ports is also the default
- emsdk now fully polulates the cache by default so no need to do that.
- always "activate" the SDK avoiding that need things to work without
  that.
- avoid making extra symlinks.
- remove python2


sbc100:
I'll probably wait until we have at least one version uploaded in dockerhub before landing this change.

sbc100:
@trzecieu

docker/Dockerfile
```diff
@@ -28,70 +23,30 @@ RUN echo "## Start building" \
...
-
-# Create a structure and make mutable folders accessible for r/w
-RUN cd ${EMSDK} \
-    &&  echo "## Create .data structure" \
-    &&  for mutable_dir in ${EM_DATA} ${EM_PORTS} ${EM_CACHE} ${EMSDK}/zips ${EMSDK}/tmp; do \
-            mkdir -p ${mutable_dir}; \
-            chmod -R 777 ${mutable_dir}; \
-        done \
-        \
+    &&  chmod -R 777 ${EMSDK} \
```
trzecieu, line 44: Previously all mutable folders were put inside `.data` folder to avoid changing recursively entire emsdk folder.

docker/Dockerfile
```diff
@@ -28,70 +23,30 @@ RUN echo "## Start building" \
...
 &&  echo "## Done"
 
 # Create an entrypoint that activates Emscripten SDK and helps running this image as non-root user
 COPY entrypoint ${EMSDK}/
 
-
-# Create symbolic links for critical Emscripten Tools
```
trzecieu, line 79: This change was fixing a problem if docker image was used on some of CI platform (CircleCI or Travis) and that platform by default was replacing entrypoint so we couldn't rely on activating emscripten prior to compilation. 

I think that the change is in general fine - but that will reduce applicability of the image.

docker/Dockerfile
```diff
@@ -100,36 +55,21 @@ RUN echo "## Aggresive optimization: Remove debug symbols" \
     # Remove debugging symbols from embedded node (extra 7MB)
     && strip -s `which node` \
     # Tests consume ~80MB disc space
-    && rm -fr ${EMSDK}/llvm/clang/emscripten/tests \
+    && rm -fr ${EMSDK}/upstreeam/emscripten/tests \
```
trzecieu, line 58: Double `e`
```suggestion
    && rm -fr ${EMSDK}/upstream/emscripten/tests \
```

docker/Dockerfile
```diff
@@ -142,27 +82,21 @@ COPY --from=stage_build /emsdk /emsdk
...
-# host machine, and in this case running any docker image will cause other random problems (mostly due `$HOME` pointing to `/`)
-# This extra user works nicely with entrypoint provided in `/emsdk/entrypoint` as it detects case explained before.
+# Thanks to that this image can be executed as non-root user and created files
+# will not require root access level on host machine Please note that this
+# solution even if widely spread (i.e. Node.js uses it) is far from perfect as
+# user 1000:1000 might not exist on host machine, and in this case running any
+# docker image will cause other random problems (mostly due `$HOME` pointing to
+# `/`)
+# This extra user works nicely with entrypoint provided in `/emsdk/entrypoint`
+# as it detects case explained before.
```
trzecieu, line 99: This over-complicated entrypoint was created to support running the image as non-root user. (We've found that just easier in some Jenkins pipelines) so that non-root user can wipe out created files from mounted volume without having `sudo`. 
I believe that once `chmod -R 777 ${EMSDK}` was performed, the folder structure wasn't fully created. Like I remember that at that moment no ports were downloaded (or cache for them). 
Then trying to compile using the image will fail as no access to write to a folder. 

That leads to the point of taking a decision if you would like to support this scenario.

docker/Dockerfile
sbc100: Do we care if the user has write access to the emsdk directory?
 
I can update to just be the cache directory if we do.

docker/Dockerfile
sbc100: Oh.. I think I can replace this with just adding emscripten to the PATH.. that should do it.

docker/Dockerfile
sbc100: Updated to be just "emsdk/cache"  (which is where the ports and the cached libraries live).

docker/Dockerfile
sbc100: OK, I updated to rely only on PATH.. with no need to activate.

Seems better to have just a single way do things, no ?   Plus it simplifies the entry point somewhat.

docker/Dockerfile
sbc100: Ah I see.   

I think we can continue to support that use case.   I've added "umask 777" to /root/.bashrc and I think we can just remove the special entrypoint.

See what you think.

sbc100:
@trzecieu what do you think?   If you prefer I can split this into several parts and we can evaluate each on separately?

trzecieu:
Hi, I'll add some tests, so that it will be much easier to verify whether it works, and it will improve maintainability. 
I think we can manage this at once

docker/Dockerfile
```diff
@@ -100,36 +51,21 @@ RUN echo "## Aggresive optimization: Remove debug symbols" \
...
-    &&  chmod -R 777 ${EM_DATA} \
-    \
-    # cleanup
-    &&  find ${EMSDK} -name "*.pyc" -exec rm {} \; \
-    \
-    &&  echo "## Done"
+# Generate sanity
+# TODO(sbc): We should be able to use just emcc -v here but it doesn't
+# currently create the sanity file.
+RUN echo "## Generate sanify" \
```
trzecieu, line 63: Spelling 
```suggestion
RUN echo "## Generate sanity" \
```

PR #525, opened 2020-06-17
Remove unused permanent in construct_env

sbc100:
This was only ever set to true if the user knew to pass an
undocumented extra argument called "perm".

Split out from #515


sbc100:
I don't think this was ever how construct_env was supposed to work.  Right @juj ?  Can you imagine any users of this?

PR #526, opened 2020-06-19
Updated make file for docker building to perform extra sanity tests

trzecieu:
Extension to: https://github.com/emscripten-core/emsdk/pull/523
I've added tests on top of your branch, so that it will be easier to maintain this image. 

CC: @sbc100:

PR #527, opened 2020-06-19
Build docker image as part of CI

sbc100:
We only publish on tag, but we want to build and test the image
on all PRs.

See #526


trzecieu:
Nice! Once another PR gets merged, this could move to `make` commands and utilize newly added `make ... test` on each commit.

PR #530, opened 2020-06-24
Drop "-upstream" suffix from docker image name

sbc100:


sbc100:
@trzecieu

sbc100:
Also remove the fastcomp binaries which we currently ship in the upstream SDK for historical reasons (this will save a bunch of space).

PR #532, opened 2020-06-24
Update node directory in Dockerfile

sbc100:


sbc100:
It did pass.. I think I added docker testing to the CI in a change the crossed with yours.

PR #533, opened 2020-06-24
Updated make file for docker building to perform extra sanity tests

trzecieu:
Reapply #526

This change:
* Adds extra sanity tests for Makefile, so that it will be easier to maintain the image
* Switches to `make` commands for CircleCI, and adds `test` step
* Adds small changes to RADME.md to better reflect how to build image locally

To be merged after: https://github.com/emscripten-core/emsdk/pull/530

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,12 +98,13 @@ jobs:
           name: install docker
           command: apt-get update -q && apt-get install -q -y docker.io
       - setup_remote_docker
+      # build and test the latest released version of EMSDK
       - run:
           name: build
-          command: make -C ./docker version=${CIRCLE_TAG} build
+          command: make -C ./docker version=$(git describe --tags --abbrev=0) build
       - run:
           name: test
-          command: make -C ./docker version=${CIRCLE_TAG} test
+          command: make -C ./docker version=$(git describe --tags --abbrev=0) test
 
   publish-docker-image:
     executor: bionic
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,7 +100,10 @@ jobs:
       - setup_remote_docker
       - run:
           name: build
-          command: docker build --network host ./docker
+          command: make -C ./docker version=${CIRCLE_TAG} build
+      - run:
+          name: test
+          command: make -C ./docker version=${CIRCLE_TAG} test
 
   publish-docker-image:
     executor: bionic
@@ -112,16 +115,15 @@ jobs:
       - setup_remote_docker
       - run:
           name: build
-          command: docker build --network host --build-arg EMSCRIPTEN_VERSION=${CIRCLE_TAG} --tag emscripten/emsdk:${CIRCLE_TAG} ./docker
+          command: make -C ./docker version=${CIRCLE_TAG} build
       - run:
-          name: tag image
-          command: docker tag emscripten/emsdk:${CIRCLE_TAG} emscripten/emsdk:latest
+          name: test
+          command: make -C ./docker version=${CIRCLE_TAG} test
       - run:
           name: push image
           command: |
             docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
-            docker push emscripten/emsdk:${CIRCLE_TAG}
-            docker push emscripten/emsdk:latest
+            make -C ./docker version=${CIRCLE_TAG} alias=latest push
 
 workflows:
   flake8:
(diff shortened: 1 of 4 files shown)


sbc100 (approved):
lgtm % testing in CI without a CIRCLE_TAG

.circleci/config.yml
```diff
@@ -100,7 +100,10 @@ jobs:
       - setup_remote_docker
       - run:
           name: build
-          command: docker build --network host ./docker
+          command: make -C ./docker version=${CIRCLE_TAG} build
+      - run:
+          name: test
+          command: make -C ./docker version=${CIRCLE_TAG} test
```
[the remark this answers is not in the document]
trzecieu: Indeed just found that on the portal, added some fix for that in following commit.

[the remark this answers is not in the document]
trzecieu: Looks like works fine that. So basically now it will always build against the latest tagged version.

trzecieu:
@sbc100 after some fighting with CircleCi I was able to reproduce an original reason why I've added entrypoint:
```
# compilation as non-root
docker run --rm -u `id -u`:`id -g` -v `pwd`:/src -w /src emscripten/emsdk:1.39.18 \
		bash -c "\
			mkdir -p .test \
			&& echo 'int main() { return 0; }' > .test/main.c \
			&& emcc -c .test/main.c -o .test/main\
		"
  File "/emsdk/upstream/emscripten/emcc.py", line 42, in <module>
    import emscripten
  File "/emsdk/upstream/emscripten/emscripten.py", line 24, in <module>
    from tools import building
  File "/emsdk/upstream/emscripten/tools/building.py", line 23, in <module>
    from . import shared
  File "/emsdk/upstream/emscripten/tools/shared.py", line 1801, in <module>
    check_vanilla()
  File "/emsdk/upstream/emscripten/tools/shared.py", line 698, in check_vanilla
    is_vanilla_file = temp_cache.get('is_vanilla.txt', get_vanilla_file)
  File "/emsdk/upstream/emscripten/tools/cache.py", line 108, in get
    self.acquire_cache_lock()
  File "/emsdk/upstream/emscripten/tools/cache.py", line 55, in acquire_cache_lock
    self.filelock.acquire(60)
  File "/emsdk/upstream/emscripten/tools/filelock.py", line 240, in acquire
    self._acquire()
  File "/emsdk/upstream/emscripten/tools/filelock.py", line 360, in _acquire
    fd = os.open(self._lock_file, open_mode)
PermissionError: [Errno 13] Permission denied: '/.emscripten_cache.lock'
make: *** [test] Error 1
```

This is result of non-standard user used in Circle Ci (that is not 1000:1000). 

Should I revert entrypoint and add support for that?

sbc100:
I think the fix is that the emscripten root needs to be writable due to the code in emscripten that checks if the emscripten root is writable.

I think adding this line should fix it:

```
&&  chmod 777 ${EMSDK}/upstream/emscripten \ 
```

We should probably update the emscripten code so that only the cache directory (and not the containing directory need to be writable).

sbc100 (approved):

docker/Makefile
```diff
@@ -13,6 +13,31 @@ endif
 build: .TEST
 	docker build --network host --build-arg=EMSCRIPTEN_VERSION=${version} --tag emscripten/emsdk:${version} .
 
+test: .TEST
+	# compilation as non-root
+	docker run --rm -u `id -u`:`id -g` -e HOME=/tmp -v `pwd`:/src -w /src emscripten/emsdk:${version} \
+		bash -c "\
+			mkdir -p .test \
```
sbc100, line 20: Why the `.` prefix here?   Why not just test?

docker/Makefile
```diff
@@ -13,6 +13,31 @@ endif
...
+			&& emcc -c .test/main.c -o .test/main\
+		"
+	# artifact should be removable from host by non-root
+	rm -fr .test
+
+	# running as non-root and forcing downloading ports
+	docker run --rm -u `id -u`:`id -g` -e HOME=/tmp --net=host emscripten/emsdk:${version} embuilder build zlib
+
+	# compilation without entrypoint
+	docker run --rm -e /bin/bash -v `pwd`:/src -w /src emscripten/emsdk:${version} \
```
sbc100, line 31: Did you mean to not include `-u` here?

docker/Makefile
trzecieu: not a particular reason, just convention - I can change that.

PR #539, opened 2020-07-06
Cleanup emsdk_env scripts

sbc100:
Remove the optional argument to contruct_env.  It simple if we
always construct the env in the same place.  Avoid using temp files,
and avoid changing directory.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2489,6 +2489,9 @@ def set_active_tools(tools_to_activate, permanently_activate):
   copy_pregenerated_cache(tools_to_activate)
 
   # Construct a .bat script that will be invoked to set env. vars and PATH
+  # We only do this on windows since emsdk.bat is able to modify the
+  # calling shell environment.  On other platform `source emsdk_env.sh` is
+  # required.
   if WINDOWS:
     env_string = construct_env(tools_to_activate)
     open(EMSDK_SET_ENV, 'w').write(env_string)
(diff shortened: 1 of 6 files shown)


sbc100:
No tests for this stuff no :(

Of course it would be nice if we had some... but I'm not signing up to do that today :)

roddyaj:
Today when I started my environment, I got this error when sourcing `emsdk/emsdk_env.sh`: `emsdk/emsdk_set_env.sh: No such file or directory`  Maybe a missed file in the commit?  Anyway I went back a commit to fix it.

zeux:
I have the same problem; this happens when emsdk_env.sh is sourced from a folder that's not emsdk (e.g. `source ../emsdk/emsdk_env.sh`) because emsdk_set_env.sh is written to the current folder instead of the folder that emsdk is in.

kripken:
Sorry about this, I think it's fixed by https://github.com/emscripten-core/emsdk/pull/541 which I am landing right now.

vargaz:
The optional argument was added to fix this:
https://github.com/emscripten-core/emscripten/issues/9090
So reverting those changes reintroduced the problem.

sbc100:
Oops, sorry about that.   I'll take a look.

sbc100:
At I see the problem.   I will see if we can swich to using stdout instead (a la `ssh-agent`).

PR #546, opened 2020-07-11
Fix set environment script name for powershell

Walt280:
Change EMSDK_SET_ENV to output correct filename for powershell.
Rename write_set_env_bat to write_set_env_script.


Walt280:
According to [this StackOverflow](https://stackoverflow.com/questions/62329296/is-it-possible-to-dot-source-a-string-variable-in-powershell), it doesn't look like "sourcing" strings works that well in Powershell.

PR #550, opened 2020-07-14
Remove emsdk_set_env.bat from .gitignore

sbc100:
This file should only ever exist momentarily.  Its written during
`emsdk construct_env` on windows, but emsdk.bat immediately uses
at deletes it before returning.

Having this file remain around on disk would be a bug so we shouldn't
ignore its presense.


sbc100:
Yeah.. well do generate it but only to be immediately consumed.     If it sticks around thats a bug.

PR #552, opened 2020-07-14
Remove bash-ism from emsdk_env.sh

sbc100:
It is not possible to determine the location of as script that is currently
being sources under a POSIX shell such as dash (the default ubuntu shell).
Only bash has this special BASH_SCRIPT variable.

Needed for #554


sbc100:
Landing TBR to unblock #554

tlively:
LGTM

PR #554, opened 2020-07-14
Fix docker build to not depend on emsdk_set_env.sh

sbc100:
This was an internal script that we used to generate up
until #544.   Now we avoid writing that file at all and
emsdk_env.sh is instead sourced directly.

This should fix the docker CI issues we have been having.


sbc100:
Landing TBR to green the emsdk CI.

PR #556, opened 2020-07-17
1.39.20

kripken:


kripken:
@sbc100 do those failures on CI look familiar perhaps?

kripken:
Testing locally, as the CI error says, we don't have precompiled libraries any more.

They are present in the downloaded `.tbz2` file, and unpacked properly during the `install` step. Then the `activate` step wipes them... :sob:

sbc100:
Strange.  Can you repro locally?  

I'll see if I can figure it out.

kripken:
I think what's going on is
```python
  # Generating .emscripten will cause emcc to clear the cache on first run (emcc
  # sees that the file has changed, since we write it here in the emsdk, and it
  # never saw it before; so it clears the cache as it assumes a new config file
  # means system libraries may need rebuilding). To avoid emcc's clearing wiping
  # out the pregenerated cache contents we want to copy in, run emcc here, then
  # copy the cache contents.
  run_emcc(tools_to_activate)
```

clears the cache.

Did we change where the cache is? We don't seem to copy in the cache anymore after that step, instead the cache was present before that quote.

kripken:
(Yes, it repros locally, that last comment was from looking at what happens)

sbc100:
We don't copy the cache anymore because it already exists in the correct place in the tarball.. so just exacting it is enough.

I can't repro it consistently yet.. working on it.

kripken:
It reproduces consistently for me with
```
rm -Rf upstream
./emsdk install latest
ls -al upstream/emscripten/cache/wasm/ # cache is there
./emsdk activate latest
ls -al upstream/emscripten/cache/wasm/ # cache is gone
```

Removing that `run_emcc(tools_to_activate)` code avoids the problem. But then the problem happens the first time emcc is run.

I guess we need to tell emcc not to clear the cache somehow? I think you recently changed how we decide to clear the cache (where those files are located) - maybe that's related?

sbc100:
On master or just with this patch?

PR #558, opened 2020-07-17
Cleanup legacy 'copy_pregenerated_cache' handling. NFC.

sbc100:


sbc100:
I noticed the run_emcc thing and its comment were linked to the legacy cache copying stuff.

sbc100:
ping.

sbc100:
The CI does install `1.38.31` and `1.38.33`, but maybe we don't tests the cache pregeneration stuff with those old version.  I'll verify manually.

Having said that, if pre-copying the cache breaks for older version it still wouldn't actually stop anyone from working so ... at some point we should just remove this code maybe?

kripken:
Good point. Yeah, seems fine to remove it eventually, at worst making very old versions slower isn't that bad.

sbc100:
Tests with `1.39.10` and it still does the right thing,.

PR #561, opened 2020-07-20
Use pre-built python3 binary on OSX

sbc100:


sbc100:
I think I finally figured this out!

scripts/update_python.py
```diff
@@ -61,23 +67,56 @@ def make_python_patch(arch):
...
-def main():
-    for arch in ('amd64', 'win32'):
-        make_python_patch(arch)
+def build_python():
+    if sys.platform.startswith('darwin'):
+        osname = 'macos'
+        # Take some rather drastic steps to link openssl statically
+        check_call(['brew', 'install', 'openssl', 'pkg-config'])
+        os.remove('/usr/local/opt/openssl/lib/libssl.dylib')
+        os.remove('/usr/local/opt/openssl/lib/libcrypto.dylib')
```
kripken, line 89: do these require root?

sbc100: No, apparently on macOS normaly users have control of /usr/local.

See https://docs.brew.sh/FAQ#why-does-homebrew-prefer-i-install-to-usrlocal

kripken: Interesting, I did not know that.

PR #568, opened 2020-07-23
Fix sanity file location within Dockerfile

### LINKED ISSUE #567: 1.39.20 is not published on Docker
It looks like PR https://github.com/emscripten-core/emscripten/pull/11606 moved the `.emscripten_sanity` file to `upstream/emscripten/cache/sanity.txt`, causing a failure in the `publish-docker-image` pipeline, see:
https://app.circleci.com/pipelines/github/emscripten-core/emsdk/560/workflows/a0c9383c-5520-4acf-905c-7e4a35e1515c/jobs/2172

As a result, the 1.39.20 tag has not been published on Docker.
https://hub.docker.com/r/emscripten/emsdk/tags

Perhaps the `cat` here:
https://github.com/emscripten-core/emsdk/blob/9b405d313a22784bae731e7bd33af867d331532a/docker/Dockerfile#L60

Should be changed to `cat ${EMSDK}/upstream/emscripten/cache/sanity.txt` instead? If you want, I could open a PR.

kleisauke:
Starting from 1.39.20 the `.emscripten_sanity` file has been moved to `upstream/emscripten/cache/sanity.txt`. It also seems that this file is now generated during the `./emsdk activate <VERSION>` command, so it is no longer necessary to call `emcc`.

Resolves #567.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,6 @@
 FROM debian:buster AS stage_build
 
-ARG EMSCRIPTEN_VERSION=1.39.18
+ARG EMSCRIPTEN_VERSION=1.39.20
 ENV EMSDK /emsdk
 
 # ------------------------------------------------------------------------------
@@ -33,6 +33,7 @@ RUN cd ${EMSDK} \
     &&  echo "## Generate standard configuration" \
     &&  ./emsdk activate ${EMSCRIPTEN_VERSION} \
     &&  chmod -R 777 ${EMSDK}/upstream/emscripten/cache \
+    &&  cat ${EMSDK}/upstream/emscripten/cache/sanity.txt \
 &&  echo "## Done"
 
 # Clean up emscripten installation and strip some symbols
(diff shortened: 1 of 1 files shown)


sbc100:
Great!  As long as the sanity is actually generated at runtime and contains the correct paths this is perfect.

sbc100:
Any idea why the CI wasn't triggered for this PR ?

sbc100:
Can you try pushing this PR again?  Or creating a new one?  I found a settings in circleCI that should enable CI for forked PRs.

kleisauke:
Ah, I was wondering why the CI wasn't triggered. I just did:
```bash
LC_ALL=C GIT_COMMITTER_DATE="$(date)" git commit --amend --no-edit --date "$(date)"
git push origin issue-567 -f
```

And it seems to run the CI checks now, thanks!

docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM debian:buster AS stage_build
 
-ARG EMSCRIPTEN_VERSION=1.39.18
+ARG EMSCRIPTEN_VERSION=1.39.20
```
kleisauke, line 3: Perhaps we could change this default value to `tot` (tip-of-tree build) to catch these issues earlier?

Another idea would be to pass on this value here:
https://github.com/emscripten-core/emsdk/blob/56f5a2bcc55576a29f17e9d14ed25ce82a341b6c/.circleci/config.yml#L100
```diff
-          command: docker build --network host ./docker
+          command: docker build --network host  --build-arg EMSCRIPTEN_VERSION=tot ./docker
```
(untested)

kleisauke: The only disadvantage of the above approach is that it is necessary to execute the `./emsdk update tags` command, which increases the Docker image size by 6.8 MB. This won't be a problem if we only do this in CI, I guess.

sbc100: Hmm .. that is sad .. we should probably fix `update tags` to just grab the latest tags from master rather than what I imagine it is doing which a full git clone of emscripten-releases.

sbc100:
I'm happy to land this change as is, if you are ok with it?

kleisauke:
Feel free to land this PR. I've tested it on a quite large project without any problems.

docker/Dockerfile
kleisauke: I've opened https://github.com/emscripten-core/emsdk/pull/571 for discussion. I noticed there's already a `--shallow` parameter for the `install` command, perhaps that should also be supported for the `update-tags` command?

sbc100: You can also just run `rm -rf releases` after `update-tags` for now if you like.

PR #569, opened 2020-07-23
Cleanup Dockerfile. NFC.

kleisauke:
While I was doing #568, I noticed that the Dockerfile had some inconsistent spaces and indentations. This PR should resolve these inconsistencies, and also slightly improves the Dockerfile by:
- Removing unnecessary dependencies from the build stage;
- Moving the `binutils` installation to the first layer;
- Merging the `ENV` declarations into a single line.

To be merged after: #568.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -60,7 +60,6 @@ COPY --from=stage_build /emsdk /emsdk
 # This will let use tools offered by this image inside other Docker images
 # (sub-stages) or with custom / no entrypoint
 ENV EMSDK=/emsdk \
-    EM_CACHE=/emsdk/upstream/emscripten/cache \
     EM_CONFIG=/emsdk/.emscripten \
     EMSDK_NODE=/emsdk/node/12.18.1_64bit/bin/node \
     PATH="/emsdk:/emsdk/upstream/emscripten:/emsdk/upstream/bin:/emsdk/node/12.18.1_64bit/bin:${PATH}"
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,40 +6,37 @@ ENV EMSDK /emsdk
 # ------------------------------------------------------------------------------
 
 RUN echo "## Start building" \
-    \
-&&	echo "## Update and install packages" \
-    &&	apt-get -qq -y update \
-    &&  apt-get -qq install -y --no-install-recommends \
-            libxml2 \
-            wget \
-            git-core \
-            ca-certificates \
-            build-essential \
-            file \
-            python3 python3-pip \
-&&  echo "## Done"
+    && echo "## Update and install packages" \
+    && apt-get -qq -y update \
+    && apt-get -qq install -y --no-install-recommends \
+        binutils \
+        build-essential \
+        ca-certificates \
+        file \
+        git \
+        python3 \
+        python3-pip \
+    && echo "## Done"
 
-RUN  echo "## Get EMSDK" \
-    &&  git clone https://github.com/emscripten-core/emsdk.git ${EMSDK} \
-&&  echo "## Done"
+RUN echo "## Get EMSDK" \
+    && git clone https://github.com/emscripten-core/emsdk.git ${EMSDK} \
+    && echo "## Done"
 
-RUN  echo "## Install Emscripten" \
-    &&  cd ${EMSDK} \
-    &&  ./emsdk install ${EMSCRIPTEN_VERSION} \
-&&  echo "## Done"
+RUN echo "## Install Emscripten" \
+    && cd ${EMSDK} \
+    && ./emsdk install ${EMSCRIPTEN_VERSION} \
+    && echo "## Done"
 
 # This generates configuration that contains all valid paths according to installed SDK
 RUN cd ${EMSDK} \
-    &&  echo "## Generate standard configuration" \
-    &&  ./emsdk activate ${EMSCRIPTEN_VERSION} \
-    &&  chmod -R 777 ${EMSDK}/upstream/emscripten/cache \
-    &&  cat ${EMSDK}/upstream/emscripten/cache/sanity.txt \
-&&  echo "## Done"
+    && echo "## Generate standard configuration" \
+    && ./emsdk activate ${EMSCRIPTEN_VERSION} \
+    && chmod -R 777 ${EMSDK}/upstream/emscripten/cache \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -1,44 +1,42 @@
...
-&&  echo "## Done"
-
-# Clean up emscripten installation and strip some symbols
-RUN echo "## Aggresive optimization: Remove debug symbols" \
-&&  apt-get -qq -y update && apt-get -qq install -y --no-install-recommends \
-        binutils \
+    && echo "## Generate standard configuration" \
+    && ./emsdk activate ${EMSCRIPTEN_VERSION} \
+    && chmod -R 777 ${EMSDK}/upstream/emscripten/cache \
+    && cat ${EMSDK}/upstream/emscripten/cache/sanity.txt \
```
sbc100, line 35: I think it was a bug that we ever shipped a copy of the sanity file.   This was fixed recently: https://github.com/WebAssembly/waterfall/pull/656.

In the past (a few weeks ago) the sanity file was not part of the cache and so didn't get shipped.   Now that it is we accidentally started shipping it. 

This means the cache cleared on first use (because the paths in the sanity file are don't match the installation).

For now the best thing to do is remove this sanity file and instead have it generated by the block below. 

Its not ideal but its the the logic works today:

1. Bad sanity file = clear cache
2. No sanity file = leave catch alone and generate sanity file.

docker/Dockerfile
```diff
@@ -71,10 +59,11 @@ COPY --from=stage_build /emsdk /emsdk
...
 # (sub-stages) or with custom / no entrypoint
-ENV EMSDK_NODE=/emsdk/node/12.18.1_64bit/bin/node
-ENV EMSDK=/emsdk
-ENV EM_CONFIG=/emsdk/.emscripten
-ENV PATH="${EMSDK}:${EMSDK}/upstream/emscripten:${EMSDK}/upstream/bin:emsdk/node/12.18.1_64bit/bin:${PATH}"
+ENV EMSDK=/emsdk \
+    EM_CACHE=/emsdk/upstream/emscripten/cache \
+    EM_CONFIG=/emsdk/.emscripten \
+    EMSDK_NODE=/emsdk/node/12.18.1_64bit/bin/node \
+    PATH="/emsdk:/emsdk/upstream/emscripten:/emsdk/upstream/bin:/emsdk/node/12.18.1_64bit/bin:${PATH}"
```
sbc100, line 66: I assume this is better from a docker perspective because it generated fewer intermediate images?    Otherwise I think its less readable that the existing on-line-per-variable form.

docker/Dockerfile
```diff
@@ -114,25 +102,24 @@ RUN echo "## Update and install packages" \
...
+    && echo "## Done"
 
 # ------------------------------------------------------------------------------
 # Internal test suite of tools that this image provides
 COPY test_dockerimage.sh /emsdk/
 
 RUN echo "## Internal Testing of image" \
-    &&  /emsdk/test_dockerimage.sh  \
-&&  echo "## Done"
+    && ./emsdk/test_dockerimage.sh \
```
sbc100, line 121: Why add the leading `.` here?  Seems like we for sure want to find emsdk in the root of the filesystem no?

docker/Dockerfile
kleisauke: fwiw, I still see this behavior with the tip-of-tree build, see:

```bash
$ ./emsdk install tot-upstream
...
Done installing SDK 'sdk-releases-upstream-1ba7fc82899ab241d0322d5ab20e27ec50cbbf9b-64bit'.
$ [ -e "upstream/emscripten/cache/sanity.txt" ] && echo "sanity.txt exists" || echo "sanity.txt not found"
sanity.txt not found
$ ./emsdk activate tot-upstream
Setting the following tools as active:
   node-12.18.1-64bit
   releases-upstream-1ba7fc82899ab241d0322d5ab20e27ec50cbbf9b-64bit
...
$ [ -e "upstream/emscripten/cache/sanity.txt" ] && echo "sanity.txt exists" || echo "sanity.txt not found"
sanity.txt exists
```

sbc100: Great.  There was just a short period of time where we were accidentally shipping the sanity file.  Hopefully that is fixed now and we are good to go.

As long as there is no sanity file generation of cache file population happening when the image gets used that is the main thing.

docker/Dockerfile
kleisauke: That's right, I did this to minimize the intermediate layers. I can undo this if you'd prefer it to be more readable.

As an aside,
I also added the missing `EM_CACHE` environment variable here. I hope it's OK.

docker/Dockerfile
kleisauke: Ah, added by mistake. Commit https://github.com/emscripten-core/emsdk/pull/569/commits/21d3e4756a516ce400405a0784f0d31715170469 should fix this.

docker/Dockerfile
sbc100: I don't think its necessary.   emscripten these days defaults to using an embedded/local cache.   The EM_CACHE settings in emsdk for legacy/old versions that don't do this.

kleisauke: Ah, I see. I thought it was still in use since `emsdk_env.sh` script sets this environment variable.
```bash
STEP 8: RUN echo "## Aggressive optimization: Remove debug symbols"     && cd ${EMSDK} && . ./emsdk_env.sh     && strip -s `which node`     && rm -fr ${EMSDK}/upstream/emscripten/tests     && rm -fr ${EMSDK}/upstream/fastcomp     && find ${EMSDK}/upstream/bin -type f -exec strip -s {} + || true     && echo "## Done"
## Aggressive optimization: Remove debug symbols
Adding directories to PATH:
PATH += /emsdk
PATH += /emsdk/upstream/emscripten
PATH += /emsdk/node/12.18.1_64bit/bin

EMSDK = /emsdk
EM_CONFIG = /emsdk/.emscripten
EM_CACHE = /emsdk/upstream/emscripten/cache
EMSDK_NODE = /emsdk/node/12.18.1_64bit/bin/node
```

Nevertheless, I will remove it.

sbc100: Yeah.. emsdk has to be conservative because it can install old versions too.. where the cache is not in the new default location.

sbc100:
Is this change ready to land now?

kleisauke:
I just made a minor cosmetic commit, which I overlooked. This PR is now ready to land.

PR #570, opened 2020-07-25
Remove error duplication around missing git executable

sbc100:
Make git a requirement for fetch_emscripten_tags but in the case
when we run it from "emsdk update" we can check for git first.

This cleans up fetch_emscripten_tags and removed the duplication
of the git-missing error messages.

Also, remove check for 'llvm-tags-64bit.txt' which always exists these days
because it is checked-in.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2104,11 +2104,10 @@ def update_emsdk():
     sys.exit(1)
 
   # attempt to run fetch_emscripten_tags if git is available.
-  git = GIT(must_succeed=False)
-  if not git:
+  if GIT(must_succeed=False):
+    fetch_emscripten_tags()
+  else:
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -2118,6 +2102,12 @@ def update_emsdk():
     sys.exit(1)
   if not download_and_unzip(emsdk_zip_download_url, emsdk_path(), download_even_if_exists=True, clobber=False):
     sys.exit(1)
+
+  # attempt to run fetch_emscripten_tags if git is available.
+  git = GIT(must_succeed=False)
+  if not git:
+    print('Update complete, however skipped fetching the Emscripten tags, since git was not found, which is necessary for update-tags.', file=sys.stderr)
+    return
```
kripken, line 2110: instead of depending on this being the end of the function, how about `if git, update tags, else warn on no git`?

kripken (approved):

PR #571, opened 2020-07-25
Docker improvements

kleisauke:
This PR is built on top of #533 with these additional changes:
- Shallow clone emsdk (saves ~2MB).
- Test tip-of-tree build within CI.
- Run the test suite outside the Docker image.
- Avoid copying the `Dockerfile` into the final image (it already exists in `/emsdk/docker`).

Testing the tip-of-tree build within CI hasn't been discussed, hence I've marked this PR as a draft.

/cc @trzecieu, as he's the author of #533.


docker/README.md
```diff
@@ -117,20 +106,20 @@ If your project uses packages that this image doesn't provide you might want to:
...
-    FROM emscripten/emsdk:1.38.25
+    FROM emscripten/emsdk:1.39.17
 
     # Install required tools that are useful for your project i.e. ninja-build
     RUN apt update && apt install -y ninja-build
 
     ```
 2. build it
-    ```shell
+    ```bash
```
sbc100, line 116: Does this make a difference anywhere?

docker/test_dockerimage.sh
```diff
@@ -13,5 +13,15 @@ emcc --version
...
 cmake --version
 
+mkdir -p test
+
+# test emcc compilation
+echo 'int main() { return 0; }' > test/main.c
+emcc -c test/main.c -o test/main
+
+# test embuilder
+embuilder build zlib
```
sbc100, line 23: Should we test this doesn't actually build anything?  (since the re-loaded cache should contain these libraries).

docker/test_dockerimage.sh
```diff
@@ -13,5 +13,15 @@ emcc --version
...
+
 # cleanup after test
+rm -rf test
 find ${EMSDK} -name "*.pyc" -exec rm {} \;
```
sbc100, line 27: Why cleanup here at all..  isn't this run with `--rm` meaning its just gets thrown away anyway?

docker/Dockerfile
```diff
@@ -113,18 +114,6 @@ RUN echo "## Update and install packages" \
     && rm -rf /usr/share/man/??_* \
     && echo "## Done"
 
-# ------------------------------------------------------------------------------
-# Internal test suite of tools that this image provides
-COPY test_dockerimage.sh /emsdk/
-
-RUN echo "## Internal Testing of image" \
-    && /emsdk/test_dockerimage.sh \
-    && echo "## Done"
```
sbc100, line 122: I'm curious, do most docker images contains self-test code like this withing the Dockerfile?

docker/Dockerfile
```diff
@@ -113,18 +114,6 @@ RUN echo "## Update and install packages" \
...
-
-# ------------------------------------------------------------------------------
-# Copy this Dockerimage into image, so that it will be possible to recreate it later
-COPY Dockerfile /emsdk/dockerfiles/emscripten-core/emsdk/
```
sbc100, line 126: Ditto, what is common practice here?

docker/Dockerfile
```diff
@@ -19,11 +19,12 @@ RUN echo "## Start building" \
     && echo "## Done"
 
 RUN echo "## Get EMSDK" \
-    && git clone https://github.com/emscripten-core/emsdk.git ${EMSDK} \
+    && git clone --depth=1 https://github.com/emscripten-core/emsdk.git ${EMSDK} \
```
sbc100, line 22: This whole step seems a little odd given that this Dockerfile now lives in the emsdk repo itself.  I've been working on a PR to remove this step and just have the Dockerfile operate of the current checkout that contains it.  I'm still up for landing this change first as-is though.

docker/README.md
kleisauke: It doesn't make any difference, `bash` is an alias for `shell` within GitHub flavored markdown. I think @trzecieu added this to align it with the other code blocks.

docker/test_dockerimage.sh
kleisauke: I added the `--force` parameter with commit https://github.com/emscripten-core/emsdk/pull/571/commits/d7021674357cf0d3bfa352e7eafc2733c2de32a7 to ensure it's not built from the cache.

docker/test_dockerimage.sh
kleisauke: The `--rm` parameter will not remove files within the mounted volume. I removed these cleanups with commit https://github.com/emscripten-core/emsdk/pull/571/commits/d7021674357cf0d3bfa352e7eafc2733c2de32a7 and compiled the test to `/tmp` instead.

docker/Dockerfile
kleisauke: As far as I know, most Docker images test the image during `docker run` instead of `docker build`. Perhaps this has been added to ensure that all cache directories of Emscripten are pre-populated (since it calls `em++ --version`)?

docker/Dockerfile
kleisauke: I haven't seen it very often, but I think it might be useful for reproducible builds. I removed it because the `Dockerfile` already exists in `/emsdk/docker` (assuming this is the same as the current build `Dockerfile`).

docker/Dockerfile
kleisauke: Great! Copying the current checkout to the Dockerfile is indeed a better way to handle this.

.circleci/config.yml
```diff
@@ -88,16 +88,16 @@ jobs:
             python scripts/test.py
 
   build-docker-image:
-    executor: bionic
+    machine: true # required for testing files permissions
```
sbc100, line 91: This is kind of sad..  I it probably slower to start up and perhaps more costly.

Can you explain a little more why the regular executor doesn't work?

docker/Dockerfile
sbc100: Great lgtm

docker/test_dockerimage.sh
```diff
@@ -13,5 +13,15 @@ emcc --version
...
+# test emcc compilation
+echo 'int main() { return 0; }' | emcc -o /tmp/main.js -xc -
+node /tmp/main.js || exit_code=$?
+if [ $exit_code -ne 0 ]; then
+  echo "Node exited with non-zero exit code: $exit_code"
+  exit $exit_code
+fi
+
+# test embuilder
+embuilder build zlib --force
```
sbc100, line 27: Sorry I under standstand now, the cleanup was being done not inside the docker image but here in the `docker` subdirectory.

In that I case I think the old way was better.   No need to use `/tmp` just write the files here.

.circleci/config.yml
kleisauke: I'm not sure why the regular executor didn't work. This was probably necessary to run privileged Docker containers.

I reverted the executor back to `bionic` with commit https://github.com/emscripten-core/emsdk/pull/571/commits/f5f2fbf730da002083d4c6f7f53e6908fc571dd2.

kleisauke: Argh, it looks like mounting volumes doesn't work for the CircleCI Docker executor, that's why CI failed for https://github.com/emscripten-core/emsdk/commit/f5f2fbf730da002083d4c6f7f53e6908fc571dd2. See:
https://discuss.circleci.com/t/docker-compose-yml-and-volume-mounts/10043/7

Fixed with https://github.com/emscripten-core/emsdk/pull/571/commits/75ca51511f2865852109cad21cef7f846cbb2dda.

docker/test_dockerimage.sh
kleisauke: Since I cannot mount volumes using the CircleCI Docker executor, writing the compiled output to the `docker` subdirectory is not possible. Would you rather use the CircleCI Docker executor (current state) or write to the compiled output to the `docker` subdirectory (and therefore using the `machine` executor)?

sbc100: Sorry, I think its fine to use the /tmp directory inside the container assuming that is what is happening?    Since that directory will be destroyed once the container finished running right?

docker/Makefile
```diff
@@ -13,6 +13,13 @@ endif
 build: .TEST
 	docker build --network host --build-arg=EMSCRIPTEN_VERSION=${version} --tag emscripten/emsdk:${version} .
 
+test: .TEST
+	# test as non-root
+	# FIXME: the `-e HOME=/tmp` is a workaround for:
+	# https://github.com/emscripten-core/emsdk/issues/535
+	docker run --rm -u `id -u`:`id -g` -e HOME=/tmp --net=host emscripten/emsdk:${version} \
+		/bin/bash -c "`cat test_dockerimage.sh`"
```
sbc100, line 21: This construction seems a little bit hacky... why not just leave test_dockerimage.sh inside the container?

docker/test_dockerimage.sh
kleisauke: You're right. The `/tmp` is only used inside the container and then cleaned up afterwards once the container is finished (thanks to the `--rm` parameter).

docker/Makefile
kleisauke: Sorry, I was overthinking the issue. We could indeed just run `/emsdk/docker/test_dockerimage.sh` instead.

Commit https://github.com/emscripten-core/emsdk/pull/571/commits/118f0919659c43ab7c282c95917ccdf8bf952fc8 removes this hackish solution.

docker/Dockerfile
kleisauke: Commit https://github.com/emscripten-core/emsdk/pull/571/commits/118f0919659c43ab7c282c95917ccdf8bf952fc8 operates on the current checkout directory within the Dockerfile. This ensures that any changes in emsdk itself or in the Docker test script (i.e. `docker/test_dockerimage.sh`) are tested during pull requests.

.dockerignore
```diff
@@ -0,0 +1,34 @@
...
+*.fish
+
+# Ignore tcsh scripts
+*.csh
+
+# Ignore linter configurations
+.flake8
+
+# Makefile is not needed
+Makefile
```
sbc100, line 34: Is there some say to inherit everything from .gitignore?      if not perhaps just include it with a comment.

docker/Dockerfile
```diff
@@ -18,12 +18,12 @@ RUN echo "## Start building" \
         python3-pip \
     && echo "## Done"
 
-RUN echo "## Get EMSDK" \
-    && git clone https://github.com/emscripten-core/emsdk.git ${EMSDK} \
-    && echo "## Done"
+# Copy the contents of this repository to the container
+COPY . ${EMSDK}
```
sbc100, line 22: I my version of this change I used `ADD . /emsdk`: https://github.com/emscripten-core/emsdk/compare/docker_use_local_emsdk?expand=1#diff-ebacf6f6ae4ee68078bb16454b23247dR26

I guess they are the same .. i can't remember why I chose one over the other.

Makefile
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env make
```
sbc100, line 1: Perhaps comment at the top saying what the purpose of this makefile is?  i.e. to build the docker image.

If you could leave the Makefile where it was and rung it with `make -f` to avoid polluting the top level?

(You can also remove the #! line from the top, I don't think anyone expects to run the Makefile like that).

sbc100:
Is this ready t do now?   Its still marked as a draft.

.dockerignore
kleisauke: Perhaps we could symlink it with `.gitignore`? Anyways, I cleaned up the file a bit with commit https://github.com/emscripten-core/emsdk/pull/571/commits/4b64400017b5d1c49a35b3f330529a1fb797829e.

docker/Dockerfile
kleisauke: It doesn't make much difference: https://stackoverflow.com/a/24958548/10952119. I think `COPY` is fine here.

Makefile
kleisauke: All done with commit https://github.com/emscripten-core/emsdk/pull/571/commits/531f6c8996ac8e1411bc77b6516845847acb5cfa.

kleisauke:
Yes, this is now ready for final review. I've removed the draft status.

Since this PR contains some outdated details, would you like me to squash all the commits and update the commit message?

sbc100:
Yes please.  We always squash all out PRs on commit anyway by policy. 

IMHO, if something it worth its own commit its probably also work its own PR which means I basically never look at individual commits in PRs.  

This also means that each commit in git also goes through CI on its own which makes bisecting easier.

sbc100:
I think you might also want rebase/integrate #578

sbc100:
I was going to suggest that we split out the "use emsdk inplace" change since it a fairly major one, and commit the other misc improvements separately.   But I don't feel strongly about it.

docker/Dockerfile
trzecieu: That was the goal also. Again it comes from early development of image, where I wanted to reproduce 1:1 images that were published to the web. I don't think it's needed currently.

docker/test_dockerimage.sh
```diff
@@ -1,7 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -ex
 
-sudo -u nobody `which emcc` --version
+if [ $EUID -eq 0 ]; then
```
kleisauke, line 4: Note that this check is needed for Podman (which is the default container engine on Fedora 32). Podman provides a Docker-compatible command line front end that can simply alias the Docker cli (`alias docker=podman`).

Without this check, it will fail to perform the tests (probably because Podman is running in rootless mode).
<details>
  <summary>Details</summary>
  
```bash
$ make -C ./docker version=tot test
make: Entering directory '/home/kleisauke/emsdk/docker'
# test as non-root
docker run --rm -u `id -u`:`id -g` -w /emsdk/docker --net=host emscripten/emsdk:tot \
	bash test_dockerimage.sh
++ which emcc
+ sudo -u nobody /emsdk/upstream/emscripten/emcc --version
sudo: unable to resolve host pc-kaw: Name or service not known

We trust you have received the usual lecture from the local System
Administrator. It usually boils down to these three things:

    #1) Respect the privacy of others.
    #2) Think before you type.
    #3) With great power comes great responsibility.

sudo: no tty present and no askpass program specified
make: *** [Makefile:20: test] Error 1
make: Leaving directory '/home/kleisauke/emsdk/docker'
```
</details>

kleisauke:
Rebased and squashed as requested. I've also polished the commit message and added @trzeci as co-author (thanks!).

If you want, I could make a separate PR for the "emsdk inplace" change. Otherwise, I think this PR is ready to land.

kleisauke:
@trzecieu I've used `trzeci <kontakt@trzeci.eu>` as co-author, as this seems your primary commit email and username. Let me know if something needs to be changed.

PR #578, opened 2020-07-31
Make emscripten directory world writable

sbc100:
Without this emscripten will default to using $HOME for the
emscripten cache rather than the embedded one.


sbc100:
@kleisauke

PR #581, opened 2020-08-05
Disable fastcomp installation for sdk versions 2.0.0 and above

sbc100:
See https://github.com/emscripten-core/emscripten/issues/11319

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2672,6 +2672,9 @@ def expand_sdk_name(name):
           backend = 'upstream'
         else:
           backend = 'fastcomp'
+      if backend == 'fastcomp' and version_key(fullname) >= (2, 0, 0):
+        print('Note: fastcomp is not longer available in version 2.0.0 and above')
+        return name
       return 'sdk-releases-%s-%s-64bit' % (backend, release_hash)
   return name
 
--- a/emsdk.py
+++ b/emsdk.py
@@ -2656,18 +2656,19 @@ def expand_sdk_name(name):
     #   sdk-x.y.z[-(upstream|fastcomp_])-64bit
     # TODO: support short notation for old builds too?
     backend = None
-    if '-upstream' in name:
-      name = name.replace('-upstream', '')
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -2672,6 +2672,9 @@ def expand_sdk_name(name):
           backend = 'upstream'
         else:
           backend = 'fastcomp'
+      if backend == 'fastcomp' and version_key(fullname) >= (2, 0, 0):
+        print('Note: fastcomp is not longer available in version 2.0.0 and above')
+        return name
```
kripken, line 2677: it might be safer to error here? both code-wise it would be clearer that everything halts here. and the error for users would be better I think.

emsdk.py
```diff
@@ -2672,6 +2672,9 @@ def expand_sdk_name(name):
           backend = 'upstream'
         else:
           backend = 'fastcomp'
+      if backend == 'fastcomp' and version_key(fullname) >= (2, 0, 0):
+        print('Note: fastcomp is not longer available in version 2.0.0 and above')
```
kripken, line 2676:
```suggestion
        print('Note: fastcomp is not longer available in version 2.0.0 and above. please use either the upstream backend, or an earlier fastcomp version')
```

emsdk.py
sbc100: The error this produces seems pretty good to me:

```
$ ./emsdk install 2.0.0-fastcomp
Note: fastcomp is not longer available in version 2.0.0 and above.
Error: No tool or SDK found by name '2.0.0-fastcomp'.
```

Basically we are failing expand the input given by the user and then the calling code displays the same error the would see if they had made any other kind of typo or something.

kripken: I don't love the fact that it depends on `2.0.0-fastcomp` not existing and being errored on elsewhere, when we know this should fail right here. But that's a reasonable error.

sbc100:
Closing in favor of #584

PR #586, opened 2020-08-08
2.0.0!

sbc100:


kripken:
Let's wait for monday perhaps? (there will probably be a newer hash by then to tag)

Meanwhile looks like tests need updating.

emscripten-releases-tags.txt
```diff
@@ -1,6 +1,7 @@
 {
-  "latest": "1.40.1",
+  "latest": "2.0.0",
   "releases": {
+    "2.0.0":  "b3458c3346012b47ede6174592074aa62a4e3a7d",
```
kripken, line 4:
```suggestion
    "2.0.0":   "b3458c3346012b47ede6174592074aa62a4e3a7d",
```

sbc100:
Yeah I wasn't proposing to actually land it the weekend.  Just getting excited for 2.0.0!

sbc100:
I think this should be good to land.  I see no particular reason to update the SHA.  Or am I missing something?

kripken:
The hash should be the very latest emscripten landing hash, at minimum, since we update emscripten after doing this, including updating emscripten-releases.txt. If the hash is old, then there would be an odd discrepancy between the two.

sbc100:
Hash updated to https://chromium.googlesource.com/emscripten-releases/+/5974288502aab433d45f53511e961aaca4079d86

emsdk.py
```diff
@@ -2637,11 +2637,14 @@ def error_on_missing_tool(name):
 
 
 def expand_sdk_name(name):
+  if name in ('latest-fastcomp', 'latest-releases-fastcomp'):
+    # Since we no longer support fastcomp in ToT emscripten that the latest
```
kripken, line 2641:
```suggestion
    # Since we no longer support fastcomp in ToT emscripten, the latest
```

emsdk.py
```diff
@@ -2637,11 +2637,14 @@ def error_on_missing_tool(name):
 
 
 def expand_sdk_name(name):
+  if name in ('latest-fastcomp', 'latest-releases-fastcomp'):
+    # Since we no longer support fastcomp in ToT emscripten, the latest
+    # fastcomp release is fixed at 1.40.1.
+    name = 'sdk-fastcomp-1.40.1'
```
kripken, line 2643: I'm actually not sure this is what we want.

This way people silently get `1.40.1` for `latest-fastcomp` forever, instead of an error suggesting they get upstream.

Not urgent in any way, but let's think about this.

sbc100: I was thinking that is the latest fastcomp release... but maybe its better to fail if people as for it that way?

PR #590, opened 2020-08-11
Error on trying to get latest-fastcomp

kripken:
I think this is less surprising than continuing to support `latest-fastcomp`
with the last fastcomp release. That's technically correct - the last release
is 1.40.1, and always will be - but when a user asks for "latest" I think they
want something up to date. Instead, give a clear error that indicates how
they can get an actually up to date build, by using upstream.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2636,12 +2636,13 @@ def error_on_missing_tool(name):
   return 1
 
 
+def exit_with_fastcomp_error(details):
+    exit_with_error(details + ' Please use the upstream llvm backend or use an older version than 2.0.0 (such as 1.40.1).')
+
+
 def expand_sdk_name(name):
   if name in ('latest-fastcomp', 'latest-releases-fastcomp'):
-    # Since we no longer support fastcomp in ToT emscripten, the latest
-    # fastcomp release is fixed at 1.40.1.
-    name = 'sdk-fastcomp-1.40.1'
-
+    exit_with_fastcomp_error('The fastcomp backend is not getting new builds.')
   if name in ('latest', 'sdk-latest', 'latest-64bit', 'sdk-latest-64bit'):
     # This is effectly the default SDK
     return str(find_latest_releases_sdk('upstream'))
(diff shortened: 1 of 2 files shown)


emsdk.py
```diff
@@ -2652,8 +2653,7 @@ def expand_sdk_name(name):
   elif name == 'tot-upstream':
     return str(find_tot_sdk('upstream'))
   elif name in ('tot-fastcomp', 'sdk-nightly-latest'):
-    print('error: The fastcomp compiler is not available tot/nightly builds. Please use the upstream llvm backend or use an older version than 2.0.0.')
-    sys.exit(1)
+    exit_with_fastcomp_error('The fastcomp backend is not getting tot/nightly builds.')
```
sbc100, line 2656: Just move these two up to the top and combine with the others?

kripken: I don't want to combine them as the error text is a little different. I can move it up though, but then it would separate the `tot-*` ones from each other. What do you prefer?

sbc100: But why not just make the error slightly more generic:  "fastcomp is no longer available in recent releases."

If you add `The available sdk with fastcomp support is `fastcomp-1.40.1` I think that would be enough to make just one single error message that would work in all tree cases.. but i'm ok if you doing do that part.

kripken: Sounds fine, I guess the error message doesn't need that much detail. I agree it's simpler code that way, done.

kripken:
The test here fails on windows. We exit the python with `sys.exit(1)` but apparently that doesn't make the `emsdk.bat` invocation fail. Do windows bat files have a concept of an exit code at all...?

sbc100:
> The test here fails on windows. We exit the python with `sys.exit(1)` but apparently that doesn't make the `emsdk.bat` invocation fail. Do windows bat files have a concept of an exit code at all...?

Yeah, the have the ERRORLEVEL thing and the `call` concept which effects whether or not a child exiting means you exit too.   I think there may be bug in emsdk.bat then.   Can you skip that test on windows and open a bug?

kripken:
Updated to skip on windows and link to an issue.

PR #594, opened 2020-08-16
Fix up emsdk_env.sh so it works in dash/zsh

rwtolbert:
This script says to avoid bashisms, but then basically
fails for folks not using bash (which is a lot of
people on macOS now)

this addition tries a bit harder to find the actual DIR
of itself and fails if that doesn't work.

This version works on zsh and dash (on Ubuntu)

Does still fail on ksh, but I hope this is progress.

### DIFF
--- a/emsdk_env.sh
+++ b/emsdk_env.sh
@@ -15,10 +15,25 @@
 #
 # which won't have any effect.
 if [ -z "$BASH_SOURCE" ]; then
-  if [ ! -f "./emsdk.py" ]; then
+  D1=`dirname $0`
+  DIR=`cd $D1 && pwd`
+  S=`basename $0`
+
+  READLINK=`which readlink`
+  if [ -x "$READLINK" ] ; then
+    DLINK=`readlink $D1/$S`
+    if [ "$DLINK" != "" ] ; then
+      D1=`dirname $DLINK`
+      DIR=`cd $D1 && pwd`
+    fi
+  fi
+  if [ ! -f "$DIR/emsdk.py" ]; then
     echo "error: You must be in the same directory as emsdk_env.sh when sourcing it (or switch to the bash shell)" 1>&2
   fi
-  DIR="."
+  unset S
+  unset D1
+  unset DLINK
+  unset READLINK
 else
   DIR="$(dirname "$BASH_SOURCE")"
 fi


emsdk_env.sh
```diff
@@ -15,10 +15,25 @@
...
+  READLINK=`which readlink`
+  if [ -x "$READLINK" ] ; then
+    DLINK=`readlink $D1/$S`
+    if [ "$DLINK" != "" ] ; then
+      D1=`dirname $DLINK`
+      DIR=`cd $D1 && pwd`
+    fi
+  fi
+  if [ ! -f "$DIR/emsdk.py" ]; then
     echo "error: You must be in the same directory as emsdk_env.sh when sourcing it (or switch to the bash shell)" 1>&2
```
sbc100, line 31: This error message could use some cleanup now perhaps.

How about: `error: unable to determine emsdk directory.  Perhaps you are using a shell or environment that this script does not support.  One possible solution is to source this script while in the emsdk directory.`

sbc100:
I wonder if we should document at the top exactly which shells this script is known to work with...  also, I guess we should really be testing under all of them too (doesn't need to be part of this PR).

emsdk_env.sh
rwtolbert: Good idea on the error message. I kind of left it as is, but am happy to modify.

rwtolbert:
I can clean up the comments at the top too. The fact that it said "avoid bashisms" and then used BASH_SOURCE confused the heck out of me at first.

The one shell it currently fails on is old ksh on macOS (no idea how old it is)

and we don't have to test on csh or fish as they have their own script.

sbc100:
> I can clean up the comments at the top too. The fact that it said "avoid bashisms" and then used BASH_SOURCE confused the heck out of me at first.
> 

That is why it has the fallback for when BASH_SOURCE is not set.  If it was a bash-only it would just assume `BASH_SOURCE` is always set.

I think why I made that change I looked at how gross the alternatives where an balked and just went with "you have source this from the emsdk dir if you are not on bash" and called it good.  But as you point out more and more folks are on non-bash these days so I agree that its probably worth the effort.

As you can see trying to fine the name of a script being sourced in other shells is somewhat painful.. :-/

> The one shell it currently fails on is old ksh on macOS (no idea how old it is)
> 
> and we don't have to test on csh or fish as they have their own script.

rwtolbert:
Just thought I'd add: If this is too gnarly, I have a `emsdk_env.zsh` that just works using some zsh-specific feature. 

We could leave the `emsdk_env.sh` version as-is and just have one more special version like there is for fish.

sbc100:
> Just thought I'd add: If this is too gnarly, I have a `emsdk_env.zsh` that just works using some zsh-specific feature.
> 
> We could leave the `emsdk_env.sh` version as-is and just have one more special version like there is for fish.

I think that might be better approach, especially since we seem to be doing that for all the other shells.  My only concern would be for other non-bash borne shell users (which I guess means mostly dash?).  Are they still stuck with the gnarly version?  Or maybe we just revert to the simple "you have to be in the current directory" version?

Anyway it seems like `zsh` is a common enough shell these days (since now apple made the switch) that it deserves its own special handling.  Is there some way we can delegate to `emsdk_env.zsh` from `emsdk_end.sh` if we detect zsh?  Or can we just embed the zsh logic inside of `emsdk_end.sh`.

rwtolbert:
Sorry this issue includes 'dash' in the title. I'm pretty sure I can't get it work. But this lastest change does work for bash, zsh and ksh. 

I can close and resubmit a cleaner version if need be.

emsdk_env.sh
```diff
@@ -1,28 +1,64 @@
...
+fi
+
+if [ ! -f "$DIR/emsdk.py" ]; then
+  echo "Error: unable to determine 'emsdk' directory. Perhaps you are using a shell or" 1>&2
+  echo "       environment that this script does not support." 1>&2
+  echo 1>&2
+  echo "A possible solution is to source this script while in the 'emsdk' directory." 1>&2
+  echo 1>&2
+  unset DIR
+  return
```
sbc100, line 58: What does return do in this position?

emsdk_env.sh
```diff
@@ -1,28 +1,64 @@
...
+  CURRENT_SCRIPT="$BASH_SOURCE"
+elif [ -n "$ZSH_VERSION" ]; then
+  CURRENT_SCRIPT="${(%):-%x}"
+elif [ -n "$KSH_VERSION" ]; then
+  CURRENT_SCRIPT=${.sh.file}
+fi
+
+if [ -n "$CURRENT_SCRIPT" ]; then
+  DIR=$(dirname "$CURRENT_SCRIPT")
+  # now work out actual DIR if this is part of a symlink
```
sbc100, line 38: Can you explain a little more why this is needed?  i.e. what would i need to do in order to setup a situation that would need this code.   (Edit: I just saw the added tests which might help explain).


I find it quite hard to follow exactly what this is doing but I guess is basically an implementation of `realpath`?

emsdk_env.sh
```diff
@@ -1,28 +1,64 @@
...
+  echo 1>&2
+  echo "A possible solution is to source this script while in the 'emsdk' directory." 1>&2
+  echo 1>&2
+  unset DIR
+  return
 fi
 
 # Force emsdk to use bash syntax so that this works in windows + bash too
 eval `EMSDK_BASH=1 $DIR/emsdk construct_env`
+unset CURRENT_SCRIPT
```
sbc100, line 63: Looks like we can unset `CURRENT_SCRIPT` up on line 50?

scripts/test_source_env.sh
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env bash
```
sbc100, line 1: Thanks for writing this!

Can you add this to the linux tests in `.github/actions/`?

emsdk_env.sh
rwtolbert: this basically returns early if the 'eval' below is going to fail.

prevents the error that you get on the last line:

```
$ . ../emsdk_env.sh
Error: unable to determine 'emsdk' directory. Perhaps you are using a shell or
       environment that this script does not support.

A possible solution is to source this script while in the 'emsdk' directory.

dash: 1: ../emsdk_env.sh: /emsdk: not found
```

so instead you just get:

```
$ . ../emsdk_env.sh
Error: unable to determine 'emsdk' directory. Perhaps you are using a shell or
       environment that this script does not support.

A possible solution is to source this script while in the 'emsdk' directory.
```

not that big of a deal, either way. Just seemed cleaner, but then I'm not shell programming expert at all.

emsdk_env.sh
rwtolbert: yeah, this is an amalgamation of a lot of code I've found and collected to help get the real path to a script. I've needed it for work more than once.

It basically tries extra hard to find the real emsdk directory whether the directory itself is symlinked elsewhere or the emsdk_env.sh script itself is a symlink.

That's why I had to write some tests to see if I could verify all the combinations.

emsdk_env.sh
rwtolbert: Nice catch.

scripts/test_source_env.sh
rwtolbert: I'm not sure where that is? I wondered how tests worked. Guess I ran out of steam last night before finding the runner.

Do you mean add this to the mac/linux tests in the .circleci/config.yml?  If so, is there a way to run that locally to see if they work?

emsdk_env.sh
sbc100: Could you add a comment here to that effect.   If its trying to emulate the `realpath` command perhaps say that, and why we can't use use it.

scripts/test_source_env.sh
sbc100: Sorry yes I meant to say `.circleci/config.yml`.  Testing just on linux should be fine.

emsdk_env.sh
rwtolbert: sounds good

Issue #598, opened 2020-08-24
Running emsdk.bat activate latest fails


zimmermandonny:
I followed the Windows installation instructions and the successfully complete the step "emsdk.bat install latest". I get the following error when I run the next step which is "emsdk.bat activate latest".

```
c:\emsdk>emsdk.bat activate latest
Traceback (most recent call last):
  File "c:\emsdk\\emsdk.py", line 13, in <module>
    import multiprocessing
  File "D:\obj\windows-release\37amd64_Release\msi_python\zip_amd64\__init__.py", line 16, in <module>
  File "D:\obj\windows-release\37amd64_Release\msi_python\zip_amd64\context.py", line 6, in <module>
  File "D:\obj\windows-release\37amd64_Release\msi_python\zip_amd64\reduction.py", line 16, in <module>
  File "D:\obj\windows-release\37amd64_Release\msi_python\zip_amd64\socket.py", line 49, in <module>
    import _socket
ImportError: DLL load failed: %1 is not a valid Win32 application.
```

zimmermandonny:
It works if I run the command "python emsdk.py activate latest". I must have a python issue?

sbc100:
Its looks like there is some issue using the bundled version of python that we include in the emsdk.

Do you have any python environment variables set?   Perhaps `PYTHONPATH` or `PYTHONPATH`?   

Do you know what `D:\obj\windows-release\37amd64_Release\msi_python` is?  Is that where you python installation is?

zimmermandonny:
You are correct in that there is an issue with the python bundled with the emsdk.

I tried to run the emsdk_env.bat file and get the same original errors. If I run:

c:\emsdk>"C:\Users\Donny Zimmerman\AppData\Local\Programs\Python\Python38-32\python.exe" emsdk.py construct_env

I get the following result:

Setting environment variables:
EMSDK = C:/emsdk
EM_CONFIG = C:\emsdk\.emscripten
EM_CACHE = C:/emsdk/upstream/emscripten\cache
EMSDK_NODE = C:\emsdk\node\12.18.1_64bit\bin\node.exe
EMSDK_PYTHON = C:\emsdk\python\3.7.4-pywin32_64bit\python.exe
JAVA_HOME = C:\emsdk\java\8.152_64bit

I get the same errors when running emcc -v in the upstream/emscripten folder unless I run it with the 3.8 version of Python.

Here are my env variables that pertain to emscripten or python:

[...]

PR #600, opened 2020-08-25
Clear PYTHONHOME when using bundled python

sbc100:
Fixes: #598

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -6,13 +6,31 @@
 
 # Wrapper script that runs emsdk.py
 
+# First look for pre-built python (macos)
 if [ -z "$EMSDK_PYTHON" ]; then
-  # Look for python3 first.  This is especially important on macOS (See:
-  # https://github.com/emscripten-core/emsdk/pull/273)
-  EMSDK_PYTHON=$(which python3 2> /dev/null)
-  if [ $? != 0 ]; then
-    EMSDK_PYTHON=python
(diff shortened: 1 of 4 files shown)


emsdk.py
```diff
@@ -2575,6 +2575,25 @@ def construct_env(tools_to_activate):
         env_string += 'export ' + key + '="' + value + '";\n'
       else:
         assert False
+      if 'EMSDK_PYTHON' in env_vars:
+        # When using our bundled python we never want the users
```
kripken, line 2579:
```suggestion
        # When using our bundled python we never want the user's
```

emsdk
```diff
@@ -6,13 +6,31 @@
...
+    # PYTHONHOME or PYTHONPATH
+    # https://github.com/emscripten-core/emsdk/issues/598
+    unset PYTHONHOME
+    unset PYTHONPATH
   fi
 fi
 
+# Look for `python3` first.  This is especially important on macOS (See:
+# https://github.com/emscripten-core/emsdk/pull/273)
+if [ -z "$EMSDK_PYTHON" ]; then
```
kripken, line 24: line 10 is identical to line 21, is that intentional?

sbc100: Yes this file is structured as a sequence of attempts to find python.. as soon as one of them passes the other ones are not required.

This matches how we do it in the `emcc` wrapper that calls `emcc.py`

kripken (approved):
Makes sense, but I don't understand the fine details here...

PR #601, opened 2020-08-25
Add stalebot

sbc100:


.github/stale.yml
```diff
@@ -0,0 +1,18 @@
...
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because there has been no
+  activity in the past year.  It will be closed automatically if no further
+  activity occurs in the next 7 days.  Feel free to re-open at any time if this
```
kripken, line 15: should be 30?

sbc100: I just copied on the from emscripten so I guess that wrong too ?  :)

kripken: In that case yes, sounds like that's wrong too, heh...

Issue #587, opened 2020-08-09
.emscripten file can be invalid when nodejs is not installed by emsdk


reosablo:
If nodejs is not installed by `emsdk`, the output ".emscripten" file by `emsdk activate` will look like this.

```python
NODE_JS = ''/usr/bin/nodejs''
# too many quotes
```

`emcc` fails to execute by this syntax error.

I think fixing [this line](https://github.com/emscripten-core/emsdk/blob/1.40.1/emsdk.py#L1450) could be a solution.

```python
activated_config['NODE_JS'] = "'%s'" % node_fallback
# the value will be quoted later again
```

kripken:
Thanks @reosablo , I think you're right! Would you like to open a PR with that?

reosablo:
> Thanks @reosablo , I think you're right! Would you like to open a PR with that?

OK! I'll do that.

PR #602, opened 2020-08-30
Fix `emsdk activate` breaking configuration file

reosablo:
related to issue #587

> If nodejs is not installed by `emsdk`, the output ".emscripten" file by `emsdk activate` will look like this.
> 
> ```python
> NODE_JS = ''/usr/bin/nodejs''
> # too many quotes
> ```
> 
> `emcc` fails to execute by this syntax error.
> 
> I think fixing [this line](https://github.com/emscripten-core/emsdk/blob/1.40.1/emsdk.py#L1450) could be a solution.
> 
> ```python
> activated_config['NODE_JS'] = "'%s'" % node_fallback
> # the value will be quoted later again
> ```

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1452,7 +1452,7 @@ def generate_dot_emscripten(active_tools):
     node_fallback = which('nodejs')
     if not node_fallback:
       node_fallback = 'node'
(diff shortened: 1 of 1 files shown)


sbc100:
Thanks!

PR #603, opened 2020-08-31
add a wasm bazel toolchain

walkingeyerobot:


kripken:
As someone that knows almost nothing about bazel I don't have many comments here :smile: but this is great to see!

What does `update_bazel_workspace.sh` do? I don't see it mentioned in the docs here.

bazel/README.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+Now you can run `bazel build :hello-world-wasm`. The result of this build will
+be the individual files produced by emscripten. Note that some of these files
+may be empty. This is because bazel has no concept of optional outputs for
+rules.
+
+## Updating to the latest version of emscripten
+When a new version of emscripten is released, emsdk maintainers can run
+`emsdk/scripts/update_bazel_workspace.sh` to update the bazel WORKSPACE file.
```
kripken, line 67: Sounds like this is something we'll need to add to our release processes? https://github.com/emscripten-core/emscripten/blob/master/docs/process.md#release-processes

Basically every time we create a PR for a new release, we need to run this script and include the results in the PR?

walkingeyerobot: Ah, yes, this makes much more sense to put there.

I've removed the note in the README here about running the update script, as users should not have to run it. I have a PR ready to go here to update emscripten docs: https://github.com/emscripten-core/emscripten/compare/master...walkingeyerobot:master.

bazel/WORKSPACE
```diff
@@ -0,0 +1,25 @@
...
+)
+
+load("@build_bazel_rules_nodejs//:index.bzl", "npm_install")
+
+# emscripten 2.0.2
+http_archive(
+    name = "emscripten",
+    sha256 = "9febbc252372efee82e62d548f510dfc85ade6661f5f3be28823848d4757d614",
+    strip_prefix = "install",
+    url = "https://storage.googleapis.com/webassembly/emscripten-releases-builds/linux/ede25d889a0abe63360d4c5d420087c8753b8bbe/wasm-binaries.tbz2",
```
sbc100, line 16: Is there some way to extract the SHA here from the URL and make it a variable?   Might make it easier to update and document the update process?

bazel/WORKSPACE
```diff
@@ -0,0 +1,25 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "build_bazel_rules_nodejs",
+    sha256 = "0f2de53628e848c1691e5729b515022f5a77369c76a09fbe55611e12731c90e3",
+    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/2.0.1/rules_nodejs-2.0.1.tar.gz"],
```
sbc100, line 6: Could/should we package up these toolchain files like this too (to avoid copying them into each project)?

bazel/bazelrc
```diff
@@ -0,0 +1,5 @@
+build:wasm --crosstool_top=//emscripten_toolchain:everything
+
+build:wasm --cpu=wasm
+
+build:wasm --host_crosstool_top=@bazel_tools//tools/cpp:toolchain
```
sbc100, line 5: Is this last line needed?  Is it not the default?

bazel/emscripten_toolchain/emar.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+source emscripten_toolchain/env.sh
+
+exec python3 $EMSCRIPTEN/emar.py "$@"
```
sbc100, line 5: I wonder if we can (one day) find a way to drop the extra wrapper scripts and just use the normal `$EMSCRIPTEN/emar` entry points that are part of emscripten.

bazel/emscripten_toolchain/env.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+mkdir -p /tmp/bazel_emscripten/
+
+export ROOT_DIR=`(pwd -P)`
+export EMSCRIPTEN=${ROOT_DIR}/external/emscripten/emscripten
+
+export EMCC_WASM_BACKEND=1
```
sbc100, line 8: This is not needed anymore.

bazel/emscripten_toolchain/env.sh
```diff
@@ -0,0 +1,13 @@
...
+
+export EMCC_SKIP_SANITY_CHECK=1
+
+export EM_CONFIG=${ROOT_DIR}/emscripten_toolchain/emscripten_config
+export EM_CACHE=/tmp/bazel_emscripten/emscripten_cache
```
sbc100, line 13: Ideally you can remove this and have the cache default to  $EMSCRIPTEN/cache (assuming that directory is read/write)?

bazel/emscripten_toolchain/emscripten_config
```diff
@@ -0,0 +1,11 @@
...
+
+ROOT_DIR = os.environ["ROOT_DIR"]
+EMSCRIPTEN_ROOT = os.environ["EMSCRIPTEN"]
+
+LLVM_ROOT = ROOT_DIR + "/external/emscripten/bin"
+EMSCRIPTEN_NATIVE_OPTIMIZER = LLVM_ROOT + "/optimizer"
+NODE_JS = ROOT_DIR + "/external/nodejs_linux_amd64/bin/node"
+BINARYEN_ROOT = ROOT_DIR + "/external/emscripten"
+
+TEMP_DIR="/tmp/bazel_emscripten"
```
sbc100, line 11: I think just letting this default to the system TEMP_DIR is probably fine?

scripts/update_bazel_workspace.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+# This command should be run from the emsdk root.
```
sbc100, line 2: Maybe a little more info about what this does and who should run it and when?

scripts/update_bazel_workspace.sh
```diff
@@ -0,0 +1,65 @@
...
+
+function cleanup {
+  popd > /dev/null
```
sbc100, line 5: I don't think this is needed since that current working directory of the script doesn't effect the calling process.

scripts/update_bazel_workspace.sh
```diff
@@ -0,0 +1,65 @@
...
+}
+
+ERR=0
+# Attempt to change to the emsdk root directory
+pushd "${PWD%%/emsdk*}/emsdk" > /dev/null 2> /dev/null
```
sbc100, line 10: If think you change this to `cd $(dirname $0)/..` it will mean you can run it from anywhere (and also avoid pushd)?

sbc100 (commented):
Nice!  I'm excited to try it.  I've not used bazel yes myself.

bazel/WORKSPACE
walkingeyerobot: This should never need to be edited by hand; I added `scripts/update_bazel_workspace.sh` to handle this when a new version is released.

bazel/WORKSPACE
walkingeyerobot: I believe Bazel intelligently caches this; I don't think this is an issue.

bazel/emscripten_toolchain/emar.sh
walkingeyerobot: I think we'll be able to eventually, but for now it's important to set up the environment via env.sh.

bazel/emscripten_toolchain/env.sh
walkingeyerobot: Done.

bazel/emscripten_toolchain/env.sh
walkingeyerobot: bazel had a lot of trouble for me when I tried to remove this. It kept giving me errors like this:
```
  File "/usr/lib/python3.8/os.py", line 223, in makedirs
    mkdir(name, mode)
OSError: [Errno 30] Read-only file system: '/home/mitchfoley/.emscripten_cache/wasm'
```
I would not be surprised to learn that this is a problem specific to how my machine is set up. I'll do more investigation with another machine and leave it as-is for now.

sbc100: But storing the cache in /tmp seems like a bad idea.. we want the cache to persist.    

How about: `export EM_CONFIG=${ROOT_DIR}/emscripten_toolchain/cache/`?

scripts/update_bazel_workspace.sh
walkingeyerobot: Done.

My bash-fu is weak. Thanks for the tip!

scripts/update_bazel_workspace.sh
walkingeyerobot: Removed.

scripts/update_bazel_workspace.sh
walkingeyerobot: Added a bit more description.

bazel/emscripten_toolchain/emscripten_config
walkingeyerobot: Done.

bazel/emscripten_toolchain/env.sh
walkingeyerobot: That appears to have worked.

I'm with you though, ideally we just leave it at the default. It looks like bazel's sandboxing stuff forbids this though.

bazel/bazelrc
walkingeyerobot: Removed; it doesn't look like it's needed.

bazel/emscripten_toolchain/env.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+export ROOT_DIR=`(pwd -P)`
+export EMSCRIPTEN=${ROOT_DIR}/external/emscripten/emscripten
+
+export EMCC_SKIP_SANITY_CHECK=1
```
sbc100, line 6: Were you not able to drop this line?

sbc100 (approved):
OK seems like a good first pass.

Can/should we add a quick test?

bazel/emscripten_toolchain/env.sh
walkingeyerobot: Just confirmed I am able to drop this line. Removed.

sbc100:
I was thinking just a simple shell script that basically runs through the setup ad builds hello_world.c?

Then you would just add that shell script to `.circleci/config.yml` in the same you see `scripts/test.sh` in there.

I won't object to hard to landing without that but its seems like we should at least check it works.   If you have verified that it works locally the test script could be a followup?

.circleci/config.yml
```diff
@@ -124,6 +124,28 @@ jobs:
...
+      - run: curl -fsSL https://bazel.build/bazel-release.pub.gpg | gpg --dearmor > bazel.gpg
+      - run: mv bazel.gpg /etc/apt/trusted.gpg.d/
+      - run: echo "deb [arch=amd64] https://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
+      - run:
+          name: install pip
+          command: |
+            apt-get update -q
+            apt-get install -q -y python-pip python3-pip
+      - run: python2 -m pip install --upgrade pip
+      - run: python3 -m pip install --upgrade pip
```
sbc100, line 141: Do you need python2 and python3?

.circleci/config.yml
```diff
@@ -124,6 +124,28 @@ jobs:
             docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
             make -C ./docker version=${CIRCLE_TAG} alias=latest push
 
+  test-bazel:
+    executor: bionic
+    steps:
+      - checkout
+      - run: apt install curl gnupg
```
sbc100, line 131: Maybe use `apt-get install -q -y` here for consistency.

.circleci/config.yml
sbc100: Maybe make the `pip install --upgrade pip` command part of the `install pip` job? 
Do you need to upgrade pip here?

sbc100 (approved):
Great!

.circleci/config.yml
walkingeyerobot: Done.

.circleci/config.yml
walkingeyerobot: Fixed.

sbc100 (approved):
lgtm.

Let me know if you are ready and I'll land this.

PR #607, opened 2020-09-09
Parse the releases info only once in emsdk.py.

j-jorge:
`load_releases_info()` from `emsdk.py` reads `emscripten-releases-tags.txt` into memory and parses it as Json. It is called many times to retrieve info about the sdk. For example `emsdk.py list` will call it 12 times.

Since each call produces the same result I think we could save some reading and parsing by caching the result, thus this pull request :)

Best regards

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2185,12 +2185,15 @@ def exit_with_error(msg):
 
 # Load the json info for emscripten-releases.
 def load_releases_info():
-  try:
-    text = open(sdk_path('emscripten-releases-tags.txt'), 'r').read()
-    return json.loads(text)
-  except Exception as e:
-    print('Error parsing emscripten-releases-tags.txt!')
-    exit_with_error(str(e))
+  if not hasattr(load_releases_info, 'cached_info'):
+    try:
+      text = open(sdk_path('emscripten-releases-tags.txt'), 'r').read()
+      load_releases_info.cached_info = json.loads(text)
+    except Exception as e:
+      print('Error parsing emscripten-releases-tags.txt!')
+      exit_with_error(str(e))
+
+  return load_releases_info.cached_info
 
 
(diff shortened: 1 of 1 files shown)


sbc100:
Could we make this change much smaller by having `load_releases_info` just load on first call?   Maybe call it get_release_info() instread?

sbc100:
> @sbc100 If you're thinking about something like a lazy loading of the releases info, I did try with a global variable, with something like:
> 
> ```
> releases_info_cached = None
>   
> # Load the json info for emscripten-releases.
> def load_releases_info():
>   if releases_info_cached:
>     return releases_info_cached
>   
>   try:
>     text = open(sdk_path('emscripten-releases-tags.txt'), 'r').read()
>     releases_info_cached = json.loads(text)
>     return releases_info_cached
>   except Exception as e:
>     print('Error parsing emscripten-releases-tags.txt!')
>     exit_with_error(str(e))
> ```
> 
> Is it what you have in mind? The above code led to the following error: `UnboundLocalError: local variable 'releases_info_cached' referenced before assignment` but there may be another way to do it (my Python-fu is low).
> 
> Also between a global variable and a parameter passed around I personally tend to prefer the second as it makes the dependencies explicit.
> 
> You tell me :)

You would need to add `global releases_info_cached` to make that work in that way.

But maybe you can avoid the global and just cache on the function itself:

```
def load_releases_info():
    if not hasattr(load_releases_info, 'cached_info'):
        .. assign to load_releases_info.cached_info...
    
   return load_releases_info.cached_info
```

I agree there is an argument for passing parameters, over using global cached state, but I also like how localized the change is if you do it this way.

j-jorge:
Thanks @sbc100, it's quite clean this way :)

sbc100 (approved):

PR #609, opened 2020-09-10
Remove copy_pregenerated_cache

sbc100:
Newer SDKs don't require this and removing it doesn't break
any older SDKs (it just slows down their first use).


kripken (approved):

kripken:
(we have tests confirming the prebuilt libc is used, correct?)

sbc100:
We do yes, but not for the older SDKs that required this copying mechanism.

PR #610, opened 2020-09-10
Fix sanity file generation in Dockerfile

sbc100:
We no longer ship the sanify file as part of the SDK but
expect it be generated on first use:
  https://github.com/WebAssembly/waterfall/pull/656

Also remove the Dockerfile itself from the docker image if
we include it then whenever you change anything in the
Dockerfile it invalides all the layers and forces a
completely rebuild.  This makes iterating on the image
very slow.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,9 +4,6 @@
 # Allow to run the test script inside the Docker container
 !/docker/test_dockerimage.sh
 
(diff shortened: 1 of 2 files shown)


sbc100:
@kleisauke

sbc100:
Landing TBR to unblock CI.

PR #613, opened 2020-09-11
update bazel workspace for emscripten 2.0.3

walkingeyerobot:
Looks like 2.0.3 was released before docs were updated describing how to update the bazel toolchain.

Also adds to the bazel test to make sure the WORKSPACE file is updated.


scripts/test_bazel.sh
```diff
@@ -5,5 +5,18 @@ echo "test bazel"
...
+        emscripten-releases-tags.txt \
+      | sed --expression "s/\./\\\./g")
+# Based on the latest version number, get the commit hash for that version.
+HASH=$(grep "${VER}" emscripten-releases-tags.txt \
+      | grep -v latest \
+      | cut -f4 -d\")
+
+# Ensure the WORKSPACE file is up to date with the latest version.
+grep ${VER} bazel/WORKSPACE
+grep ${HASH} bazel/WORKSPACE
```
sbc100, line 19: So this is going to force anyone who updates emscripten-releases-tag.txt to also run the bazel update script?

If so, could we make the failure state that clearly and how to resolve it?

walkingeyerobot: Done.

If there's a better way to make this clear, let me know.

sbc100 (approved):

PR #615, opened 2020-09-15
Add missing python locations to emsdk.ps1

aminya:
This adds the missing python locations to emsdk.ps1 so it matches emsdk.bat. 

On Windows, if you add the emsdk dir to the path and use PowerShell as your terminal (which is almost everyone), then any call to "emsdk" will actually use the ps1 files. This means the Python version that is shipped with esmdk is not used!

### DIFF
--- a/emsdk.ps1
+++ b/emsdk.ps1
@@ -1,6 +1,8 @@
 $ScriptDirectory = Split-Path -parent $PSCommandPath
 
 $PythonLocations = $(
+    "python\3.7.4-pywin32_64bit\python.exe",
+    "python\3.7.4_64bit\python.exe",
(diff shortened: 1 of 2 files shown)


sbc100:
Thanks!

I didn't realize `python-3.7.4-64bit` was still used in a couple of places

Would you mind also changing `python-3.7.4-64bit` to `python-3.7.4-pywin32-64bit` in the two places where it is used?

aminya:
@sbc100 It is done in https://github.com/emscripten-core/emsdk/pull/615/commits/d26db7497946ba701852c9789cf07b62a4c6aee1

sbc100 (approved):

Issue #618, opened 2020-09-16
Please release a new version to include the Bazel toolchain


attilaolah:
Would it be possible for future versioned releases to include the `bazel` subdirectory?

Alternatively, it could be provided as a separate download, e.g. `emsdk-bazel.tar.gz`, so that we can fetch it with Bazel and reference all the goodness inside. (We can still do that via Git but using HTTP is prettier.)

attilaolah:
Oh I just realised that the directory was only added four days ago in #603, but the latest release is five days old.

I guess then, please tag a new release :)

kripken:
2.0.4 has just been tagged!

Good to hear this will be useful @attilaolah :smile:

PR #619, opened 2020-09-16
2.0.4

kripken:
First release with Bazel support! (see request in #618)


sbc100 (approved):
I guess maybe we should have single script that does all of this..?

kripken:
Maybe, yeah. But the harder work is on the emscripten repo. A single script might help there more, for example right now when I tagged I messed up and I added two commits instead of a single one. But it's hard to write a script to automate that as it depends on the local setup across multiple repos, which is why I do it manually.

PR #621, opened 2020-09-18
Set user environment variables permanently by using `--permanent` + deprecate `--global`

### LINKED ISSUE #189: emsdk construct_env on Windows nukes local paths

### LINKED ISSUE #467: Way to set user environment variables

aminya:
Closes https://github.com/emscripten-core/emsdk/pull/468

Fixes #467

This makes ./emsdk activate --global to set user environment variables by default instead of previous system-wide-only activation.

Also removes some magic that implicitly nukes user envvars when system ones are shadowed, as a system admin should understand what's going on when setting envvars. Thus fixes #189 and closes #192.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2480,7 +2480,7 @@ def to_msys_path(p):
 
 # Looks at the current PATH and adds and removes entries so that the PATH reflects
 # the set of given active tools.
-def adjusted_path(tools_to_activate, log_additions=False, system=False):
+def adjusted_path(tools_to_activate, system=False):
   # These directories should be added to PATH
   path_add = get_required_path(tools_to_activate)
   # These already exist.
--- a/emsdk.py
+++ b/emsdk.py
@@ -2730,12 +2730,17 @@ def main():
    emsdk activate [--global] [--build=type] [--vs2017/--vs2019] <tool/sdk>
 
                                 - Activates the given tool or SDK in the
-                                  environment of the current shell. If the
-                                  --global option is passed, the registration
-                                  is done globally to all users in the system
-                                  environment.  If a custom compiler version was
-                                  used to override the compiler to use, pass
-                                  the same --vs2017/--vs2019 parameter
+                                  environment of the current shell.
+
+                                - If the `--global` option is passed, the the environment
+                                  variables are set permanently for the current user.
+
+                                - If the `--system` option is passed, the registration
+                                  is done for all users of the system
+                                  (uses Machine environment variables).
+
+                                - If a custom compiler version was used to override
+                                  the compiler to use, pass the same --vs2017/--vs2019 parameter
                                   here to choose which version to activate.
 
    emcmdprompt.bat              - Spawns a new command prompt window with the
--- a/emsdk.py
+++ b/emsdk.py
@@ -1760,18 +1760,13 @@ class Tool(object):
           return False
     return True
 
-  def win_activate_env_vars(self, permanently_activate):
+  def win_activate_env_vars(self, system):
     if WINDOWS:
       envs = self.activated_environment()
       for env in envs:
         key, value = parse_key_value(env)
 
-        if permanently_activate:
-          # If there is an env var for the LOCAL USER with same name, it will
-          # hide the system var, so must remove that first.
-          win_delete_environment_variable(key, False)
-
-        win_set_environment_variable(key, value, permanently_activate)
+        win_set_environment_variable(key, value, system)
 
   # If this tool can be installed on this system, this function returns True.
   # Otherwise, this function returns a string that describes the reason why this
@@ -2411,7 +2406,7 @@ def write_set_env_script(env_string):
 # Reconfigure .emscripten to choose the currently activated toolset, set PATH
 # and other environment variables.
 # Returns the full list of deduced tools that are now active.
-def set_active_tools(tools_to_activate, permanently_activate):
+def set_active_tools(tools_to_activate, permanently_activate, system):
   tools_to_activate = process_tool_list(tools_to_activate, log_errors=True)
 
   if tools_to_activate:
@@ -2433,13 +2428,13 @@ def set_active_tools(tools_to_activate, permanently_activate):
   if WINDOWS and permanently_activate:
     # Individual env. vars
     for tool in tools_to_activate:
-      tool.win_activate_env_vars(permanently_activate=True)
+      tool.win_activate_env_vars(system=system)
 
     # PATH variable
-    newpath, added_items = adjusted_path(tools_to_activate, system_path_only=True)
+    newpath, added_items = adjusted_path(tools_to_activate, system=system)
     # Are there any actual changes?
     if newpath != os.environ['PATH']:
-      win_set_environment_variable('PATH', newpath, system=True)
+      win_set_environment_variable('PATH', newpath, system=system)
 
   return tools_to_activate
 
@@ -2485,17 +2480,12 @@ def to_msys_path(p):
 
 # Looks at the current PATH and adds and removes entries so that the PATH reflects
 # the set of given active tools.
-def adjusted_path(tools_to_activate, log_additions=False, system_path_only=False):
+def adjusted_path(tools_to_activate, log_additions=False, system=False):
   # These directories should be added to PATH
   path_add = get_required_path(tools_to_activate)
   # These already exist.
   if WINDOWS and not MSYS:
-    existing_path = win_get_environment_variable('PATH', system=True)
-    if not system_path_only:
-      current_user_path = win_get_environment_variable('PATH', system=False)
-      if current_user_path:
-        existing_path += ENVPATH_SEPARATOR + current_user_path
-    existing_path = existing_path.split(ENVPATH_SEPARATOR)
+    existing_path = win_get_environment_variable('PATH', system=system).split(ENVPATH_SEPARATOR)
   else:
     existing_path = os.environ['PATH'].split(ENVPATH_SEPARATOR)
   emsdk_root_path = to_unix_path(emsdk_path())
@@ -2776,6 +2766,7 @@ def main():
   arg_old = extract_bool_arg('--old')
   arg_uses = extract_bool_arg('--uses')
   arg_global = extract_bool_arg('--global')
+  arg_system = extract_bool_arg('--system')
   if extract_bool_arg('--embedded'):
     errlog('embedded mode is now the only mode available')
   if extract_bool_arg('--no-embedded'):
@@ -2987,7 +2978,7 @@ def main():
     if not tools_to_activate:
       errlog('No tools/SDKs specified to activate! Usage:\n   emsdk activate tool/sdk1 [tool/sdk2] [...]')
       return 1
-    active_tools = set_active_tools(tools_to_activate, permanently_activate=arg_global)
+    active_tools = set_active_tools(tools_to_activate, permanently_activate=arg_global, system=arg_system)
     if not active_tools:
       errlog('No tools/SDKs found to activate! Usage:\n   emsdk activate tool/sdk1 [tool/sdk2] [...]')
       return 1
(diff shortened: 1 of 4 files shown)


sbc100:
According to the comments in #468 we decided it was best *not* to change existing meaning of `--global`.

I think the best option was to add two new options `--persistent` (meaning set in the registry for the current user) and `--persistent-system` meaning set in the registry for all users.  

`--global` can the becomes a legacy alias for `--persistent-system`.

aminya:
I deprecated `--global`. Now, we have `--permanent` to set the environment variables permanently.

aminya:
I deprecated `--global`. Now, we have `--permanent` to set the environment variables permanently.

aminya:
I added the tests, which shows that pywin32 code part of emsdk is broken.

https://github.com/emscripten-core/emsdk/pull/91

saschanaz:
Yay, thanks for taking over!

emsdk.py
```diff
@@ -2775,7 +2770,14 @@ def extract_bool_arg(name):
 
   arg_old = extract_bool_arg('--old')
   arg_uses = extract_bool_arg('--uses')
+  arg_permanent = extract_bool_arg('--permanent')
```
sbc100, line 2773: Could you use persistent as discussed in the other bug?  I know that permanent and persistent are essentially synonyms but my sense is that persistent is more commonly used the tech world.

emsdk.py
```diff
@@ -2775,7 +2770,14 @@ def extract_bool_arg(name):
...
   arg_global = extract_bool_arg('--global')
+  if arg_global:
+    print('--global is deprecated. Use `--system` to set the environment variables for all users')
+  arg_system = extract_bool_arg('--system')
+  if arg_system:
+    arg_permanent = True
+    arg_global = True
```
sbc100, line 2780: Did you mean to set arg_global here?  I would imagine `arg_global` should no longer be used anywhere after this change?

emsdk.py
```diff
@@ -2775,7 +2770,14 @@ def extract_bool_arg(name):
 
   arg_old = extract_bool_arg('--old')
   arg_uses = extract_bool_arg('--uses')
+  arg_permanent = extract_bool_arg('--permanent')
   arg_global = extract_bool_arg('--global')
+  if arg_global:
+    print('--global is deprecated. Use `--system` to set the environment variables for all users')
```
sbc100, line 2776: Why not print the warning but continue to do the right thing and set arg_system here?  We can phase it out eventually.

scripts/test_permanent.ps1
```diff
@@ -0,0 +1,67 @@
...
+    $EMSDK_UPSTREAM_Path_USER = $path_split | Where-Object { $_ -match 'emsdk\\upstream\\emscripten' }
+    if (!$EMSDK_UPSTREAM_Path_USER) {
+        throw "emsdk\upstream\emscripten is not added to path."
+    }
+
+}
+finally
+{
+# return back to cwd
+cd $cwd
```
sbc100, line 66: If you don't do this then does the directory change effect the calling shell?

sbc100 (commented):
I've not had a chance to look at the new test script but generally this change lgtm!

emsdk.py
aminya: That is the term that has been used in the readme: https://github.com/emscripten-core/emsdk#how-do-i-change-the-currently-active-sdk-version

I don't mind changing it, but I found permanent more natural.

emsdk.py
aminya: Fixed in https://github.com/emscripten-core/emsdk/pull/621/commits/25efaf464a58404ecc72d39342a375df6a43f866

emsdk.py
aminya: Fixed in https://github.com/emscripten-core/emsdk/pull/621/commits/bd028b9b6278e6e7b58b972c11de62ce3b0e3748

scripts/test_permanent.ps1
aminya: This is so if the program encounters an error or the user press CTRL+C, the current working directory will be restored to where it was. This is a safe way to change the current working directory.

sbc100: Just to be clear, it sounds like the `cd` command effect *does* effect the calling process?   e.g. if I run `foo.sp1` can it change the current working directory in my shell or are the changes limited to the duration of the `foo.ps1` script?

aminya: Yes. It does. It changes the current working directory of the parent process. That's why I need to use `finally`.

aminya: Another way is to use `Push-Location` in `try` and then call `Pop-Location` in the `finally`. It is the same as my method.

kripken:
Is the CI error the last issue here? (maybe also the permanent/persistent naming question)

sbc100:
I think we need a tie breaker :)   @kripken @juj how to do you feel about `--persistent` vs `--permanent` . 

As a non-windows-user I don't care too much either way.

aminya:
> Is the CI error the last issue here?

This is not an issue of this PR. I added the tests, so we know that environment variables are not set on Windows (the behavior is broken). We need to fix the tests in other pull requests such as in #91.

> (maybe also the permanent/persistent naming question)

I don't feel strongly about this. I think `--permanent` is more descriptive. `--persistent` is a little fuzzy since it does not explicitly say persisting between what.

sbc100:
> > Is the CI error the last issue here?
> 
> This is not an issue of this PR. I added the tests, so we know that environment variables are not set on Windows (the behavior is broken). We need to fix the tests in other pull requests such as in #91.
> 

Can you remove the tests from this PR then?   somehow disable them or mark them as currently known/expected to fail?  We don't want to land a change with the CI being red.

scripts/test_permanent.ps1
```diff
@@ -0,0 +1,67 @@
+$cwd=(Get-Location).Path
+
+try {
+
+    cd([System.IO.Path]::GetDirectoryName((resolve-path "$PSScriptRoot")))
+
+    ./emsdk.ps1 install latest
+    ./emsdk.ps1 activate latest --permanently
```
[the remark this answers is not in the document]
aminya: Fixed in https://github.com/emscripten-core/emsdk/pull/621/commits/405cf3baba48f2d49544d2b143522673f853b4e9

aminya:
> > > Is the CI error the last issue here?
> > 
> > 
> > This is not an issue of this PR. I added the tests, so we know that environment variables are not set on Windows (the behavior is broken). We need to fix the tests in other pull requests such as in #91.
> 
> Can you remove the tests from this PR then? somehow disable them or mark them as currently known/expected to fail? We don't want to land a change with the CI being red.

I caught the failures in https://github.com/emscripten-core/emsdk/pull/621/commits/f2575062c60e438a75e3e3ecc7efa7d28e56e6c5

This PR is ready to go.

kripken:
I don't have a preference between `--persistent` vs `--permanent `.

I think if the code looks good to you @sbc100 , let's land it!

emsdk.py
```diff
@@ -2971,8 +2973,8 @@ def print_tools(t):
     fetch_emscripten_tags()
     return 0
   elif cmd == 'activate':
-    if arg_global:
-      print('Registering active Emscripten environment globally for all users.')
+    if arg_permanent:
+      print('Registering active Emscripten environment premenantly')
```
sbc100, line 2977: sp: permanently

emsdk.py
```diff
@@ -2775,7 +2770,14 @@ def extract_bool_arg(name):
 
   arg_old = extract_bool_arg('--old')
   arg_uses = extract_bool_arg('--uses')
+  arg_permanent = extract_bool_arg('--permanent')
   arg_global = extract_bool_arg('--global')
+  if arg_global:
+    print('--global is deprecated. Use `--system` to set the environment variables for all users')
+    arg_permanent = True
```
sbc100, line 2777: I think the old behaviour was the "system" behaviour so this should probably be `arg_system = True` and `arg_permenant = True` and happen after the arg_system handler?

emsdk.py
```diff
@@ -2737,15 +2727,20 @@ def main():
...
                                 - Activates the given tool or SDK in the
-                                  environment of the current shell. If the
-                                  --global option is passed, the registration
-                                  is done globally to all users in the system
-                                  environment.  If a custom compiler version was
-                                  used to override the compiler to use, pass
-                                  the same --vs2017/--vs2019 parameter
+                                  environment of the current shell.
+
+                                - If the `--permanent` option is passed, the the environment
```
sbc100, line 2735: the the -> then the

emsdk.py
```diff
@@ -2737,15 +2727,20 @@ def main():
...
+                                  variables are set permanently for the current user.
+
+                                - If the `--system` option is passed, the registration
+                                  is done for all users of the system
+                                  (uses Machine environment variables).
```
sbc100, line 2740: Should we mention this requires admin privs here?

sbc100 (approved):
lgtm with a couple more comments.   Ready to land once they are addressed.

emsdk.py
aminya, line 2977:
```suggestion
      print('Registering active Emscripten environment permanently')
```

emsdk.py
aminya, line 2735:
```suggestion
                                - If the `--permanent` option is passed, then the environment
```

emsdk.py
```diff
@@ -2737,15 +2727,20 @@ def main():
...
-                                  used to override the compiler to use, pass
-                                  the same --vs2017/--vs2019 parameter
+                                  environment of the current shell.
+
+                                - If the `--permanent` option is passed, then the environment
+                                  variables are set permanently for the current user.
+
+                                - If the `--system` option is passed, the registration
+                                  is done for all users of the system
+                                  (uses Machine environment variables).
```
aminya, line 2740:
```suggestion
                                  is done for all users of the system. 
                                  This needs admin privileges
                                  (uses Machine environment variables).
```

emsdk.py
```diff
@@ -2775,7 +2771,14 @@ def extract_bool_arg(name):
 
   arg_old = extract_bool_arg('--old')
   arg_uses = extract_bool_arg('--uses')
+  arg_permanent = extract_bool_arg('--permanent')
   arg_global = extract_bool_arg('--global')
+  if arg_global:
+    print('--global is deprecated. Use `--system` to set the environment variables for all users')
+    arg_permanent = True
```
aminya, line 2778:
```suggestion
    arg_system = True
```

emsdk.py
aminya: https://github.com/emscripten-core/emsdk/pull/621/commits/0f85911cfad108444f97c4e40402a592845fc07e

emsdk.py
sbc100: Sorry I think this still doesn't work becaue arg_system is assigned on the line below.

sbc100: I think if you just move it below it should work.

aminya: It is fixed now.

sbc100:
Looks like the removal of the second call to causes the issue:

```
current_user_path = win_get_environment_variable('PATH', system=False)
```

Previously it was adding the user and system PATHs together I think.

I guess we should probably just ignore the registry when running in normal (non-persistent) mode and just look at the current os.environ['PATH']?

PR #627, opened 2020-09-30
Minor fix to emsdk.bat file

kudaba:
It seems that using ``::`` style comments in bat file function blocks causes weird errors in the command line as it's interpreted as a path.

https://stackoverflow.com/questions/12407800/which-comment-style-should-i-use-in-batch-files/12408045#12408045

### DIFF
--- a/emsdk.bat
+++ b/emsdk.bat
@@ -4,11 +4,11 @@
 
 setlocal
 
+:: When using our bundled python we never want the users
+:: PYTHONHOME or PYTHONPATH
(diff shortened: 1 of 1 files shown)


sbc100:
Oh god.. horrible.

Why does moving the comment fix the issue ?

sbc100 (approved):

kudaba:
It seems it's only a problem in the function block when using the style
```bat
rem comments out here have no problem with ::
if condition (
rem comments in here break when using :: style
action
)
```

Issue #385, opened 2019-11-04
MSYS2: Exception: Duplicate sdk sdk-incoming-32bit! 


MayaPosch:
Trying to run the emsdk script on current MSYS2 (Windows 7 x64) with Python 3.7.4:

```
$ ./emsdk install latest
Traceback (most recent call last):
  File "./emsdk.py", line 2991, in <module>
    sys.exit(main())
  File "./emsdk.py", line 2680, in main
    load_sdk_manifest()
  File "./emsdk.py", line 2230, in load_sdk_manifest
    add_sdk(sdk)
  File "./emsdk.py", line 1843, in add_sdk
    raise Exception('Duplicate sdk ' + str(sdk) + '! Existing:\n{' + ', '.join("%s: %s" % item for item in vars(find_sdk(str(sdk))).items()) + '}, New:\n{' + ', '.join("%s: %s" % item for item in vars(sdk).items()) + '}')
Exception: Duplicate sdk sdk-incoming-32bit! Existing:
{version: incoming, bitness: 32, uses: ['clang-incoming-32bit', 'node-12.9.1-32bit', 'python-2.7.13.1-32bit', 'java-8.152-32bit', 'emscripten-incoming-32bit', 'binaryen-master-32bit'], os: win, id: sdk, name: sdk-incoming-32bit, is_old: False, is_sdk: True}, New:
{version: incoming, bitness: 32, uses: ['clang-incoming-32bit', 'node-12.9.1-32bit', 'emscripten-incoming-32bit', 'binaryen-master-32bit'], os: linux, id: sdk, name: sdk-incoming-32bit, is_old: False, is_sdk: True}
```

Managed to install the SDK by commenting out the 'find_sdk()' check in the emsdk.py script (lines 1,842 & 1,843) and manually copying the contents of the binaries ZIP to the 'upstream' directory when the script died again while activating, as it couldn't handle UNIX style paths on Windows, apparently, and errored out during copying the unzipped files.

TimMensch:
Same error here, but with 64 bits:

```
Traceback (most recent call last):
  File "./emsdk.py", line 3147, in <module>
    sys.exit(main())
  File "./emsdk.py", line 2847, in main
    load_sdk_manifest()
  File "./emsdk.py", line 2411, in load_sdk_manifest
    add_sdk(sdk)
  File "./emsdk.py", line 2024, in add_sdk
[...]

PR #628, opened 2020-10-01
Fix MSYS2/MinGW python mistaking system as Linux

mmcparlane:
When running `source emsdk_env.sh` along with python provided by MSYS2/MinGW the environment gets falsely identified as Linux.

See discussion for details:
https://github.com/emscripten-core/emscripten/issues/12376#issuecomment-702245993

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -74,7 +74,7 @@ if platform.mac_ver()[0] != '':
   MACOS = True
 
 LINUX = False
-if not MACOS and (platform.system() == 'Linux' or os.name == 'posix'):
+if not MACOS and (platform.system() == 'Linux'):
   LINUX = True
 
 UNIX = (MACOS or LINUX)


sbc100:
IIUC that problem is not related to MSYS2/MinGW terminals but the use of msys2 version of python to run emsdk.py?   If I'm correct can you update the PR description and title?

In fact could you not recreate the but in the windows cmd prompt by running under msys2 version of python?

I'm happy to land this as is but it would also be create to get some basic testing under msys2 python at some point.   How hard would it be to install msys2 python in the `test-windows` part of our circle CI do you think?

sbc100:
Fixes #385

mmcparlane:
> IIUC that problem is not related to MSYS2/MinGW terminals but the use of msys2 version of python to run emsdk.py? If I'm correct can you update the PR description and title?

Yep, that sounds correct—I'll update it.

> I'm happy to land this as is but it would also be create to get some basic testing under msys2 python at some point. How hard would it be to install msys2 python in the test-windows part of our circle CI do you think?

It shouldn't be too bad to add the tests. Would you like me to do it as part of this PR?

sbc100:
Do you know how to install python in the "git for windows" bash environment that is available in the circleci setup?

sbc100:
I think we can land this change as is .. and work on getting gets added as a followup.

Can you update the description and I will land/

sbc100 (approved):

mmcparlane:
> Can you update the description and I will land/

Finished updating.

> Do you know how to install python in the "git for windows" bash environment that is available in the circleci setup?

I'm not sure off hand but I'd be inclined to add a separate one for msys32 bash. That way we can control what is installed on there more tightly. This should probably be stored in a docker image for speed reasons.

sbc100:
When running `emsdk.py` along with python provided by MSYS2/MinGW the environment gets falsely identified as Linux.

See discussion for details:
https://github.com/emscripten-core/emscripten/issues/12376#issuecomment-702245993

Fixes #385

PR #631, opened 2020-10-03
Add test-msys2 job to circleci pipeline

mtmcp:
Run steps are modeled on the test-linux job but executed inside an MSYS2 shell.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -191,6 +191,7 @@ workflows:
   test-windows:
     jobs:
       - test-windows
+      - test-msys2
   build-docker-image:
     jobs:
       - build-docker-image
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -111,16 +111,16 @@ jobs:
           command: choco install --confirm msys2
       - run:
           name: Install msys2 packages
-          command: $MSYS_CMD "pacman --sync --noconfirm --needed base-devel cmake gcc python"
+          command: %MSYS_CMD% "pacman --sync --noconfirm --needed base-devel cmake gcc python"
       - run:
           name: test.sh
-          command: $MSYS_CMD "export EMSDK_NOTTY='1' && /usr/bin/sh.exe scripts/test.sh"
+          command: %MSYS_CMD% "export EMSDK_NOTTY='1' && /usr/bin/sh.exe scripts/test.sh"
       - run:
           name: test_source_env.sh
-          command: $MSYS_CMD "export EMSDK_NOTTY='1' && /usr/bin/sh.exe scripts/test_source_env.sh"
+          command: %MSYS_CMD% "export EMSDK_NOTTY='1' && /usr/bin/sh.exe scripts/test_source_env.sh"
       - run:
           name: test.py
-          command: $MSYS_CMD "source emsdk_env.sh && export EMSDK_NOTTY='1' && /usr/bin/python scripts/test.py"
+          command: %MSYS_CMD% "source emsdk_env.sh && export EMSDK_NOTTY='1' && /usr/bin/python scripts/test.py"
 
   build-docker-image:
     executor: bionic
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,6 +97,26 @@ jobs:
               Write-Host "`n Error Message: [$($_.Exception.Message)"] -ForegroundColor Red
               Write-Host "`n Line Number: "$_.InvocationInfo.ScriptLineNumber -ForegroundColor Red
             }
+            
+  test-msys2:
+    executor:
+      name: win/vs2019
+      shell: cmd.exe
+    environment:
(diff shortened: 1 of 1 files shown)


mmcparlane:
Hmm.. It's not showing the Windows tests on my side, and also not showing the newly added test-msys2. Perhaps due to my personal circleci not having credits... Can someone else verify these?

kripken:
I think you need to also add the new job to the list of jobs to run, and not just define it. See the list near the end of the file, under `workflows:`

mmcparlane:
Oops. Easy enough!

mmcparlane:
Yikes. The `base-devel` installs way too many packages. I'll have to remove all but the essentials to run the basic test scripts...

mmcparlane:
Anyone want to test this PR?

.circleci/config.yml
```diff
@@ -92,6 +92,44 @@ jobs:
...
+          command: C:\msys64\usr\bin\bash -lc 'pacman --noconfirm --sync --needed msys2-devel python cmake'
+      - run:
+          name: test.sh
+          command: C:\msys64\usr\bin\bash -lc "export EMSDK_NOTTY='1' && ./scripts/test.sh"
+      - run:
+          name: test_source_env.sh
+          command: C:\msys64\usr\bin\bash -lc "export EMSDK_NOTTY='1' && ./scripts/test_source_env.sh"
+      - run:
+          name: test.py
+          command: C:\msys64\usr\bin\bash -lc "source emsdk_env.sh && export EMSDK_NOTTY='1' && /usr/bin/python scripts/test.py"
```
sbc100, line 132: BTW, although it would be nice to run emsdk itself under msys python, my current thinking is that we will require windows python when running emcc itself.

This is because emcc is a wrapper around a windows (not msys) version of llvm.   i.e,  I think its reasonable for emcc, just like llvm which is wraps to accept *only* valid windows paths as inputs.

Not that this work you are doing won't be useful as way to bootstrap emsdk which can then install windows python.

mmcparlane: It actually shouldn't matter which python is run because the next set of changes to get this working will be to update the python scripts to recognize (via environment variable) the msys/mingw environments so it can adjust the path separators properly. But it makes sense to use the system python as most users will have that anyhow—saves the CI system some cycles/bandwidth too! ⚡

sbc100: Do you mean the emscripten python scripts?

Do you imagine `emcc /mnt/c/path/to/src.cpp -L/mnt/c/lib`  working?    I think the amount of changes to emcc required to make that would be a lot.    Also.. even if you convert all those paths the windows paths that llvm/clang would need under the hood.. you than get clang error message that contains the windows paths ("C:\path\to\src") so now you need to think about capuring the clang output and do string replacements within it just to get reasonable errors messages... I think it will be long tail of annoying work..   it might be easier to instread require an msys version of llvm.. then everything should just work.

In summary:  I believe to get emcc working on msys2 you first need to get clang that works on msys2.

sbc100: Getting emsdk working on msys2 is different question.. we can support that without supporting msys paths in emcc.

mmcparlane: > In summary: I believe to get emcc working on msys2 you first need to get clang that works on msys2.

Ah, good point. Looks like I'll have to make a couple changes to use the mingw environment provided by msys2—prebuilt clang packages already provided:

```
mtmcp@JollyBox MSYS ~
$ pacman -Ss llvm
mingw32/mingw-w64-i686-clang 10.0.1-1
    C language family frontend for LLVM (mingw-w64)
mingw32/mingw-w64-i686-compiler-rt 10.0.1-1
    Runtime libraries for Clang and LLVM (mingw-w64)
mingw32/mingw-w64-i686-lld 10.0.1-1
    Linker tools for LLVM (mingw-w64)
mingw32/mingw-w64-i686-llvm 10.0.1-1
    Low Level Virtual Machine (mingw-w64)
mingw32/mingw-w64-i686-polly 10.0.1-1
    Polly - Polyhedral optimizations for LLVM (mingw-w64)
mingw64/mingw-w64-x86_64-clang 10.0.1-1
    C language family frontend for LLVM (mingw-w64)
mingw64/mingw-w64-x86_64-compiler-rt 10.0.1-1
    Runtime libraries for Clang and LLVM (mingw-w64)
mingw64/mingw-w64-x86_64-lld 10.0.1-1
    Linker tools for LLVM (mingw-w64)
mingw64/mingw-w64-x86_64-llvm 10.0.1-1
    Low Level Virtual Machine (mingw-w64)
mingw64/mingw-w64-x86_64-polly 10.0.1-1
    Polly - Polyhedral optimizations for LLVM (mingw-w64)
```

> Getting emsdk working on msys2 is different question.. we can support that without supporting msys paths in emcc.

Personally, I would probably just use Windows shells to pull down emsdk stuff (that way I could build on Windows as well). In that regard, what is the goal in terms of the test scripts? Do we want to be able to run all of these from msys2?

1. scripts/test.sh
2. scripts/test_source_env.sh
3. scripts/test.py

PR #635, opened 2020-10-07
No win32api

aminya:
Revives and Closes https://github.com/emscripten-core/emsdk/pull/91


emsdk.py
```diff
@@ -294,11 +294,11 @@ def remove_readonly_and_try_again(func, path, exc_info):
...
-      import win32api
-      import win32con
-      return win32api, win32con
-    except Exception:
-      exit_with_error('Failed to import Python Windows extensions win32api and win32con. Make sure you are using the version of python available in emsdk, or install PyWin extensions to the distribution of Python you are attempting to use. (This script was launched in python instance from "' + sys.executable + '")')
+        import winreg
+    except ImportError:
+        import _winreg as winreg
+    import ctypes
+    return winreg, ctypes
```
sbc100, line 301: Given that these are builtin module can we just do this at the top level no?   No need to delay the import I think.

emsdk.py
```diff
@@ -294,11 +294,11 @@ def remove_readonly_and_try_again(func, path, exc_info):
...
   if WINDOWS:
     try:
-      import win32api
-      import win32con
-      return win32api, win32con
-    except Exception:
-      exit_with_error('Failed to import Python Windows extensions win32api and win32con. Make sure you are using the version of python available in emsdk, or install PyWin extensions to the distribution of Python you are attempting to use. (This script was launched in python instance from "' + sys.executable + '")')
+        import winreg
+    except ImportError:
+        import _winreg as winreg
```
sbc100, line 299: Can you add a comment saying something like "python 2 support"... or "old python 2 name"

sbc100 (approved):
This is great!

Is it really this simple?  We don't loose anything in this transition?

emsdk.py
aminya: Lazy loading will still have performance benefits. Python is not intelligent enough to no spend time importing these when they are not used.

emsdk.py
aminya:
```suggestion
		# old python 2 name
        import _winreg as winreg
```

aminya:
> This is great!
> 
> Is it really this simple? We don't loose anything in this transition?

Thank you! Based on my personal tests and the unit tests, there is no behavior change.

If we can merge #515 before this, we can become fully sure that this works.

emsdk.py
sbc100: I don't think we need to worry about the performance penalty of loading winreg and ctypes in this script.

I believe these imports were imported late like this we could report an error and handle them being missing.   Now that can't happen which is great, so there is no need to delay the importing IMHO.

Keeping all imports at the top level helps make the program more readable IMHO.

aminya: Here you go: https://github.com/emscripten-core/emsdk/pull/635/commits/1590416a9a08f6028a0068b9e27897d947294e7b

sbc100: Alternatively maybe just inline these imports the top of `win_set_environment_variable_direct` before the `try` block?

And you can drop the `if WINDOWS` too since `win_set_environment_variable_direct` is already windows only?

sbc100:
I'm happy to land this now if you think its safe.    I've not yet has a change to look at #515 but it looks a little more invasive.   Do you want to land this now or wait for #515?

aminya:
I tested this with #515 on my master branch. It works:
```
Registering active Emscripten environment permanently

Setting the following tools as active:
   node-12.18.1-64bit
   python-3.7.4-pywin32-64bit
   java-8.152-64bit

Adding directories to PATH:
PATH += C:\Users\aminy\Documents\GitHub\Cpp\emsdk
PATH += C:\Users\aminy\Documents\GitHub\Cpp\emsdk\node\12.18.1_64bit\bin
PATH += C:\Users\aminy\Documents\GitHub\Cpp\emsdk\python\3.7.4-pywin32_64bit
PATH += C:\Users\aminy\Documents\GitHub\Cpp\emsdk\java\8.152_64bit\bin

Setting global environment variables:
PATH = C:\Users\aminy\Documents\GitHub\Cpp\emsdk;C:\Users\aminy\Documents\GitHub\Cpp\emsdk\node\12.18.1_64bit\bin;C:\Users\aminy\Documents\GitHub\Cpp\emsdk\python\3.7.4-pywin32_64bit;C:\Users\aminy\Documents\GitHub\Cpp\emsdk\java\8.152_64bit\bin;C:\Program Files\Conan\conan;C:\Users\aminy\AppData\Local\atom\bin;C:\Program Files (x86)\JetBrains\WebStorm 2020.2\bin;C:\Users\aminy\AppData\Local\atom-nightly\bin;C:\Program Files\Google\Chrome\Application;C:\Program Files\CMake\bin
EMSDK = C:/Users/aminy/Documents/GitHub/Cpp/emsdk
EM_CONFIG = C:\Users\aminy\Documents\GitHub\Cpp\emsdk\.emscripten
EMSDK_NODE = C:\Users\aminy\Documents\GitHub\Cpp\emsdk\node\12.18.1_64bit\bin\node.exe
EMSDK_PYTHON = C:\Users\aminy\Documents\GitHub\Cpp\emsdk\python\3.7.4-pywin32_64bit\python.exe
JAVA_HOME = C:\Users\aminy\Documents\GitHub\Cpp\emsdk\java\8.152_64bit
```

emsdk.py
```diff
@@ -340,15 +338,13 @@ def win_get_environment_variable(key, system=True):
       py_path = to_native_path(py.expand_vars(py.activated_path))
       os.environ['PATH'] = os.environ['PATH'] + ';' + py_path
```
sbc100, line 339: Out of interest do you have any insight as to why we over ride `$PATH` during this function?

aminya: Good catch. I think we don't need this anymore! This is adding the python that ships with emsdk to PATH to perform win32api operations. Considering that it is removed, we can get rid of these.

aminya: https://github.com/emscripten-core/emsdk/pull/635/commits/990f370945bc829c005bb7c5619ec1e7780db37f

Issue #632, opened 2020-10-06
emsdk 'manual install' of google-closure-compiler-[*] packages breaks all non-x86 source builds


ebeasant-arm:
PR #630  breaks all non-x86 from source emsdk builds, because the specified packages in the npm invocation are arch-specific for x86.

curiousdannii:
Ah, sorry, I didn't think of that. Anyone know the best path from here? Change to `npm install`?

kripken:
First let's revert the breakage, I think, to stop people from hitting this right now. I opened https://github.com/emscripten-core/emsdk/pull/633

sbc100:
I think the solution is to check for OS = linux and ARCH = x86_64?

ebeasant-arm:
I guess that a slightly more refined logic would work, and keep the image size down on the x86/CI platforms, whilst permitting fallback to the top level 'bloaty' package with other arches: Npm won't install the optional dependencies anyway on those platforms, so... win/win?

ebeasant-arm:
> generic google-closure-compiler should work in the absence of the arch-specific package.
Data point: tested that here, works nicely 👍

PR #630, opened 2020-10-02
Skip installing optional npm deps, but manually install Closure Compiler native package

curiousdannii:
Follow on to https://github.com/emscripten-core/emsdk/pull/625
Fixes https://github.com/emscripten-core/emscripten/issues/12406

PR #636, opened 2020-10-08
Skip installing optional npm deps, but manually install Closure Compiler native package

curiousdannii:
Now checks ARCH

Follow on to #625, #630, #632, #633
Fixes emscripten-core/emscripten#12406

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1223,12 +1223,40 @@ def emscripten_npm_install(tool, directory):
   print('Running post-install step: npm ci ...')
   try:
     subprocess.check_output(
-        [npm, 'ci', '--production'],
+        [npm, 'ci', '--production', '--no-optional'],
         cwd=directory, stderr=subprocess.STDOUT, env=env,
         universal_newlines=True)
   except subprocess.CalledProcessError as e:
     print('Error running %s:\n%s' % (e.cmd, e.output))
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -1223,13 +1223,41 @@ def emscripten_npm_install(tool, directory):
...
+  # There are two problems here:
+  #   1. npm ci does not consider the platform of optional dependencies
+  #      https://github.com/npm/cli/issues/558
+  #   2. A bug with the native compiler has bloated the packages from
+  #      30MB to almost 300MB
+  #      https://github.com/google/closure-compiler-npm/issues/186
+  # If either of these bugs are fixed then we can remove this exception
+  closure_compiler_native = ''
+  if LINUX and (ARCH == 'x86' or ARCH == 'x86_64'):
+    closure_compiler_native = 'google-closure-compiler-linux'
```
sbc100, line 1245: Do these binaries actually exist and run on x86 as will as x86_64?

You can do `.. and ARCH in ('x86' , 'x86_64'):` here

sbc100 (approved):

emsdk.py
curiousdannii: I checked which architectures the npm packages said they supported.

Issue #634, opened 2020-10-06
On Windows emsdk_env.bat removes most of the PATH environment variable


yowl:
Checkout commit with hash 807e55cbec8fec292045ebcaf547ed7466762d7f, start a command shell:
```
E:\>cd github

E:\GitHub>emsdk\emsdk_env.bat
The system cannot find the drive specified.
Adding directories to PATH:
PATH += E:\GitHub\emsdk
PATH += E:\GitHub\emsdk\node\12.18.1_64bit\bin
PATH += E:\GitHub\emsdk\python\3.7.4-pywin32_64bit
PATH += E:\GitHub\emsdk\java\8.152_64bit\bin

Setting environment variables:
EMSDK = E:/GitHub/emsdk
EM_CONFIG = E:\GitHub\emsdk\.emscripten
EMSDK_NODE = E:\GitHub\emsdk\node\12.18.1_64bit\bin\node.exe
EMSDK_PYTHON = E:\GitHub\emsdk\python\3.7.4-pywin32_64bit\python.exe
JAVA_HOME = E:\GitHub\emsdk\java\8.152_64bit

E:\GitHub>powershell
'powershell' is not recognized as an internal or external command,
operable program or batch file.

E:\GitHub>echo %PATH%
E:\GitHub\emsdk;E:\GitHub\emsdk\node\12.18.1_64bit\bin;E:\GitHub\emsdk\python\3.7.4-pywin32_64bit;E:\GitHub\emsdk\java\8.152_64bit\bin;C:\python3;C:\Users\xxxxx\AppData\Local\Microsoft\WindowsApps;C:\Users\xxxxx\.dotnet\tools;C:\Program Files\Wasmtime\bin;C:\Users\xxxxx\AppData\Local\Programs\Fiddler;C:\Program Files (x86)\Sophos\Sophos SSL VPN Client\bin;C:\Users\xxxxx\AppData\Roaming\npm;C:\Users\xxxxx\AppData\Local\Microsoft\WindowsApps;C:\Users\xxxxx\.dotnet\tools

E:\GitHub>
```
The PATH has been pretty much replaced.  Before running `emsdk_env.bat`  PATH is much longer, fills the screen.

Check out emsdk before this commit #621  and it is fine.

sbc100:
@aminya can you take a look?

aminya:
I don't quite understand the issue. What do you mean by it removes most of the PATH environment variable?

sbc100:
[...]

PR #638, opened 2020-10-08
Preserve windows system path when activating a tool

fibbo:
This is with regard to #634


aminya (changes requested):
That is irrelevant. `System=True` means that emsdk will try to access Machine environment variables instead of User even when the user does not have admin access.

This PR will probably solve this issue.
https://github.com/emscripten-core/emsdk/pull/515

aminya (changes requested):
This should be closed now. #634 was fixed in #642

PR #641, opened 2020-10-13
2.0.7

kripken:


curiousdannii:
Unfortunately the Docker image didn't go down in size. I haven't done any investigation yet.

sbc100:
I believe this change is required to return the size of the docker image to normal: https://github.com/emscripten-core/emsdk/pull/640

sbc100:
Actually it should make any difference :(

PR #642, opened 2020-10-14
fix: emsdk swallows/overwrites parts of PATH

### LINKED ISSUE #634: On Windows emsdk_env.bat removes most of the PATH environment variable

### LINKED ISSUE #645: When constructing PATH for current shell, we ignore current shell PATH
On windows, `emsdk activate` or `emsdk construct_env` is run we modify the current shell to include a working set of environment variables. 

However, when calculating the new `%PATH%` to set we look into the registry for the existing path and ignore the current shell's `%PATH%`

This means that any `%PATH%` elements that were added locally we get clobbered.  See #634 

I think the fix is that on windows we need to create two different path:

1. What path should be set in the registry
2. What path should be set in the current shell.

aminya:
Fixes #634
Fixes #645 

This PR has three changes: 

1) Fixing --permanent flag:
The previous algorithm looped over the values of PATH 4 times to check conditions while these conditions are the opposite of each other. Also, the result did not include all the parts in the whole_path. 

	I have combined these loops and fixed the algorithm, so no path is lost. Now using --permanent the PATH and Env variables are correctly updated.

2) Fixing --system flag:
Now, the system flag does not cause the PATH values to be overwritten/clobbered. The PATH/Env variables are correctly updated.

3) Fixing flagless flag:
When no flag is provided the value of process/shell is correctly updated and preserved.

This adds the tests for all the situations.


sbc100:
I can't see how this actaully fix the problem of clobbering the user's current PATH though.

Isn't the core problem that `existing_path` is being read from the registry rather than the environment?   What happens if I have made local additions to my path in my current shell?

emsdk.py
```diff
@@ -2510,13 +2510,24 @@ def adjusted_path(tools_to_activate, system=False):
...
+    else:
+      existing_nonemsdk_path.append(entry)
+
+  new_emsdk_tools = []
+  kept_emsdk_tools = []
+  for entry in path_add:
+    if not normalized_contains(existing_emsdk_tools, entry):
+      new_emsdk_tools.append(entry)
+    else:
+      kept_emsdk_tools.append(entry)
```
sbc100, line 2527: I'm not quite sure what the purpose of this loop is.    Its seems that it takes `path_add` and splits it into two lists but those lists then both used in their entirety below.. 

How is this different to just doing `whole_path = unique_items(new_path + existing_nonemsdk_path)` below and skipping this loop?   I guess ordering would be the only different?  Do we care about that?

aminya: new_emsdk_tools is returned as the second output of the function

sbc100: Ah I see, thanks!

Still, can we drop `kept_emsdk_tools` completely and use `whole_path = unique_items(new_path + existing_nonemsdk_path)`?

aminya: I think it is useful to be explicit. We can use this `kept_emsdk_tools` later to optimize some of the things. The following is true:
```
path_add == new_emsdk_tools + kept_emsdk_tools 
```

The following is what you meant. I have changed the names to prevent any confusion.

```python
whole_path = unique_items(path_add+ existing_nonemsdk_path)
```

emsdk.py
```diff
@@ -103,7 +103,10 @@ def errlog(msg):
     BASH = True
 
 if WINDOWS:
```
sbc100, line 105: How about `if WINDOWS and not MSYS:`

Also, maybe this can be split into its own change as it a seems like nice separable cleanup.

sbc100:
Can you give an example of a path that was previously lost but is not any longer?   i.e. what was the actual bug?

emsdk.py
aminya: Yeah. That sounds good. This can be a follow up.

aminya:
> Can you give an example of a path that was previously lost but is not any longer? i.e. what was the actual bug?

When I run `activate` using this branch no PATH is lost. No matter how many times I run this.



> I can't see how this actaully fix the problem of clobbering the user's current PATH though.
> 
> Isn't the core problem that `existing_path` is being read from the registry rather than the environment? What happens if I have made local additions to my path in my current shell?

The existing_path should be read based on the user-provided option `--system`. Reading PATH from the environment instead of the registry will result in exchanging the PATH variables. If the user has started this as admin, they will get the system PATH, but if they have not provided `--system`, their user system will be overwritten by the system PATH.

sbc100:
> > Can you give an example of a path that was previously lost but is not any longer? i.e. what was the actual bug?
> 
> When I run `activate` using this branch no PATH is lost. No matter how many times I run this.

Maybe I can rephrase:   Can we give an example of a sequence of commands that demonstrate the problem that exists on master today?   (I imagine it will look different to the sequence below where I demonstrate the problem I see, which  I think is separate / different).


> 
> > I can't see how this actaully fix the problem of clobbering the user's current PATH though.
> > Isn't the core problem that `existing_path` is being read from the registry rather than the environment? What happens if I have made local additions to my path in my current shell?
> 
> The existing_path should be read based on the user-provided option `--system`. Reading PATH from the environment instead of the registry will result in exchanging the PATH variables. If the user has started the shell as admin, they will get the system PATH, but if they have not provided `--system`, their user system will be overwritten by the system PATH.
> 
> We should not assume what the user wants based on their shell status, and instead of we should use `--system` as the guide.

But we read the PATH from the registry even when no `--system` or `--persistent` flags are passed.

Even if we fix that, what about this example:

```
$ set PATH=%PATH%:/path/to/important/tool
$ emsdk activate latest
$ ... can I still access my tool?
```

Since we unconditionally use the registry to get the PATH we always ignore the actual PATH in the environment, no?

aminya:
> But we read the PATH from the registry even when no `--system` or `--persistent` flags are passed.
> 
> Even if we fix that, what about this example:
> 
> ```
> $ set PATH=%PATH%:/path/to/important/tool
> $ emsdk activate latest
> $ ... can I still access my tool?
> ```
> 
> Since we unconditionally use the registry to get the PATH we always ignore the actual PATH in the environment, no?

I was inaccurate a little. When someone modifies the environment variables using `set, setx` or equivalent PowerShell APIs, they instantly appear in the registry, so, as long as, we read these from the registry they will include the changes.

My point was about changing the values only locally in the current shell. These changes will "not be saved" by the shell or our script. So, we don't need to be worried about those. 
```
$env:PATH = "something"
```

aminya:
> > > Can you give an example of a path that was previously lost but is not any longer? i.e. what was the actual bug?
> > 
> > 
> > When I run `activate` using this branch no PATH is lost. No matter how many times I run this.
> 
> Maybe I can rephrase: Can we give an example of a sequence of commands that demonstrate the problem that exists on master today? (I imagine it will look different to the sequence below where I demonstrate the problem I see, which I think is separate / different).

I have not read all of these 2K-3K lines of code. I just saw a problem in this area. I think this fixes all or some parts of the issue.

Currently, when I use 
```
./emsdk activate latest
```
or 
```
./emsdk activate latest --permanent
```

The PATH values are set correctly without the removal of anything.

sbc100:
> > But we read the PATH from the registry even when no `--system` or `--persistent` flags are passed.
> > Even if we fix that, what about this example:
> > ```
> > $ set PATH=%PATH%:/path/to/important/tool
> > $ emsdk activate latest
> > $ ... can I still access my tool?
> > ```
> > 
> > 
> > Since we unconditionally use the registry to get the PATH we always ignore the actual PATH in the environment, no?
> 
> I was inaccurate a little. When someone modifies the environment variables using `set, setx` or equivalent PowerShell APIs (`[Environment]::SetEnvironmentVariable`), they instantly appear in the registry, so, as long as, we read these from the registry they will include the changes.
> 

Are you sure about that?  I though the whole point of `set` vs `setx` is that one is local, and one is written back to the registry: 

https://superuser.com/questions/916649/what-is-the-difference-between-setx-and-set-in-environment-variables-in-windows

If this is true, as it seems to be, a user can use `set` to add thing to their shell-local-environment.. making the local PATH different to anything in the registry.

My question is, is this the bug that was reported in #634  ?   i.e. did that user have a locally-modified PATH.


Getting back to this fix, were you able to reproduce the problem locally when making this fix?  If so how did you do it?   I don't see how the old code can fail to include stuff that was in `existing_path`.   The oldcode does:

```
existing_path = [i for i in existing_path if not to_unix_path(i).startswith(emsdk_root_path)]
whole_path = unique_items(new_path + existing_path)
```

Can you explain where the bug is?  I don't see how elements from existing_path are being dropped?   I guess what I'm saying is, I don't yet understand where bug lies?  Do you?

sbc100:
Sorry for all the back and forth on this.  I'm not trying to be difficult and I think we do want this patch, I just want to have a good understanding of what exactly the issue is.

aminya:
> > I was inaccurate a little. When someone modifies the environment variables using `set, setx` or equivalent PowerShell APIs (`[Environment]::SetEnvironmentVariable`), they instantly appear in the registry, so, as long as, we read these from the registry they will include the changes.
> 
> Are you sure about that? I though the whole point of `set` vs `setx` is that one is local, and one is written back to the registry:
> 
> https://superuser.com/questions/916649/what-is-the-difference-between-setx-and-set-in-environment-variables-in-windows

You are right about `set`. I don't use CMD a lot. Anyways, my point is that:

**We should not use shell (process) variables to set/get PATH or other environment variables**

aminya:
If this is what you are saying:
```
❯ $env:PATH="temp"
❯ $env:PATH
temp
> .\emsdk activate latest
❯ $env:PATH
C:\emsdk;C:\emsdk\upstream\emscripten;C:\emsdk\node\12.18.1_64bit\bin;C:\emsdk\python\3.7.4-pywin32_64bit;C:\emsdk\java\8.152_64bit\bin;C:\AppData\Roaming\npm;C:\ProgramData\chocolatey\bin;C:\Python39\Scripts\;C:\Python39\;C:\Program Files (x86)\Common Files\Oracle\Java\javapath;C:\Program Files\PowerShell\7\;C:\Program Files\Cppcheck;C:\Program Files\doxygen\bin;C:\Program Files\Git\cmd;C:\Program Files\PuTTY\;C:\Program Files\dotnet\;C:\Program Files\Microsoft SQL Server\130\Tools\Binn\;C:\Program Files\gnuplot\bin;C:\Program Files\Microsoft VS Code\bin;C:\Program Files\nodejs\;C:\Program Files\LLVM\bin
```

Yes, we always read the PATH from the registry even if `--permanent` or `--system` are not passed. But the temporary value of `$env:PATH` should not be important. Keeping the current shell values is another issue.

sbc100:
> If this is what you are saying:
> 
> ```powershell
> ❯ $env:PATH="temp"
> ❯ $env:PATH
> temp
> > .\emsdk activate latest
> ❯ $env:PATH
> C:\emsdk;C:\emsdk\upstream\emscripten;C:\emsdk\node\12.18.1_64bit\bin;C:\emsdk\python\3.7.4-pywin32_64bit;C:\emsdk\java\8.152_64bit\bin;C:\AppData\Roaming\npm;C:\ProgramData\chocolatey\bin;C:\Python39\Scripts\;C:\Python39\;C:\Program Files (x86)\Common Files\Oracle\Java\javapath;C:\Program Files\PowerShell\7\;C:\Program Files\Cppcheck;C:\Program Files\doxygen\bin;C:\Program Files\Git\cmd;C:\Program Files\PuTTY\;C:\Program Files\dotnet\;C:\Program Files\Microsoft SQL Server\130\Tools\Binn\;C:\Program Files\gnuplot\bin;C:\Program Files\Microsoft VS Code\bin;C:\Program Files\nodejs\;C:\Program Files\LLVM\bin
> ```
> 
> Yes, we always read the PATH from the registry even if `--permanent` or `--system` are not passed. But the temporary value of `$env:PATH` should not be important. Here, we can add an option to `win_get_env` variable function to get values from the current shell when `--permanent` or `--system` are not passed. But keeping the current shell values is another issue.

I agree it could well be separate issue,  but are we sure this is not the issue being reported in #634?    

Have you been able to reproduce some other issue where PATH elements are dropped?

aminya:
> Can you explain where the bug is? I don't see how elements from existing_path are being dropped? I guess what I'm saying is, I don't yet understand where bug lies? Do you?

The previous algorithm does not make sense to me. 

```
  # I merged these two together. In my code, `existing_nonemsdk_tools` is kept intact and so there is no chance of loosing any entry from that portion.
  existing_emsdk_tools = [i for i in existing_path if to_unix_path(i).startswith(emsdk_root_path)]
  existing_path = [i for i in existing_path if not to_unix_path(i).startswith(emsdk_root_path)]

  # this is translated to my second loop
  new_emsdk_tools = [i for i in path_add if not normalized_contains(existing_emsdk_tools, i)]
  
  # this is a redundant check which does not make sense at all. `existing_path` does not include any emsdk tools at this line of code!
  new_path = [i for i in path_add if not normalized_contains(existing_path, i)]

  whole_path = unique_items(new_path + existing_path)
```

sbc100:
But is there any way that an element of `existing_path` (that is outside of emsdk) can be dropped this algorithm.

It seems to be that existing_path which comes from the registry is clearly preserved (at least the parts that are not under emsdk).

Perhaps his is another clue:

https://superuser.com/questions/867728/user-vs-system-environment-variables-do-system-variables-override-user-variabl

```
The Path is constructed from the system path, which can be viewed in the System Environment Variables field in the System dialog box. The User path is appended to the system path.
```

This is apparently from MS.   So maybe the problem is that we are not taking into account both user and system variables when constructing the new path?

aminya:
Can we merge this? I have removed the linked issue. The changes are still valid.

aminya:
I added PATH preservation tests. Now, we can make sure that nothing is removed from the PATH!

aminya:
I added PATH preservation tests. Now, we can exactly test if anything is removed from the PATH!

aminya:
OK. I fixed the issue. 😃 The tests helped me to find the issue. This is fixed in 
https://github.com/emscripten-core/emsdk/pull/642/commits/aa10aaaaa85734a7b7becd517334792678e2ba69

emsdk.py
```diff
@@ -2510,13 +2510,24 @@ def adjusted_path(tools_to_activate, system=False):
...
+
+  new_emsdk_tools = []
+  kept_emsdk_tools = []
+  for entry in path_add:
+    if not normalized_contains(existing_emsdk_tools, entry):
+      new_emsdk_tools.append(entry)
+    else:
+      kept_emsdk_tools.append(entry)
+
+  whole_path = unique_items(new_emsdk_tools + kept_emsdk_tools + existing_nonemsdk_path)
```
sbc100, line 2529: Can you revert this part of the fix and it will still work?

Could we move this to its own PR?     I'm not against this part but it seems like a refactor that is independent of the actual fix maybe?

sbc100 (commented):
Great!

scripts/test_path_preservation.ps1
```diff
@@ -0,0 +1,95 @@
...
+        echo $PATH_USER_BEFORE
+        echo "Current machine path is.............."
+        echo $PATH_USER
+        throw "MACHINE PATH are changed while --system was not provided"
+    }
+}
+
+
+
+
```
sbc100, line 83: Remove extra whitespace?   Here and elsewhere in this file.

scripts/test_path_preservation.ps1
```diff
@@ -0,0 +1,95 @@
+refreshenv
```
sbc100, line 1: Maybe add a comment at the top about what this test does?

scripts/test_path_preservation.ps1
```diff
@@ -0,0 +1,95 @@
...
+} finally {
+# Recover pre-split PATH
+refreshenv
+
+[Environment]::SetEnvironmentVariable("Path", $PATH_USER_BEFORE,  "User")
+if ($env:SYSTEM_FLAG) {
+    [Environment]::SetEnvironmentVariable("Path", $PATH_MACHINE_BEFORE,  "Machine")
+}
+
+refreshenv
```
sbc100, line 93: Maybe indent the body of the try{} and finally{} blocks?

scripts/test_permanent.ps1
```diff
@@ -1,56 +1,86 @@
...
 }
+
+
+} finally {
+# Recover pre-split PATH
+refreshenv
+
+[Environment]::SetEnvironmentVariable("Path", $PATH_USER_BEFORE,  "User")
+if ($env:SYSTEM_FLAG) {
+    [Environment]::SetEnvironmentVariable("Path", $PATH_MACHINE_BEFORE,  "Machine")
```
sbc100, line 81: Why not just unconditionally restore both paths here?

scripts/test_permanent.ps1
```diff
@@ -1,56 +1,86 @@
...
-$EMSDK_NODE_USER = [System.Environment]::GetEnvironmentVariable("EMSDK_NODE", "User")
-$EMSDK_PYTHON_USER = [System.Environment]::GetEnvironmentVariable("EMSDK_PYTHON", "User")
-$JAVA_HOME_USER = [System.Environment]::GetEnvironmentVariable("JAVA_HOME", "User")
-$EM_CACHE_USER = [System.Environment]::GetEnvironmentVariable("EM_CACHE", "User")
-$PATH_USER = [System.Environment]::GetEnvironmentVariable("PATH", "User")
+$esc = '--%'
+& "$repo_root/emsdk.ps1" activate latest $esc $env:PERMANENT_FLAG $env:SYSTEM_FLAG
+
+if ($env:SYSTEM_FLAG) {
+    $path_type="Machine"
```
sbc100, line 17: Maybe `path_type` is the wrong name since it applies to all variable?  How about `env_type`?  Or `reg_type`?

sbc100 (commented):
I guess its a separate issue, but I am right there is problem that the shell PATH is supposed to be the combination of path USER and SYSTEM paths?   So by only reading one of them we are not setting the shell's PATH correctly?   If you agree this is an issue I can open separate bug to get that fixed maybe?

emsdk.py
aminya: I would prefer not to. These changes fix the `--permanent` flag, and without this, I am not certain that #634

Additionally, we have already discussed/justified/reviewed the changes here. I do not want to redo all of these just to make these two separate PRs.

scripts/test_permanent.ps1
aminya: Fixed in https://github.com/emscripten-core/emsdk/pull/642/commits/763f44467560f18d153971b3690bf3d5ee18d3c1

scripts/test_permanent.ps1
aminya: Fixed in https://github.com/emscripten-core/emsdk/pull/642/commits/3b00b38c358204f059cf128d6a332d610ceaed54

scripts/test_path_preservation.ps1
aminya: Fixed in https://github.com/emscripten-core/emsdk/pull/642/commits/47debcf0a3aced630870f26a0c0960c1df1dce8e

scripts/test_path_preservation.ps1
aminya: Fixed in https://github.com/emscripten-core/emsdk/pull/642/commits/047b74a3760050c7851ef7da4e05f4b73bd1cf49

scripts/test_path_preservation.ps1
aminya: Fixed in https://github.com/emscripten-core/emsdk/pull/642/commits/47debcf0a3aced630870f26a0c0960c1df1dce8e

emsdk.py
sbc100: The reason I'm asking you to do this is because this change seem separate orthogonal to the actually fix being made.

If you revert this block all the tests, including the new ones will still pass right?   Or am I wrong about that?   If I'm right, I'd really like to see this change land without this block.. and this we can land this block as a separate NFC (Non-functional change) cleanup later.

aminya: It is not orthogonal. That fixes `--permanent`. When I made these changes, my local tests started to work properly. Before that, parts of my PATH were getting removed.

This PR fixes both `--permanent` and `--system`.

sbc100:
You can fix the "Image xcode:9" failure on the mac builder by rebasing (or merging if your prefer).

aminya:
> I guess its a separate issue, but I am right there is problem that the shell PATH is supposed to be the combination of path USER and SYSTEM paths? So by only reading one of them we are not setting the shell's PATH correctly? If you agree this is an issue I can open separate bug to get that fixed maybe?

That's separate. 

Windows updates the shell's PATH automatically when User/Machine PATH is updated. The shell will just need restarting. That's how every other program interacts with Windows, so there is no problem.

The only case which I have not tested yet is when none of `--system` or `--permanent` are provided. The registry is not edited in this case, and only the shell's Path is edited.

emsdk.py
sbc100: I just don't see how the old version of this could ever remove anything from `existing_path`... what am I missed?  Where is the bug in the existing code?

Are you saying that if you revert this block one of the tests will start failing?   If I could see which test would fail it might really help me understand where the bugfix is in this block of code.

I'm not trying to be antagonistic, and I just don't want to land something where I don't understand where the original problem is.

As far as I can tell both these block start out with `existing_path` and include all elements of it that are not part of emsdk... how can the old code drop elements from `existing_path`?

sbc100:
> > I guess its a separate issue, but I am right there is problem that the shell PATH is supposed to be the combination of path USER and SYSTEM paths? So by only reading one of them we are not setting the shell's PATH correctly? If you agree this is an issue I can open separate bug to get that fixed maybe?
> 
> That's separate.
> 
> Windows updates the shell's PATH automatically when User/Machine PATH is updated. The shell will just need restarting. That's how every other program interacts with Windows, so there is no problem.
> 
> The only case which I have not tested yet is when none of `--system` or `--permanent` are provided. The registry is not edited in this case, and only the shell's Path is edited.

I agree its separate, so I will open a new issue for that.   However, our users do expect to be able to use the tools and have an updated `PATH` in the *current* shell, without restarting.   Indeed the user who reported #634 is specifically referring to his current (non-restarted) shell.   They run `emsdk_env.bat` then then run `echo %PATH%` and expect it to work without restarting the shell.   There is no `restart the shell` in our installation instructions.

aminya:
> > > I guess its a separate issue, but I am right there is problem that the shell PATH is supposed to be the combination of path USER and SYSTEM paths? So by only reading one of them we are not setting the shell's PATH correctly? If you agree this is an issue I can open separate bug to get that fixed maybe?
> > 
> > 
> > That's separate.
> > Windows updates the shell's PATH automatically when User/Machine PATH is updated. The shell will just need restarting. That's how every other program interacts with Windows, so there is no problem.
> > The only case which I have not tested yet is when none of `--system` or `--permanent` are provided. The registry is not edited in this case, and only the shell's Path is edited.
> 
> I agree its separate, so I will open a new issue for that. However, our users do expect to be able to use the tools and have an updated `PATH` in the _current_ shell, without restarting. Indeed the user who reported #634 is specifically referring to his current (non-restarted) shell. They run `emsdk_env.bat` then then run `echo %PATH%` and expect it to work without restarting the shell. There is no `restart the shell` in our installation instructions.

That is obvious. But we can document it. Instead of restarting the shell, we can ask them to run `refreshenv.cmd`.

That's similar to running `source ~/.bashrc`

sbc100:
> > There is no `restart the shell` in our installation instructions.
> 
> That is obvious. But we can document it. Instead of restarting the shell, we can ask them to run `refreshenv.cmd`.
> 
> That's similar to running `source ~/.bashrc`

But what about the users who don't run with `--permanent` .. i.e. the default case.    This is the primary use case and is designed to effect the current shell in an additive way.. and not destroy the local PATH.  I think its should be fixable, but lets move further discussion to #645 since its is completely separate I think.

aminya:
Yes. That's separate.

emsdk.py
aminya: That's fine. You have proper write access to my branch.

I have a local backup that finally allows me to spend my time writing the WASM code rather than meticulously proving/disproving a PATH editing algorithm.

aminya:
I added the test for shell/process activation (when no flag is provided). The activation works properly. However, the shell's PATH value is not preserved.

scripts/test_path_preservation.ps1
```diff
@@ -0,0 +1,148 @@
+# This test installs emsdk and activates the latest toolchain using `--system` or `--permanent` flags, and checks if parts of PATH are lost or overwritten. Set $env:SYSTEM_FLAG and $env:PERMANENT_FLAG to test each. If no flag is provided the process/shell values are tested. See the CI file for an example.
```
sbc100, line 1: Can you wrap these this line at 80 columns.

sbc100, line 1: I think maybe there is still a bug here.

If I pass `--permanant` is still don't want any of my local PATH elements removed.

So I think that the `process/shell values` for PATH should be checked in all modes.

So I think when constucting the new PATH for the current shell we should always look in os.environ and not call `win_get_environment_variable` at all.

In other works we need to do two completely different PATH constructions.  Something like this:

```
environ_path = constuct_new_path_for_os_environ()
if permanant:
  registry_path = constuct_new_path_for_registry()
```

Notice that even in the registry mode we still need to construct a separate new path for the current environment.

emsdk.py
```diff
@@ -322,7 +322,10 @@ def win_set_environment_variable_direct(key, value, system=True):
       folder.Close()
 
 
-def win_get_environment_variable(key, system=True, fallback=True):
+def win_get_environment_variable(key, system=True, user=True, fallback=True):
+  if (not system and not user and fallback):
+    # if no --system or --permanent flag is provided use shell's value
+    return os.environ[key]
```
sbc100, line 328: Instead of this change, I think we need call `win_get_environment_variable` only when constructing the path to store in the registry.

When constructing the path for the current environ we can simply not call `win_get_environment_variable` at all.  See my comment below.

sbc100:
This is my proposed change to address this issue: https://github.com/emscripten-core/emsdk/pull/648.    I think its a more precise fix, but I'm open to discussion.   Maybe there other changes here could land on top of it?

scripts/test_path_preservation.ps1
aminya: Fixed in https://github.com/emscripten-core/emsdk/pull/642/commits/9b6acc9ee7cd917b196dd826c845010b34120fff

emsdk.py
aminya: `win_get_environment_variable` is a function for getting environment variables. It has options for making it shell only. When we do not pass `--permanent` or `--system` it becomes shell only. That's how we determine shell only situation.

I don't see any benefit in introducing a new function.

scripts/test_path_preservation.ps1
aminya: > So I think when constucting the new PATH for the current shell we should always look in os.environ and not call win_get_environment_variable at all.

With my current changes, we already do this. Also see https://github.com/emscripten-core/emsdk/pull/642#discussion_r515405800

aminya: > If I pass --permanant is still don't want any of my local PATH elements removed.

The shell's PATH is not changed when `--permanent` is passed. If the user calls `refreshenv` then Windows updates the shell's PATH with the registry's PATH,

sbc100: I don't think that is accurate.   The calculated environment is always applied to the current shell.   On windows this is done by `write_set_env_script` which is called unconditionally. 

The idea is that emsdk is always activated in the current shell.  `--permanant` means that in addition we make the changes in the registry.

aminya: You are trying to circumvent Windows behavior. Registry changes will apply to the shell, no matter what. This is baked in Windows. It may just be deferred depending on the shell that is used. There are some shells that automatically refresh the env changes.

aminya:
Can we merge this before #648? The changes that PR makes are orthogonal to this one, and also, this includes a lot of tests that can prove a new change messes the code.

scripts/test_path_preservation.ps1
sbc100: I'm merely describing the current (and desired) behaviour.  

"Registry changes will apply to the shell, no matter what." -- I'm not sure this is ture.  Registry changes apply to new shells.   The current shell will only be effected if you take specific action to refresh from the registry (and opt into clobbering any local change).

If I run a command prompt and call `set FOO=bar`, I do do not think its true that adding or changing `FOO` in the registry will effect my current shell.

I would suggest we we could simply call `refreshenv` on behalf of the user, but I do not think that is polite because they may have other local changes (e.g. local PATH additions) that they don't want to clobber. 

Please take a look at my proposed change and see what you think?  I am proposing that we construct two environments, one for the current shell and (optionally) one for the registry which is different.   I believe this is the behaviour that our users expect.

sbc100:
I think the tests you have added here are very useful.  How you would feel about landing #648 first and then added your new tests on top of it (along with any fixed here that are needed to make them pass)?

scripts/test_path_preservation.ps1
aminya: > "Registry changes will apply to the shell, no matter what." -- I'm not sure this is ture. Registry changes apply to new shells. The current shell will only be effected if you take specific action to refresh from the registry (and opt into clobbering any local changes).

Well, there are some shells that watch the registry changes, and they call refreshenv.cmd in case of a change. This allows a long life shell. This technique is used in servers. 

> I would suggest we we could simply call refreshenv on behalf of the user, but I do not think that is polite because they may have other local changes (e.g. local PATH additions) that they don't want to clobber.

No. We should not. As I said, we should not circumvent Windows behavior.

> Please take a look at my proposed change and see what you think? I am proposing that we construct two environments, one for the current shell and (optionally) one for the registry which is different. I believe this is the behaviour that our users expect.

If you can merge this PR, my tests will show if those changes actually fix the problem. Once you merge this, you can add process test by pasting this code at [here](https://github.com/aminya/emsdk/blob/9b6acc9ee7cd917b196dd826c845010b34120fff/scripts/test_path_preservation.ps1#L81) and [here](https://github.com/aminya/emsdk/blob/9b6acc9ee7cd917b196dd826c845010b34120fff/scripts/test_path_preservation.ps1#L90)
```ps1
        if (Compare-Object -ReferenceObject $PATH_Process_BEFORE.Split(';') -DifferenceObject $PATH_Process.Split(';') ) {
            echo "Old process path is ...................."
            echo $PATH_Process_BEFORE
            echo "Current process path is ................"
            echo $PATH_Process
            throw "User PATH are changed while --system had been provided"
        }
```

I prefer a test-driven approach considering that this codebase is not small.

aminya:
I am afraid that #648 introduces conflicts. This PR has been open for 17 days and fixes "more severe problems" (keeping User/Machine/Process environment variables when respective flags are passed). My Windows has been messed up a couple of times when I was developing this PR (I had to use system restore), and now, it finally, works.

Your PR is trying to fix an edge case (keeping process changes when permanent is passed).

sbc100:
> I am afraid that #648 introduces conflicts. This PR has been open for 17 days and fixes "more severe problems" (keeping User/Machine/Process environment variables when respective flags are passed). My Windows has been messed up a couple of times when I was developing this PR (I had to use system restore), and now, it finally, works.
> 
> Your PR is trying to fix an edge case (keeping process changes when permanent is passed).

Indeed it is trying to solve that problem.    I worry that this change *does* modify current process when `--permanant` is passed because it calle `write_set_env_script` (which effects the current process) with the wrong path (specifically either the system or the user path, not the concatenation of the two).

So I think this change will break the current process path when run with `--permanent`.  Do you agree?

aminya:
> > I am afraid that #648 introduces conflicts. This PR has been open for 17 days and fixes "more severe problems" (keeping User/Machine/Process environment variables when respective flags are passed). My Windows has been messed up a couple of times when I was developing this PR (I had to use system restore), and now, it finally, works.
> > Your PR is trying to fix an edge case (keeping process changes when permanent is passed).
> 
> Indeed it is trying to solve that problem. I worry that this change _does_ modify current process when `--permanant` is passed because it calls `write_set_env_script` (which effects the current process) with the wrong path (specifically either the system or the user path, not the concatenation of the two).
> 
> So I think this change will break the current process path when run with `--permanent`. Do you agree?

I have not tested that edge case, but it is possible. 

However, as I said, there are more critical bugs on the master that this PR is fixing. Currently, the master branch destroys the Windows registry of anyone who runs `activate` with `--permanent` and `--system`. Not everyone has a backup of their registry and this damage may not be recoverable. To me, this is so critical that we should not wait anymore. This puts this PR as a high priority.

emsdk.py
```diff
@@ -2564,8 +2566,8 @@ def get_env_vars_to_add(tools_to_activate):
   return env_vars_to_add
 
 
-def construct_env(tools_to_activate):
-  return construct_env_with_vars(get_env_vars_to_add(tools_to_activate))
+def construct_env(tools_to_activate, system, user):
```
sbc100, line 2569: How about using `persistent` and `system` as the names for these args everywhere to match the names that we use on the command line?  (since they map 1-to-1).

emsdk.py
```diff
@@ -355,21 +358,9 @@ def win_get_environment_variable(key, system=True, fallback=True):
...
-
-def win_get_active_environment_variable(key):
-  value = win_get_environment_variable(key, False)
-  if value is not None:
-    return value
-  return win_get_environment_variable(key, True)
-
-
-def win_set_environment_variable(key, value, system=True):
+def win_set_environment_variable(key, value, system=True, user=True):
```
sbc100, line 361: Maybe remove the default parameters here since we actually want to require all these argument everywhere now right?

emsdk.py
```diff
@@ -443,9 +434,9 @@ def win_set_environment_variables(env_vars_to_add, system):
     errlog('SendMessageTimeout failed with error: ' + str(e))
 
 
-def win_delete_environment_variable(key, system=True):
+def win_delete_environment_variable(key, system=True, user=True):
```
sbc100, line 437: Ditto remove the defaults.

sbc100 (approved):
OK, since there are tests added I think we can move forward and land this.  Thanks for taking the time to go back and forth with me.

I'm afraid I still don't really understand how the registry corruption you are talking about happens but the tests ultimately prove that is not happening.

I've made just a couple more suggestions but otherwise I'm ready land.

sbc100:
I think the mac test just requires you to rebase or merge and it will go away.

emsdk.py
aminya: I tried to use `persistent` only on the top-level function. From the Windows point of view, persistent means `user`, so using that is a clearer abstraction. I did this so we can safely move these functions to other files/libraries later. I kept `arg_persistent` however on the top-level function so it matches the name of the flag.

emsdk.py
aminya: This function is not used anywhere, so we can remove the whole thing. I was planning to do a clean up after this PR.

emsdk.py
aminya: Fixed in https://github.com/emscripten-core/emsdk/pull/642/commits/e7d8a37db14f089451adaa23f3006db704d861d5

aminya:
> I think the mac test just requires you to rebase or merge and it will go away.

Rebased!

PR #644, opened 2020-10-24
Update mac circleci environment

sbc100:
Apparently `Image xcode:9.0 is not supported` these days.

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -11,11 +11,17 @@ if [ -z "$EMSDK_PYTHON" ]; then
   PYTHON3=$(dirname $0)/python/3.7.4-2_64bit/bin/python3
   if [ -e $PYTHON3 ]; then
     EMSDK_PYTHON=$PYTHON3
+
     # When using our bundled python we never want the users
     # PYTHONHOME or PYTHONPATH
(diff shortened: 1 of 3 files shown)


sbc100:
Landed now to green the CI

emsdk
```diff
@@ -11,11 +11,17 @@ if [ -z "$EMSDK_PYTHON" ]; then
...
     # When using our bundled python we never want the users
     # PYTHONHOME or PYTHONPATH
     # https://github.com/emscripten-core/emsdk/issues/598
     unset PYTHONHOME
     unset PYTHONPATH
+
+    # This is needed for MacOS.  Without this, the urlopen
+    # code will try to use /usr/local/etc/openssl/cert.pem
+    # which may or may not exist on the system.
+    export SSL_CERT_FILE=$(dirname $0)/python/3.7.4-2_64bit/lib/python3.7/site-packages/certifi/cacert.pem
```
kripken, line 24: Why is this needed in two places?

sbc100: Its needed here so that emsdk works correctly even the sdk is not active in the current env.  This block of code here runs when EMSDK_PYTHON is *not* set... which means that sdk has not activated in the current env.

PR #647, opened 2020-10-28
Fix regression where reinstalling tools from git sources would not git pull and rebuild them.

juj:


emsdk.py
```diff
@@ -1871,7 +1871,11 @@ def install_sdk(self):
     return True
 
   def install_tool(self):
-    if self.is_installed():
+    # Avoid doing a redundant reinstall of the tool, if it has already been installed.
+    # However all tools that are sourced directly from git branches do need to be
+    # installed every time when requested, since the install step is then used to git
+    # pull the tool to a newer version.
+    if self.is_installed() and not hasattr(self, 'git_branch'):
```
sbc100, line 1878: Probably not worth it but you should imagine the git sha being embedded in the installation stamp file so it would re-build only when the sha changed?

sbc100 (approved):

emsdk.py
juj: The build will be incremental, so if nothing actually changed, then the build will not do much anything either.

PR #648, opened 2020-10-30
On windows, construct environment separately when storing to registry

### LINKED ISSUE #634: On Windows emsdk_env.bat removes most of the PATH environment variable

sbc100:
We can't use the same set of environment variables when setting local
values and when storing to the windows registry.  

This is because the new PATH value is constructed very differently
when updating the local shell compared to when updating the
registry.

Add test that verifies that no path elements are removed by
emsdk_env.  This includes not re-ordering or removing duplicates
that might exist already in the user's path.

Fixes: #634

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -298,7 +298,7 @@ def remove_tree(d):
     debug_print('remove_tree threw an exception, ignoring: ' + str(e))
 
 
-def win_set_environment_variable_direct(key, value, system=True):
+def win_set_environment_variable_direct(key, value, system):
   folder = None
   try:
     if system:
@@ -322,10 +322,7 @@ def win_set_environment_variable_direct(key, value, system=True):
       folder.Close()
 
 
-def win_get_environment_variable(key, system=True, user=True, fallback=True):
-  if (not system and not user and fallback):
-    # if no --system or --permanent flag is provided use shell's value
-    return os.environ[key]
+def win_get_environment_variable(key, system):
   try:
     folder = None
     try:
@@ -337,12 +334,6 @@ def win_get_environment_variable(key, system=True, user=True, fallback=True):
         folder = winreg.OpenKey(winreg.HKEY_CURRENT_USER, 'Environment')
       value = str(winreg.QueryValueEx(folder, key)[0])
     except Exception:
-      # If reading registry fails for some reason - read via os.environ. This has the drawback
-      # that expansion items such as %PROGRAMFILES% will have been expanded, so
-      # need to be precise not to set these back to system registry, or
-      # expansion items would be lost.
-      if fallback:
-        return os.environ[key]
       return None
     finally:
       if folder is not None:
@@ -358,9 +349,9 @@ def win_get_environment_variable(key, system=True, user=True, fallback=True):
   return value
 
 
-def win_set_environment_variable(key, value, system, user):
+def win_set_environment_variable(key, value, system):
   debug_print('set ' + str(key) + '=' + str(value) + ', in system=' + str(system), file=sys.stderr)
-  previous_value = win_get_environment_variable(key, system=system, user=user)
+  previous_value = win_get_environment_variable(key, system=system)
   if previous_value == value:
     debug_print('  no need to set, since same value already exists.')
     # No need to elevate UAC for nothing to set the same value, skip.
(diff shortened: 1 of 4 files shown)


aminya (changes requested):
See https://github.com/emscripten-core/emsdk/pull/642#discussion_r515405800. That PR considers all the cases (`--system`, `--permamnet`, `none of them`). It includes the tests for all of them.

aminya (changes requested):
I am interested to know if this fixes the Process PATH preservation.

You can add process test by pasting this code at [here](https://github.com/aminya/emsdk/blob/9b6acc9ee7cd917b196dd826c845010b34120fff/scripts/test_path_preservation.ps1#L81) and [here](https://github.com/aminya/emsdk/blob/9b6acc9ee7cd917b196dd826c845010b34120fff/scripts/test_path_preservation.ps1#L90)
```ps1
        if (Compare-Object -ReferenceObject $PATH_Process_BEFORE.Split(';') -DifferenceObject $PATH_Process.Split(';') ) {
            echo "Old process path is ...................."
            echo $PATH_Process_BEFORE
            echo "Current process path is ................"
            echo $PATH_Process
            throw "User PATH are changed while --system had been provided"
        }
```

sbc100:
> I am interested to know if this fixes the Process PATH preservation.
> 
> You can add process test by pasting this code at [here](https://github.com/aminya/emsdk/blob/9b6acc9ee7cd917b196dd826c845010b34120fff/scripts/test_path_preservation.ps1#L81) and [here](https://github.com/aminya/emsdk/blob/9b6acc9ee7cd917b196dd826c845010b34120fff/scripts/test_path_preservation.ps1#L90)
> 
> ```powershell
>         if (Compare-Object -ReferenceObject $PATH_Process_BEFORE.Split(';') -DifferenceObject $PATH_Process.Split(';') ) {
>             echo "Old process path is ...................."
>             echo $PATH_Process_BEFORE
>             echo "Current process path is ................"
>             echo $PATH_Process
>             throw "User PATH are changed while --system had been provided"
>         }
> ```

I don't think that will work because activation *is* supposed to add to the process path so it will be different afterwards.  

The test that I've added added in this RR in `scripts/check_path.py` instread checked that all of the element of the original path still appear in the new path and that none were removed.  i.e. that the operation only added to the process path and did not remove or re-order the existing process path.

PR #665, opened 2020-12-07
Add Git-LFS package to EMSDK Docker image

haraldreingruber:
I could imagine that using Git-LFS together with emscripten is not that uncommon, when asset files are pre-loaded.

This PR adds the git-lfs apt package to the docker image. This way, git-lfs doesn't have to be installed during each run in a Github Action build or similar.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -101,6 +101,7 @@ RUN echo "## Update and install packages" \
         zip \
         unzip \
         git \
+        git-lfs \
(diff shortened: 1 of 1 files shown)


sbc100:
Since this is a PR rather than a bug report can you update the title and description such that is describes the change?

haraldreingruber:
Good remark. I've updated the title and description.
Let me know if you have any improvement ideas.

Thanks!

PR #682, opened 2021-01-19
fix: \`file\` command output without mimetype on MacOS

levenon:
fix: bazel/emscripten_toolchain/wasm_binary.py  \`file\` command output without mimetype on MacOS

### DIFF
--- a/bazel/emscripten_toolchain/emscripten_config
+++ b/bazel/emscripten_toolchain/emscripten_config
@@ -1,9 +1,13 @@
 import os
+import platform
 
 ROOT_DIR = os.environ["ROOT_DIR"]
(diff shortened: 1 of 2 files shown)


levenon:
Linux: 

file -Lib wasm_tar_file

the output is `application/x-tar; charset=binary`

But MacOS:

file -Lib wasm_tar_file

the output is `regular file`

I think the command line tool `file` is different 

so I changed the options to use for compatibility between two systems

file -b --mime-type --mime-encoding wasm_tar_file

the output of both systems is same

levenon:
@walkingeyerobot it works by adding -L on macOS,

walkingeyerobot (approved):

PR #689, opened 2021-01-25
2.0.13

kripken:


kripken:
@walkingeyerobot @sbc100 It looks like our bazel integration has broken with the new sysroot changes?

sbc100:
The linux build is failing due to binaryen cmake requirements being increased.   I can take a look at both issues.

PR #690, opened 2021-01-25
WIP: Bazel use sysroot

sbc100:


sbc100:
Hopefully @walkingeyerobot you can figure out what I'm going wrong here?

I also wonder if we can make this one bazel file work with older and newer versions of emscripten?  (perhaps by detecting the precise of `cache/sysroot` directory?)

bazel/WORKSPACE
```diff
@@ -8,12 +8,12 @@ http_archive(
...
 http_archive(
     name = "emscripten",
-    sha256 = "bfff9fb0326363c12e19b542f27a5f12cedbfc310f30621dc497c9af51d2d2e3",
+    sha256 = "a0602a8266864664e5a932016e50b11d64491a4b766e01011339f8f317d5eaed",
     strip_prefix = "install",
-    url = "https://storage.googleapis.com/webassembly/emscripten-releases-builds/linux/dcf819a7821f8db0c8f15ac336fea8960ec204f5/wasm-binaries.tbz2",
+    url = "https://storage.googleapis.com/webassembly/emscripten-releases-builds/linux/425903692d0ecac3420bdf36eb4d2ca144dfdfeb/wasm-binaries.tbz2",
     build_file = "//emscripten_toolchain:emscripten.BUILD",
     type = "tar.bz2",
 )
```
attilaolah, line 19: Since we're here: it would be nice if this http_archive was loaded from an external .bzl file, something like:

`WORKSPACE`:
```bzl
load("package.bzl", "install_dependencies")
istall_dependencies()
```

`package.bzl`:
```
def install_dependencies():
    http_archive(…)
```

That way this workspace could be used as an external workspace without having to copy these rules. (It is possible today to use `@emsdk` as an external repo, but we need to install `@emscripten` manually then, and update every time. It would be nice if `@emsdk` contained the repo definition of `@emscripten` in a `.bzl` file so it could be imported into another workspace.)

sbc100: Sounds reasonable.  Would you mind opening a separate issue?

attilaolah: Sure, I created #696. I might give it a stab this weekend if I can find the time.

PR #691, opened 2021-01-26
Update CircleCI to more modern docker image

sbc100:
See https://circleci.com/developer/images/image/cimg/python

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,8 @@ orbs:
 executors:
   bionic:
     docker:
-      - image: buildpack-deps:bionic
+      # 3.6 is the current minimum requirement for emscripten
+      - image: cimg/python:3.6
 
 jobs:
   flake8:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -170,20 +165,18 @@ jobs:
...
       - run: curl -fsSL https://bazel.build/bazel-release.pub.gpg | gpg --dearmor > bazel.gpg
-      - run: mv bazel.gpg /etc/apt/trusted.gpg.d/
-      - run: echo "deb [arch=amd64] https://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
-      - run:
-          name: install pip
-          command: |
-            apt-get update -q
-            apt-get install -q -y python3-pip
+      - run: sudo mv bazel.gpg /etc/apt/trusted.gpg.d/
+      - run: echo "deb [arch=amd64] https://storage.googleapis.com/bazel-apt stable jdk1.8" | sudo tee /etc/apt/sources.list.d/bazel.list
```
kripken, line 173: `sudo` seems odd here... but if it works...

kripken (approved):

.circleci/config.yml
sbc100: This is installing a new system-wide settings so it makes sense.

In general I think its better that the admin commands require sudo.. since it means that other are not running as root but a normal user.

PR #693, opened 2021-01-28
Set LLVM's git branch to `main` in manifest.

### LINKED ISSUE #692: LLVM fails to build using `./emsdk install sdk-upstream-master-64bit`
Similar to #683, LLVM just renamed their default branch from `master` to `main`.
This causes a `git checkout master` step to fail.

Matthew-McRaven:
LLVM renamed their default branch to main from master.
This breaks builds of the SDK from the sources.

This commit retargets LLVM builds to use the git branch main when appropriate.
The version name "master" remains the same to maintain compatibility with existing build scripts

Fixes #692.


sbc100 (approved):

sbc100:
Thanks!

Matthew-McRaven:
I verified that the patch worked by building from the sources in a fresh container, and this path resolves the error I received.

Matthew-McRaven:
Thank you!!

PR #694, opened 2021-01-29
Release 2.0.13 and update bazel toolchain

sbc100:
This change comes with fairly major change to the bazel toolchain.

- Use pre-built cache that comes with emsdk
- Mark cache as readonly using FROZEN_CACHE
- Pass `--sysroot` to match upstream emscripten change

### DIFF
--- a/bazel/emscripten_toolchain/crosstool.bzl
+++ b/bazel/emscripten_toolchain/crosstool.bzl
@@ -70,9 +70,8 @@ def _impl(ctx):
 
     cc_target_os = "emscripten"
     emscripten_version = ctx.attr.emscripten_version
-    emscripten_root = "external/emscripten/" + emscripten_version
 
-    builtin_sysroot = None
+    builtin_sysroot = "external/emscripten/emscripten/cache/sysroot"
 
     ################################################################
     # Tools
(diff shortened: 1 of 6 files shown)


bazel/emscripten_toolchain/crosstool.bzl
```diff
@@ -1070,18 +1043,10 @@ def _impl(ctx):
...
-        emscripten_version + "/system/lib/compiler-rt/include",
-        emscripten_version + "/system/lib/libcxxabi/include",
-        emscripten_version + "/system/include/libc",
-        emscripten_version + "/system/include/gfx",
-        emscripten_version + "/system/include/SDL",
-        emscripten_version + "/lib/clang/12.0.0/include",
+        "external/emscripten/lib/clang/12.0.0/include",
+        "external/emscripten/emscripten/cache/sysroot/include",
+        "external/emscripten/emscripten/cache/sysroot/include/c++/v1",
+        "external/emscripten/emscripten/cache/sysroot/include/compat",
```
trybka, line 1049: what about .../sysroot/local/include (to match above) ? 

It would be nice if they were consistently ordered, but that might be a bit onerous.

walkingeyerobot (approved):
Awesome, hope this wasn't too much of a pain to figure out.

kripken (commented):
lgtm aside from that there is an emscripten roll ongoing, so for tagging we need to wait and then use a later hash.

bazel/emscripten_toolchain/crosstool.bzl
sbc100: Actually I think local should not be needed in either case.  We use local to allow folks to install their own headers.. but in this case there is no way to write the sysroot anyway.

walkingeyerobot (approved):

kripken (approved):

PR #698, opened 2021-02-07
Fix search of VCTargetsPath for VS2019 build.

juj:
This is a somewhat cryptic change about VS2017 -> VS2019, I am struggling to find information about this on the web.

It looks like a change in Windows SDK or VS2017 -> VS2019 has shuffled paths around the toolchains, hence what used to be `VCTargetsPath\Common7\IDE\VC\VCTargets` can now be located in `VCTargetsPath\MSBuild\Microsoft\VC\v160\`.

Also not quite sure I understand the role of the `VCTargetsPath` environment variable. The intended role of `VCTargetsPath` is something I don't understand properly, and it is not really well documented.

My impression of this situation is that the `VCTargetsPath` environment variable would have been something that Microsoft Visual Studio has once had as a property that's needed, and CMake was built to rely on it. But as newer Visual Studios came around, `VCTargetsPath` has been demoted to an internal VS project/solution property, and not a full environment variable. CMake however never changed, and it still expects to see `VCTargetsPath` when targeting Visual Studio.

The above is a complete guess however, I am struggling to find information about this on the web.

Current official VS 2019 Command Prompt does not have that environment variable. Neither does VS 2017 Command Prompt. It is something that CMake expects to see when compiling using the Visual Studio generator.

In https://discourse.cmake.org/t/cmake-failed-to-get-the-value-of-vctargetspath-with-vs2019-16-7/1839/16 Brad King, a core CMake developer suggested that `If the imported file is missing, that means Visual Studio’s own VCTargetsPath is incorrect. This likely indicates a problem with its installation.`. However the core issue is that Visual Studio does no longer has that env. var in 2017 or 2019, so not sure what to make of it.

In any case, this change fixes VS2019 build for me. In the future it is possible that we need to manually update the directory search list to `Microsoft.Cpp.Default.props` as new toolchain versions come out; unless someone knows how this stuff is supposed to work in CMake and/or VS and can offer an automated search feature :)

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -903,7 +903,29 @@ def build_env(generator):
       path = vswhere(16)
     else:
       path = vswhere(15)
-    build_env['VCTargetsPath'] = os.path.join(path, 'Common7\\IDE\\VC\\VCTargets')
+
+    # Configuring CMake for Visual Studio needs and env. var VCTargetsPath to be present.
+    # How this is supposed to work is unfortunately very undocumented. See
+    # https://discourse.cmake.org/t/cmake-failed-to-get-the-value-of-vctargetspath-with-vs2019-16-7/1839/16
+    # for some conversation. Try a couple of common paths if one of them would work.
+    # In the future as new versions of VS come out, we likely need to add new paths into this list.
+    if 'VCTargetsPath' not in build_env:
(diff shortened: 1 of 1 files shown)


juj:
Looks like no objections, landing..

Issue #705, opened 2021-02-12
Make adding node and python to PATH optional


sbc100:
Neither emsdk nor emscripten requires that these be in the PATH.

We add them to the path anyway since it is convenient for some users to be able to to do:

```
$ emcc hello.c
$ node a.out.js
```

However, by adding our own python and node builds to the users path we are potentially screwing with any existing python and/node installations they might have.   See #704 for an example of how our internal version of python can leak out in unexpected ways.

Proposed solution for python:

Lets just stop adding to the PATH.  All our tools launch using .bat/sh files that honor EMSDK_PYTHON so we don't need it in the path.  Also we don't want users who run `pip` to interfere with our internal version of python.

Proposed solutions(s) for node:

Option 1:  Make adding node to the PATH optional.  Perhaps on by default so users who have their own node would have to disable with it with `--no-node-path`?

Option 2: Add our version of node the PATH if and only if no version of node already exists there.

Option 2: Provide some kind of `emrun` or `emnode` script that will run under the configured version of node *without* having to place it in the PATH. 

I suggest we remove python from the PATH ASAP since there is no reason that I know of not to.

kripken:
Sgtm about python.

For node, I think all internal usage of node (running compiler.js, the JS optimizer, etc.) should look at `NODE_JS` in `.emscripten`. So this would only affect users running `node example.js` etc. If that's true, then option 2 sounds good to me.

sbc100:
Closing for now until we get folks that really care about getting their node installation covered by ours when using emsdk.

PR #706, opened 2021-02-12
Avoid adding our python and java to user's PATH

sbc100:
For Java we use the value in the config file so it doesn't need to be in
the PATH.

For python, all our tools should launch via scripts that check for
EMSDK_PYTHON so having python in the PATH for emsdk users should not be
needed.

The motivation for this is that we don't want to clobber any existing
python or java versions that users might already have in their PATH.

This is the easy part of #705.

### DIFF
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -232,7 +232,6 @@
     "bitness": 32,
     "arch": "x86",
     "windows_url": "WinPython-32bit-2.7.13.1Zero.zip",
-    "activated_path": "%installation_dir%/python-2.7.13",
     "activated_cfg": "PYTHON='%installation_dir%/python-2.7.13/python%.exe%'",
     "activated_env": "EMSDK_PYTHON=%installation_dir%/python-2.7.13/python%.exe%"
   },
@@ -242,7 +241,6 @@
     "bitness": 64,
     "arch": "x86_64",
     "windows_url": "WinPython-64bit-2.7.13.1Zero.zip",
-    "activated_path": "%installation_dir%/python-2.7.13.amd64",
     "activated_cfg": "PYTHON='%installation_dir%/python-2.7.13.amd64/python%.exe%'",
     "activated_env": "EMSDK_PYTHON=%installation_dir%/python-2.7.13.amd64/python%.exe%"
   },
@@ -252,7 +250,6 @@
     "bitness": 32,
     "arch": "x86",
     "windows_url": "python-3.7.4-embed-win32-patched.zip",
-    "activated_path": "%installation_dir%",
     "activated_cfg": "PYTHON='%installation_dir%/python.exe'",
     "activated_env": "EMSDK_PYTHON=%installation_dir%/python.exe"
   },
@@ -262,7 +259,6 @@
     "bitness": 64,
     "arch": "x86_64",
     "windows_url": "python-3.7.4-embed-amd64-patched.zip",
-    "activated_path": "%installation_dir%",
     "activated_cfg": "PYTHON='%installation_dir%/python.exe'",
     "activated_env": "EMSDK_PYTHON=%installation_dir%/python.exe"
   },
@@ -272,7 +268,6 @@
     "bitness": 32,
     "arch": "x86",
     "windows_url": "python-3.7.4-embed-win32+pywin32.zip",
-    "activated_path": "%installation_dir%",
     "activated_cfg": "PYTHON='%installation_dir%/python.exe'",
     "activated_env": "EMSDK_PYTHON=%installation_dir%/python.exe"
   },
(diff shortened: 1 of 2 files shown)


kripken (approved):

juj:
This does have the drawback that on Windows, the workflow
```
emsdk install latest
emsdk activate latest
cd %EMSCRIPTEN%
python tests/runner.py ...
```
will no longer work, because python is not in PATH. That syntax is documented in https://emscripten.org/docs/getting_started/test-suite.html#running-tests . I use that workflow daily.

One will need to pass `%EMSDK_PYTHON% tests/runner.py` instead. Or first install a system python separately out of emsdk, but then the tests won't be run using the same python interpreter as the compiler is.

Agree with removing java from PATH, there are no workflows that would use that. But maybe we should keep `python` in PATH, and resolve taking Python and Node away from PATH via some other means for users that want it? Not sure what the best solution here would be...

sbc100:
Perhaps we can just add runner.py to the list of python launchers?

As a side note, do we want emsdk users to be able to run the test suite?   It seems at least like an unusal use case.  Most folks running the test suite I imagine will be using a git checkout won't they?     We already strip the bulky parts of the test suite to save space for emsdk users (specifically tests/third_party) and we don't install the node dev dependencies so such users already have to take extra steps to run the test suite.

juj:
> As a side note, do we want emsdk users to be able to run the test suite? It seems at least like an unusal use case. Most folks running the test suite I imagine will be using a git checkout won't they? We already strip the bulky parts of the test suite to save space for emsdk users (specifically tests/third_party) and we don't install the node dev dependencies so such users already have to take extra steps to run the test suite.

I use emsdk and use the test suite. I don't think it should be hard for people to both use emsdk and contribute to emscripten? If we asked contributors to use a different path, then that would be kind of not dogfooding our own product, making it a second tier platform/workflow?

sbc100:
> > As a side note, do we want emsdk users to be able to run the test suite? It seems at least like an unusal use case. Most folks running the test suite I imagine will be using a git checkout won't they? We already strip the bulky parts of the test suite to save space for emsdk users (specifically tests/third_party) and we don't install the node dev dependencies so such users already have to take extra steps to run the test suite.
> 
> I use emsdk and use the test suite. I don't think it should be hard for people to both use emsdk and contribute to emscripten? If we asked contributors to use a different path, then that would be kind of not dogfooding our own product, making it a second tier platform/workflow?

This change should address this concern: https://github.com/emscripten-core/emscripten/pull/13494

I agree it should be easy to use all the emsdk components when contributing to emscripten... aside from emscripten itself right? 
 Contributors would all need to checkout emscripten itself via git right?

juj:
> Contributors would all need to checkout emscripten itself via git right?

Yes, via git, but one can still use emsdk. Emsdk offers the now named `sdk-upstream-master-64bit` as a git checkout. I have used that method since the start of developing Emsdk+Emscripten in 2013. E.g.

```
git clone https://github.com/emscripten-core/emsdk.git
cd emsdk
emsdk install sdk-upstream-master-64bit binaryen-master-64bit
emsdk activate sdk-upstream-master-64bit binaryen-master-64bit
emsdk_env

# If I want to contribute to Emscripten repo:
pushd emscripten\master
git remote add juj https://github.com/juj/emscripten.git
git fetch juj
popd

# And/or if I want to contribute to LLVM repo:
pushd llvm\git\src
git remote add juj https://github.com/juj/llvm-project.git
git fetch juj
popd

# And/or if I want to contribute to Binaryen repo:
pushd binaryen\master
git remote add juj https://github.com/juj/emscripten.git
git fetch juj
popd

```
and I'm good to go. I think that is the quickest workflow for anyone contributing to Emscripten, since one does not then need to manually set up LLVM git clones, emscripten configs or other tools, and same workflow works whether you are contributing from Windows, Linux or macOS.

sbc100:
> https://github.com/juj/emscripten.git

Great that makes sense.   I just wanted to make sure that you were not expecting the full emscripten test suite to be runnable from the "packaged" as opposed to "cloned" version of emscripten.  We try to save space in the packaged version so it doesn't come with everything one needs to run the full test suite.

For reference, I do a similar thing to you although my workflow is. 

1.  download, install and activate emsdk.
2.  `git clone` emscripten anywhere I like. 
3.  Use the cloned version of emscripten when I want to by putting in first in PATH (and everything else automatically comes from emsdk).

sbc100:
Now that emscripten-core/emscripten#13494 is landed can we land this?    If its still controversial I can split the java and python changes into separate PRs.

juj:
> Great that makes sense. I just wanted to make sure that you were not expecting the full emscripten test suite to be runnable from the "packaged" as opposed to "cloned" version of emscripten. We try to save space in the packaged version so it doesn't come with everything one needs to run the full test suite.

Agree, in an end user zipped version, slimming down the test suite is a good idea.

This PR lgtm to land. Nice work overall!

PR #708, opened 2021-02-12
Update node version to 14.15.5 (latest TLS release)

sbc100:
See https://github.com/WebAssembly/binaryen/issues/3551

### DIFF
--- a/scripts/update_node.py
+++ b/scripts/update_node.py
@@ -16,8 +16,8 @@ import subprocess
 import os
 import shutil
 
-version = '12.18.1'
-base = 'https://nodejs.org/dist/latest-v12.x/'
+version = '14.15.5'
+base = 'https://nodejs.org/dist/latest-v14.x/'
 upload_base = 'gs://webassembly/emscripten-releases-builds/deps/'
 
 suffixes = [
@@ -40,6 +40,7 @@ for suffix in suffixes:
       shutil.move(dirname, 'bin')
       os.mkdir(dirname)
       shutil.move('bin', dirname)
+      os.remove(filename)
       subprocess.check_call(['zip', '-rq', filename, dirname])
       shutil.rmtree(dirname)
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -66,8 +66,8 @@ COPY --from=stage_build /emsdk /emsdk
 # (sub-stages) or with custom / no entrypoint
 ENV EMSDK=/emsdk \
     EM_CONFIG=/emsdk/.emscripten \
-    EMSDK_NODE=/emsdk/node/12.18.1_64bit/bin/node \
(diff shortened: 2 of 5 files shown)


emsdk.py
```diff
@@ -1732,6 +1732,11 @@ def each_path_exists(pathlist):
       return all(os.path.exists(p) for p in pathlist)
 
     content_exists = os.path.exists(self.installation_path()) and each_path_exists(activated_path) and (os.path.isfile(self.installation_path()) or num_files_in_directory(self.installation_path()) > 0)
+    print(self.installation_path())
+    print(os.path.exists(self.installation_path()))
+    for p in activated_path:
+      print(p)
+      print(os.path.exists(p))
```
kripken, line 1739: looks like debug code?

scripts/update_node.py
```diff
@@ -40,6 +40,7 @@
       shutil.move(dirname, 'bin')
       os.mkdir(dirname)
       shutil.move('bin', dirname)
+      os.remove(filename)
```
kripken, line 43: doesn't this break the test?

kripken (changes requested):
lgtm aside from comments

emsdk.py
sbc100: Yup.. will remove before landing.

emsdk.py
```diff
@@ -1732,6 +1732,11 @@ def each_path_exists(pathlist):
       return all(os.path.exists(p) for p in pathlist)
 
     content_exists = os.path.exists(self.installation_path()) and each_path_exists(activated_path) and (os.path.isfile(self.installation_path()) or num_files_in_directory(self.installation_path()) > 0)
+    errlog(self.installation_path())
+    errlog(os.path.exists(self.installation_path()))
+    for p in activated_path:
+      errlog(p)
+      errlog(os.path.exists(p))
```
kripken, line 1739: looks like `errlog` prints to stderr - that still seems excessive here?

scripts/update_node.py
sbc100: This fixed a bug where we were including two copies of node in windows.  We were calling zip with the existing zip file present which was adding the new "bin" directory, but not removing the old zip contents.   So it saves a bit of space for all windows users.

emsdk.py
sbc100: Sorry, I'm just debugging the windows failure.

kripken (approved):
lgtm (with logging removed and assuming the windows fix does not change anything)

sbc100:
This test starts to fail with the upgrade:  test_node_code_caching

@kripken  can you verify that this test fails with recent versions of node ?   Is this a known issue?

kripken:
@sbc100 Oh, yes - this is a known issue, node changed/removed the API we depended on, https://github.com/nodejs/node/issues/18265#issuecomment-622971547 . There is a note on this in settings.js.

We can just disable the test, I guess. I hope node adds a new API at some point.

sbc100:
Ok I will land this and roll a PR to disable the test.

PR #714, opened 2021-02-15
Only add node/bin to user's PATH if one is not already found

sbc100:
If the user already has a version of node in their PATH don't clobber
it.  This doesn't effect emscripten since the version of node we use
there is controlled via the config file, not via PATH.

Part of fix for #705.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1431,6 +1431,8 @@ def get_required_path(active_tools):
   path_add = [to_native_path(emsdk_path())]
   for tool in active_tools:
     if hasattr(tool, 'activated_path'):
+      if hasattr(tool, 'activated_path_skip') and which(tool.activated_path_skip):
+        continue
       path = to_native_path(tool.expand_vars(tool.activated_path))
       path_add.append(path)
   return path_add
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -191,6 +191,7 @@
     "arch": "x86",
     "windows_url": "node-v14.15.5-win-x86.zip",
     "activated_path": "%installation_dir%/bin",
+    "activated_path_skip": "node",
     "activated_cfg": "NODE_JS='%installation_dir%/bin/node%.exe%'",
     "activated_env": "EMSDK_NODE=%installation_dir%/bin/node%.exe%"
   },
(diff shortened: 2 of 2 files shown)


juj:
I don't think this kind of conditional behavior is a good idea, it can lead to cryptic corner cases.

Also when people enter the emsdk environment with emsdk_env, I don't think it is a problem to activate the emsdk tools in it. The user did in fact enter the emsdk environment! (With python the root issue was that our bundled python has broken pip, not that we activated python in PATH)

If users don't want to get node, they could emsdk activate the individual tools that make up the sdk, and leave out node from it. Then emsdk won't add node in it?

sbc100:
> I don't think this kind of conditional behavior is a good idea, it can lead to cryptic corner cases.

Sure, I agree with this point.

> 
> Also when people enter the emsdk environment with emsdk_env, I don't think it is a problem to activate the emsdk tools in it. The user did in fact enter the emsdk environment! (With python the root issue was that our bundled python has broken pip, not that we activated python in PATH)

IMHO, its not just that pip3 didn't work, its that by doing that we clobber any python installation the user might have on their machine.  We completely switch up the python installation.   Even if our installation works fine that is still rather rude.

> 
> If users don't want to get node, they could emsdk activate the individual tools that make up the sdk, and leave out node from it. Then emsdk won't add node in it?

I'm not sure how to do that (how could I install `latest` and leave out node today?),   but I don't care to much because I actually do want to use emsdk's node.

I guess I will abandon this until we get users who really care that we clobber the node in their existing PATH.

sbc100:
It seems that some folks really don't want our version of node clobber the one in their PATH: https://github.com/emscripten-core/emscripten/issues/4848

Perhaps we could at least put our version at the end of the PATH?

sbc100:
Another use on #4848 surprised/annoyed by having node PATH clobbered by emsdk.

I'm not sure how easy it is to install our tools without node...  and what is more we *do* want to use our own version of node *internally* even if the user has a more recent one in their path.

In short there are two versions of node in play here:

1. The node we use internall to run js compiler (this should be invisible to the users and does not need to be in the PATH)
2. The node version users will run the output in.

We think we should find some way for users to easily overide (2)...  my person opinion is that emsdk should not even provide (2) at all.. at least it shouldn't put it on the PATH.   We don't provide a browser runtime.. why should we provide a node runtime? We expect folks have their own runtimes if they want to execute the output.

PR #718, opened 2021-02-18
Rename llvm-git-master -> llvm-git-main

sbc100:
To match the branch rename that happened in #693.

### DIFF
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -2,7 +2,7 @@
   "tools": [
   {
     "id": "llvm-git",
-    "version": "master",
+    "version": "main",
     "bitness": 32,
     "install_path": "llvm/git",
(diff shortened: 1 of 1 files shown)


emsdk_manifest.json
```diff
@@ -464,25 +464,25 @@
   {
     "version": "upstream-master",
```
tlively, line 465: Should "upsteam-master" be renamed as well or will that mess things up?

tlively (approved):
Rubberstamp because I don't actually know what effect this change will have.

emsdk_manifest.json
sbc100: I think I'll leave that for now.   This is made of several components some of which do still use master (i.e. emscripten) so it not clear main is better here.

Issue #623, opened 2020-09-25
sdk-upstream-master-32bit cmake error


Juanot:
Everything went okay up to this point...
```
Running CMake: ['cmake', '-G', 'Unix Makefiles', '-DCMAKE_BUILD_TYPE=Release', '-DPYTHON_EXECUTABLE=/usr/bin/python3', '$HOME/Apps/emsdk-master/emscripten/master/tools/optimizer']
CMake Error: The source directory "$HOME/Apps/emsdk-master/emscripten/master/tools/optimizer" does not exist.
Specify --help for usage, or press the help button on the CMake GUI.
CMake invocation failed due to exception!
Working directory: $HOME/Apps/emsdk-master/emscripten/master_32bit_optimizer
Command '['cmake', '-G', 'Unix Makefiles', '-DCMAKE_BUILD_TYPE=Release', '-DPYTHON_EXECUTABLE=/usr/bin/python3', '$HOME/Apps/emsdk-master/emscripten/master/tools/optimizer']' returned non-zero exit status 1
Installation failed!
```

The installation seems to be missing some files.
Is this a bug?

sbc100:
I believe this was recently fixed in https://github.com/emscripten-core/emsdk/pull/622.   Can you try updating emsdk and try again?

Juanot:
I have updated, but I'm now getting this error:
Failed to run "npm ci" in installed Emscripten root directory /emsdk-master/emscripten/master! Please install node.js first!
Installation failed!

This despite having nodejs and npm installed.

sbc100:
Oh that message is a little misleading.  Looking that code it doing this :

```
  node_tool = find_latest_installed_tool('node')                                 
  if not node_tool:                                                              
    print('Failed to run "npm ci" in installed Emscripten root directory ' + tool.installation_path() + '! Please install node.js first!')                        
    return False    
```

So I I guess this is due to the fact that we don't have a any 32-bit node binaries available for x86 linux... because upstream node stopped supported 32-bit x86 binaries (see https://nodejs.org/en/download/).

You could hack the script such that `find_latest_installed_tool` return `/path/to/your/node/binary`?

I'm tempted to say that we simply don't support 32-bit linux because this 1/2 support is clearly not working.

Remind me again (sorry if we have been over this already) exactly why you can't run x86 linux?    

[...]

PR #723, opened 2021-02-21
Allow system npm in "emsdk install emscripten"

truboxl:
"emsdk install emscripten-master-64bit" is currently dependent on whether emscripten node is installed or not.

This change allow using system-provided node / npm command to proceed the installation. Also clarify the error message.

Other alternatives:
#640 maybe addresses this for a different reason
I don't think #714 addresses this
Another alternative is to provide node.js in source form (and compile it) if "emsdk install emscripten" must use emscripten node / npm, to which also can address #623

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1243,10 +1243,17 @@ def find_latest_installed_tool(name):
 def emscripten_npm_install(tool, directory):
   node_tool = find_latest_installed_tool('node')
   if not node_tool:
-    print('Failed to run "npm ci" in installed Emscripten root directory ' + tool.installation_path() + '! Please install node.js first!')
-    return False
+    try:
+      subprocess.check_output(['npm' + ('.cmd' if WINDOWS else ''), '--version'], stderr=subprocess.STDOUT, universal_newlines=True)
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -1243,10 +1243,17 @@ def find_latest_installed_tool(name):
 def emscripten_npm_install(tool, directory):
   node_tool = find_latest_installed_tool('node')
   if not node_tool:
-    print('Failed to run "npm ci" in installed Emscripten root directory ' + tool.installation_path() + '! Please install node.js first!')
-    return False
+    try:
+      subprocess.check_output(['npm' + ('.cmd' if WINDOWS else ''), '--version'], stderr=subprocess.STDOUT, universal_newlines=True)
+      node_path = os.path.dirname(shutil.which('npm' + ('.cmd' if WINDOWS else '')))
+    except subprocess.CalledProcessError:
```
sbc100, line 1249: We have a local `which` function in this file which you can use rather than shtuil.which (since we want to support python2). 

You don't need the try/catch then either.   You an just do 'node_path = which ...` and then `if not node_path`

truboxl: Ok

sbc100 (approved):

Issue #729, opened 2021-02-25
Allow install of arbitrary emscripten-releases builds


sbc100:
It should be possible to at least attempt to install any emscripten-releases build.

e.g.:
sdk-releases-upstream-67a19e434a3ae89eab13f4e6ed0bf927bd570238-64bit

sbc100:
This now works.

PR #730, opened 2021-02-26
Fix crash attempting to install fastcmp SDK by hash

sbc100:
Using the full SDK name was causing a crash here because the
version is not of the form 1.2.3.  e.g:

sdk-releases-upstream-3b8cff670e9233a6623563add831647e8689a86b-64bit


sbc100:
This is a tidy up in preparation for #729

kripken (approved):

PR #731, opened 2021-02-26
Limit expansion of %releases-tag% for fastcomp

sbc100:
Without this, all releases are made available as fastcomp
versions.

Also, sort the releases to that the most recent come last.
This fixes the `emsdk list` display which only displays
the two most recent versions.


sbc100:
This is a tidy up in preparation for #729

kripken (approved):

PR #734, opened 2021-02-26
Cleanup sys.argv handling. NFC.

sbc100:


kripken (approved):
lgtm. I think we have enough testing to cover changes like this? But it might be good to review those.

PR #735, opened 2021-02-26
Refactor test.py to use python unittest. NFC.

sbc100:
This change should not effect the actual test coverage, it just refactors code into methods
on a TestCase class.


kripken (approved):
(lgtm, but has some changes from another PR)

PR #736, opened 2021-02-27
Avoid post-install when nothing is installed

sbc100:
This avoids re-running the post-install scripts when commands such as
`./emsdk install latest` a re-run.  This re-running of npm ci can be
significant slowdown especially during testing and developerment.

Becuase of the refactoring this change change also means we exit ealier
when a given tool fails to install.  In general we want to error out as
early as possible on the first failure so as not to bury it.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1828,9 +1828,11 @@ class Tool(object):
       return None
 
   def install(self):
+    """Returns True if the Tool was installed of False if was skipped due to
+    already being installed.
+    """
     if self.can_be_installed() is not True:
-      print("The tool '" + str(self) + "' is not available due to the reason: " + self.can_be_installed())
-      return False
+      exit_with_error("The tool '" + str(self) + "' is not available due to the reason: " + self.can_be_installed())
 
     if self.id == 'sdk':
       return self.install_sdk()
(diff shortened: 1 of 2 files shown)


sbc100:
ping..

emsdk.py
```diff
@@ -1828,43 +1828,54 @@ def download_url(self):
...
-      return False
+      exit_with_error("The tool '" + str(self) + "' is not available due to the reason: " + self.can_be_installed())
 
     if self.id == 'sdk':
       return self.install_sdk()
     else:
       return self.install_tool()
 
   def install_sdk(self):
+    """Returns True if any SDK component was installed of False all componented
```
kripken, line 1843:
```suggestion
    """Returns True if any SDK component was installed or False if all components
```

emsdk.py
```diff
@@ -1828,43 +1828,54 @@ def download_url(self):
       return None
 
   def install(self):
+    """Returns True if the Tool was installed of False if was skipped due to
```
kripken, line 1831:
```suggestion
    """Returns True if the Tool was installed or False if was skipped due to
```

kripken (approved):

sbc100:
Darn auto-merge is maybe not as cool as I thought it was...

PR #738, opened 2021-03-04
Remove `update-tags` phase

sbc100:
Its not really very useful to cache the tot release in a local text
file.  Instead just fetch the revsion each time `install tot` is run.

This avoids folks accidentally installing "old" tot releases.

Also, make the output a little less chatty when looking up tot version.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -787,7 +787,7 @@ def git_recent_commits(repo_path, n=20):
 def git_clone(url, dstpath):
   debug_print('git_clone(url=' + url + ', dstpath=' + dstpath + ')')
   if os.path.isdir(os.path.join(dstpath, '.git')):
-    print("Repository '" + url + "' already cloned to directory '" + dstpath + "', skipping.")
+    debug_print("Repository '" + url + "' already cloned to directory '" + dstpath + "', skipping.")
     return True
   mkdir_p(dstpath)
   git_clone_args = []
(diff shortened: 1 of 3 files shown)


sbc100:
This change builds on the back of the #732

kripken:
Hmm, but with this one doesn't know what tot will get installed. So one can't edit the txt file. But I guess we don't want people doing that anymore now that that PR landed that you quoted, is that the idea?

sbc100:
Exactly, the text file doesn't really serve much purpose once you can just install any SHA you like directly from the command line.

Also, was editing this text file by hand ever documents or recommended anywhere?

sbc100:
I always found it a little odd that `./emsdk install tot` did "just work" ... not it does (at least according to my definition or "just work").

emsdk.py
```diff
@@ -2990,8 +2946,7 @@ def print_tools(t):
       rmfile(sdk_path(EMSDK_SET_ENV))
     return 0
   elif cmd == 'update-tags':
-    update_tags()
-    return 0
+    exit_with_error('`update-tags` is not longer needed.  To install the latest tot release just run `install tot`')
```
kripken, line 2949: How about making this a warning, and saying it's a no-op? I'd rather not break users.

kripken (commented):
lgtm % one issue

emsdk.py
sbc100: OK sounds good.

kripken (approved):

PR #740, opened 2021-03-05
Remove special version check for vs-tool. NFC.

sbc100:
As of #505 all tools write a check a version file so we don't need to
special case vs-tool anymore.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1551,13 +1551,6 @@ def find_msbuild_dir():
   return ''
 
 
-def get_installed_vstool_version(installed_path):
-  try:
-    return open(installed_path + "/version.txt", "r").read()
-  except:
-    return None
-
-
 class Tool(object):
   def __init__(self, data):
     # Convert the dictionary representation of the tool in 'data' to members of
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -1695,12 +1688,12 @@ def is_installed_version(self):
...
     if os.path.isfile(version_file_path):
       with open(version_file_path, 'r') as version_file:
-        return version_file.read() == self.name
+        return version_file.read().strip() == self.name
     return False
 
   def update_installed_version(self):
     with open(self.get_version_file_path(), 'w') as version_file:
-      version_file.write(self.name)
+      version_file.write(self.name + '\n')
```
kripken, line 1696: why add this to remove it elsewhere?

sbc100: Just a drive-by cleanup.  Since its text file it should always end in newline (i noticed because I run `cat <version_file>` which looks odd without the ending newline).

kripken: I'm not opposed but maybe it makes this technically not NFC? sgtm either way.

kripken (approved):

sbc100:
I think this is NFC either way.. the contents of those version text files is an implementation detail.    Changing the contents of these files or renaming them even is NFC from an emsdk user perspective IMHO.

PR #741, opened 2021-03-05
Fix for `construct_env` and `list` with tot build installed

sbc100:


emsdk.py
```diff
@@ -2182,6 +2182,17 @@ def load_releases_tags():
   if extra_release_tag:
     tags.append(extra_release_tag)
 
+  # Adds the currently installed SDK version, which might be custom
```
kripken, line 2185:
```suggestion
  # Adds the currently installed SDK version, which might be a custom
```

sbc100:
I know.. its not tested.   The problem is that test.py runs after activation so its environment is inherited from test.sh.    I will look into added a test in test.sh

emsdk.py
```diff
@@ -2182,6 +2182,17 @@ def load_releases_tags():
...
+  # version.  This means it will show up in `list` and work with
+  # `construct_env`
```
kripken, line 2187: is this function the right place to do this? I think `load_releases_tags` should just load the tags from the network. this seems like a separate operation?

sbc100: load_releases_tags is mostly loading tags from a local file: emscripten-releases-tags.txt.  Its nothing to do with the network.

sbc100:
Add testing.

emsdk.py
sbc100: I guess we could rename this function "find_available_sdk_releases" ?

sbc100: It can have extra versions inject either from the command line (`emsdk install sdk-upstream-<SHA>`) or from a version that already was installed previously using that method.

emsdk.py
```diff
@@ -2182,6 +2182,17 @@ def load_releases_tags():
   if extra_release_tag:
     tags.append(extra_release_tag)
 
+  # Adds the currently installed SDK version, which might be a custom
+  # version.  This means it will show up in `list` and work with
```
kripken, line 2186:
```suggestion
  # version that is not in the main list loaded before.
```

emsdk.py
kripken: I see what you mean, yeah.

emsdk.py
```diff
@@ -2182,6 +2182,17 @@ def load_releases_tags():
...
+  # `construct_env`
```
kripken, line 2187:
```suggestion
```

kripken (approved):
lgtm with minor comment suggestions

kripken:
(not sure if the comment suggestions are good or not. but it seems like the key idea is that the main list was loaded before, and this is an extra one not in the list)

sbc100:
(sorry I should stop using auto-merge until its easier for you to see when its going to fire.)

PR #742, opened 2021-03-05
Feedback from #741.  NFC.

sbc100:


emsdk.py
```diff
@@ -2182,9 +2182,8 @@ def load_releases_tags():
   if extra_release_tag:
     tags.append(extra_release_tag)
 
-  # Adds the currently installed SDK version, which might be a custom
-  # version.  This means it will show up in `list` and work with
-  # `construct_env`
+  # Explicitly add the currently installed SDK version.  This could be a custom
+  # version (installed explicitly) so might be part of the main list loaded above.
```
kripken, line 2186:
```suggestion
  # version (installed explicitly) so it might not be part of the main list loaded above.
```

kripken (commented):
lgtm % comment (and not pressing "approve" to avoid a possible automerge :wink: )

PR #747, opened 2021-03-08
Change base to Ubuntu 20.04.2 LTS (Focal Fossa).

MarkCallow:
This is principally, for me at least, to be able to include a more up to date version of CMake. Focal has 3.16.3. The Debian Buster based image had CMake 3.13. Installing the most recent CMake from KitWare into that Docker image proved impossible because it had a dependency on a version of libssl not in Buster.

`test_dockerimage.sh` runs successfully using the updated image as does my own project's build.

The size of this image is 1.67GB. The Buster-based image is 1.59GB.

Note that I have left intact all the explicit package installs as I did not want to spend time determining which of them may be installed by default in Focal.

After you merge this PR, how can I determine when the emscripten/emsdk image at docker.com has been updated?

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:buster AS stage_build
+FROM ubuntu:focal AS stage_build
 
 ARG EMSCRIPTEN_VERSION=tot
 ENV EMSDK /emsdk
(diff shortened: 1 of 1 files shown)


sbc100:
Can you update the title to say "Ubuntu 20.04.2 LTS (Focal Fossa)" so its clear that its TLS and when its from?

sbc100 (approved):
Seems reasonable to me.

PR #749, opened 2021-03-08
Fix race between `install tot` and `active tot`

sbc100:
Since #732 we lookup tot release dynamically, but we don't
necessarily want to do this for the `activate` command otherwise
`install` followed by `activate` a can fail if a new build was
produced inbetween.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2155,6 +2155,15 @@ def load_releases_info():
   return load_releases_info.cached_info
 
 
+def get_installed_sdk_version():
+  version_file = sdk_path(os.path.join('upstream', '.emsdk_version'))
+  if not os.path.exists(version_file):
+    return None
+  with open(version_file) as f:
+    version = f.read()
+  return version.split('-')[2]
+
+
 # Get a list of tags for emscripten-releases.
 def load_releases_tags():
   tags = []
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -2587,6 +2592,15 @@ def expand_sdk_name(name):
   elif name in ('latest-upstream', 'latest-clang-upstream', 'latest-releases-upstream'):
     return str(find_latest_releases_sdk('upstream'))
   elif name in ('tot', 'sdk-tot', 'tot-upstream'):
+    if activating:
+      # When we are activating a tot release, assume that the currently
+      # installed SDK, if any, is the tot release we want to activate.
+      # Without this `install tot && activate tot` will race with the builders
+      # that are producing new builds.
```
kripken, line 2599: What if the user does `emsdk install tot ; emsdk install 2.0 ; emsdk activate tot` - that would activate 2.0? So `activate tot` is a little misleading.

If we want to change it this way, perhaps it should be just `activate` without specifying `tot` (not immediately, but deprecate the `tot`)?

Or, we could save the last installed `tot` version in a file, and activate that. That seems safer to me.

sbc100: Sure.. I did think of doing that but decided it probably wan't worth it.

Since we can only have one sdk installed a time the whole idea of activation make little sense with today's "only one SDK at a time" approach.  

Do you think we need to worry about folks who run `emsdk activate tot` without running `emsdk install tot` first?

sbc100: (I actually miss the ability to install many SDKs and activate just one of them).

kripken: I forgot we only allow one at a time... we did at least fix things to avoid repeated downloads, but not installs, then?

If we only allow one then `emsdk activate` seems unnecessary, and maybe we can remove it, unless we do want to support multiple installs again.

emsdk.py
```diff
@@ -2587,6 +2592,15 @@ def expand_sdk_name(name):
...
+      installed = get_installed_sdk_version()
+      if installed:
+        debug_print('activating currently installed SDK; not updating tot version')
```
kripken, line 2602: would this warning not look worrying to people? this goes in the other way of my last comment I guess. but we'd be showing this to most of our users if they enable debug printing. I guess that is rare?

kripken (approved):
lgtm for now, but we should think more about the multi-install situation.

emsdk.py
sbc100: Yes this is debug message, not a warning, so most folks will never see it.

PR #753, opened 2021-03-10
Fix support for Apple M1

juj:
Fix support for Apple M1. Node.js will still run via Rosetta 2 emulation since they do not yet have M1 support, but Python, LLVM, Emscripten and Binaryen will be native.

In order for Python to actually be native, will need to upload new Python to Google CDN built with update_python.py.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,6 @@
 *.pyc
 __pycache__
 
-.DS_Store
-
 # Support for --embedded configs
 /.emscripten
 /.emscripten.old
@@ -33,6 +31,3 @@ __pycache__
 /mingw
 /spidermonkey
 /binaryen
-
-/scripts/cpython
-/scripts/python-*
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 *.pyc
 __pycache__
 
+.DS_Store
+
 # Support for --embedded configs
 /.emscripten
 /.emscripten.old
@@ -31,3 +33,6 @@ __pycache__
 /mingw
 /spidermonkey
 /binaryen
+
+/scripts/cpython
+/scripts/python-*
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -483,13 +483,13 @@
   {
     "version": "upstream-main",
     "bitness": 64,
-    "uses": ["llvm-git-main-64bit", "node-14.15.5-64bit", "python-3.7.4-pywin32-64bit", "emscripten-main-64bit", "binaryen-main-64bit"],
+    "uses": ["python-3.7.4-pywin32-64bit", "llvm-git-main-64bit", "node-14.15.5-64bit", "emscripten-main-64bit", "binaryen-main-64bit"],
     "os": "win"
   },
   {
     "version": "upstream-main",
     "bitness": 64,
-    "uses": ["llvm-git-main-64bit", "node-14.15.5-64bit", "python-3.9.2-1-64bit", "emscripten-main-64bit", "binaryen-main-64bit"],
+    "uses": ["python-3.9.2-1-64bit", "llvm-git-main-64bit", "node-14.15.5-64bit", "emscripten-main-64bit", "binaryen-main-64bit"],
     "os": "macos"
   },
   {
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -292,12 +292,21 @@
   },
   {
     "id": "python",
-    "version": "3.7.4-2",
+    "version": "3.9.2-1",
+    "bitness": 64,
+    "arch": "x86_64",
+    "macos_url": "python-3.9.2-1-macos.tar.gz",
+    "activated_cfg": "PYTHON='%installation_dir%/bin/python3'",
+    "activated_env": "EMSDK_PYTHON=%installation_dir%/bin/python3;SSL_CERT_FILE=%installation_dir%/lib/python3.9/site-packages/certifi/cacert.pem"
+  },
+  {
+    "id": "python",
+    "version": "3.9.2-1",
     "bitness": 64,
     "arch": "aarch64",
-    "macos_url": "python-3.7.4-2-macos.tar.gz",
+    "macos_url": "python-3.9.2-1-macos-arm64.tar.gz",
     "activated_cfg": "PYTHON='%installation_dir%/bin/python3'",
-    "activated_env": "EMSDK_PYTHON=%installation_dir%/bin/python3;SSL_CERT_FILE=%installation_dir%/lib/python3.7/site-packages/certifi/cacert.pem"
+    "activated_env": "EMSDK_PYTHON=%installation_dir%/bin/python3;SSL_CERT_FILE=%installation_dir%/lib/python3.9/site-packages/certifi/cacert.pem"
   },
   {
     "id": "java",
@@ -480,7 +489,7 @@
   {
     "version": "upstream-master",
     "bitness": 64,
-    "uses": ["llvm-git-main-64bit", "node-14.15.5-64bit", "python-3.7.4-2-64bit", "emscripten-master-64bit", "binaryen-master-64bit"],
+    "uses": ["llvm-git-main-64bit", "node-14.15.5-64bit", "python-3.9.2-1-64bit", "emscripten-master-64bit", "binaryen-master-64bit"],
     "os": "macos"
   },
   {
@@ -573,6 +582,7 @@
     "bitness": 64,
     "uses": ["node-14.15.5-64bit", "python-3.7.4-2-64bit", "releases-upstream-%releases-tag%-64bit"],
     "os": "macos",
+    "arch": "x86_64",
     "custom_install_script": "emscripten_npm_install"
   },
   {
@@ -594,6 +604,7 @@
     "bitness": 64,
     "uses": ["node-14.15.5-64bit", "python-3.7.4-2-64bit", "releases-fastcomp-%releases-tag%-64bit"],
     "os": "macos",
+    "arch": "x86_64",
     "custom_install_script": "emscripten_npm_install"
   },
   {
@@ -644,6 +655,7 @@
     "bitness": 32,
     "uses": ["fastcomp-clang-e%precompiled_tag32%-32bit", "node-8.9.1-32bit", "python-3.7.4-2-64bit", "emscripten-%precompiled_tag32%"],
     "os": "macos",
+    "arch": "x86_64",
     "version_filter": [
       ["%precompiled_tag32%", ">", "1.37.22"]
     ]
@@ -653,6 +665,7 @@
     "bitness": 64,
     "uses": ["fastcomp-clang-e%precompiled_tag64%-64bit", "node-8.9.1-64bit", "python-3.7.4-2-64bit", "emscripten-%precompiled_tag64%"],
     "os": "macos",
+    "arch": "x86_64",
     "version_filter": [
       ["%precompiled_tag64%", ">", "1.37.22"]
     ]
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -295,7 +295,7 @@
     "version": "3.9.2-1",
     "bitness": 64,
     "arch": "x86_64",
-    "macos_url": "python-3.9.2-1-macos.tar.gz",
+    "macos_url": "python-3.9.2-1-macos-x86_64.tar.gz",
     "activated_cfg": "PYTHON='%installation_dir%/bin/python3'",
     "activated_env": "EMSDK_PYTHON=%installation_dir%/bin/python3;SSL_CERT_FILE=%installation_dir%/lib/python3.9/site-packages/certifi/cacert.pem"
   },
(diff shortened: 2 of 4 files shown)


emsdk_manifest.json
```diff
@@ -289,6 +290,15 @@
     "activated_cfg": "PYTHON='%installation_dir%/bin/python3'",
     "activated_env": "EMSDK_PYTHON=%installation_dir%/bin/python3;SSL_CERT_FILE=%installation_dir%/lib/python3.7/site-packages/certifi/cacert.pem"
   },
+  {
+    "id": "python",
+    "version": "3.7.4-2",
+    "bitness": 64,
+    "arch": "aarch64",
+    "macos_url": "python-3.7.4-2-macos.tar.gz",
```
sbc100, line 298: Won't this need that architecture string embedded in the filename to disambiguate from the one right above?

How about `python-3.7.4-2-macos-aarch64.tar.gz`? 

Also this version seems now out of sync with the updated version in the script below (3.9.2)

scripts/update_python.py
```diff
@@ -24,15 +24,17 @@
...
 import subprocess
 import sys
 from subprocess import check_call
 
-version = '3.7.4'
+version = '3.9.2'
+major_minor_version = '.'.join(version.split('.')[:2])  # e.g. '3.9.2' -> '3.9'
 base = 'https://www.python.org/ftp/python/%s/' % version
-revision = '2'
+revision = '1'
```
sbc100, line 37: Can we do this without bumping the version, or if this needed for M1 support?

.gitignore
```diff
@@ -31,3 +33,6 @@ __pycache__
 /mingw
 /spidermonkey
 /binaryen
+
+/scripts/cpython
+/scripts/python-*
```
sbc100, line 38: In general I try to resist adding stuff to gitignore.    In this case I kind of like that I can cleanup the build stuff with `git clean`.   If you disagree then maybe we can make a single `build` or `out` directory that we can ignore and have the scripts use that do to their building.

.gitignore
```diff
@@ -1,6 +1,8 @@
 *.pyc
 __pycache__
 
+.DS_Store
```
sbc100, line 4: .. I don't like adding OS-specific stuff like this but I guess its unavoidable (we have this in emscripten too).

Out if interest, as a mac user presumably you can this to your `~/.gitignore` and then it will apply to all the git repositories you ever user?    I've wondered in past this I/we should recommend that mac users do that.. I'm wondering if there are any downsides?

sbc100 (commented):
Generally looks great.  If you want to attach the python tar archive I'm happy to upload it to google storage for you.

scripts/update_python.py
juj: The old version does not build for M1, updating to latest version was necessary to build clean.

.gitignore
juj: I think we should have in .gitignore the files that the build scripts in this very repository itself generate, that will be most straightforward. What is the downside in having these in here anyway?

.gitignore
juj: The downside is that it is asking people to add their own .gitignores. I think it is better to have this here, so nobody will accidentally submit a PR with these files. The cost for having this here is practically zero anyways.

sbc100: Sure, I'm OK with leaving this here. 

What do you think about adding this to your `~/.gitignore` too though?   It might make it easier to work with repos that don't have this entry in their `.gitignore`?

emsdk_manifest.json
sbc100: What do you think about attaching the `.tar.gz` and I can upload it before we land this?

juj:
macOS Arm64 python: http://clb.confined.space/dump/python-3.9.2-1-macos-arm64.tar.gz

.gitignore
juj: `git clean -x` is the standard flag to clean also in .gitignore. It sounds like what you want to run is a `git clean -xdf` to nuke everything including gitignore?

juj: Sounds like it is better to leave that for later, I don't have time to spend more back and forth on this but need to land this support. The .gitignore contents are not relevant to M1 support, so left this out.

scripts/update_python.py
sbc100: sgtm

sbc100 (approved):
lgtm!

arm64 python uploaded.

Do you want me to upload the x86-64 mac python too or (perhaps better) leave that version as if for now (will require forking `upstream-main` on line 490 I guess?)

juj:
Our min spec for Intel macOS is version 10.13, so I've been wrangling today a chance to do the Intel side build either on real macOS 10.13, or with environment variables `MACOSX_DEPLOYMENT_TARGET=10.13 SDKROOT=/Developer/SDKs/MacOSX10.13.sdk` and 
```
./configure CFLAGS="-mmacosx-version-min=10.13 -isysroot /Developer/SDKs/MacOSX10.13.sdk" \
            CXXFLAGS="-mmacosx-version-min=10.13 -isysroot /Developer/SDKs/MacOSX10.13.sdk" \
            LDFLAGS="-mmacosx-version-min=10.13 -isysroot /Developer/SDKs/MacOSX10.13.sdk"
```
(where `/SDKs/MacOSX10.13.sdk` should be replaced with whatever SDK path one actually has installed), but I wasn't able to get that going today.

If you'd like to do an upload, if you can do it with the above flags, that'd be great. Or you can do just an upload with a regular build, and I can also do a local build just for our Unity deployment purposes. Or if you want to wait, I can submit that build as well once I get my hands on a suitable Intel mac environment to work with.

sbc100:
> Can we just least the x86-64 mac python on 3.7.4 and not upload anything new there?

This would also make this change more precise as it will effect M1 users.

juj:
> Can we just least the x86-64 mac python on 3.7.4 and not upload anything new there?

We could, but it will require splitting the manifest sdks part per both macOS archs. I would prefer not to do that.

sbc100:
> > Can we just least the x86-64 mac python on 3.7.4 and not upload anything new there?
> 
> We could, but it will require splitting the manifest sdks part per both macOS archs. I would prefer not to do that.

I don't really mind which option we choose, but this PR is currently blocked on getting a new python x86-64 binary right?   Forking the manifest is one way to unblock it.   I don't have access to a macOS machine myself.   (I guess its possible to use the circleci mac via SSH bot perhaps?)

juj:
>  this PR is currently blocked on getting a new python x86-64 binary right?

That's right - I'll unblock that tomorrow and upload the new binary, did not have a chance to do it today.

juj:
Added http://clb.confined.space/dump/python-3.9.2-1-macos-x86_64.tar.gz for macOS 10.13 min version build.

Also can you reupload http://clb.confined.space/dump/python-3.9.2-1-macos-arm64.tar.gz . I updated that one to run with minimum version 11.0 (rather than 11.1 that it was previously targeting).

Also updated the PR to specify these 10.13 and 11.0 minimum requirements for the build script.

sbc100:
> Added http://clb.confined.space/dump/python-3.9.2-1-macos-x86_64.tar.gz for macOS 10.13 min version build.
> 
> Also can you reupload http://clb.confined.space/dump/python-3.9.2-1-macos-arm64.tar.gz . I updated that one to run with minimum version 11.0 (rather than 11.1 that it was previously targeting).
> 
> Also updated the PR to specify these 10.13 and 11.0 minimum requirements for the build script.

Done

What was the min version in the previous/current version of python for x86_64?

juj:
> What was the min version in the previous/current version of python for x86_64?

It would be whatever macOS version was installed on the system where the zips were created on.

scripts/update_python.py
```diff
@@ -86,19 +88,36 @@ def build_python():
...
       check_call(['git', 'clone', 'https://github.com/python/cpython'])
     check_call(['git', 'checkout', 'v' + version], cwd=src_dir)
-    check_call(['./configure'], cwd=src_dir)
-    check_call(['make', '-j', str(multiprocessing.cpu_count())], cwd=src_dir)
-    check_call(['make', 'install', 'DESTDIR=install'], cwd=src_dir)
+
+    min_macos_version_line = '-mmacosx-version-min=' + min_macos_version  # Specify the min OS version we want the build to work on
+    build_flags = min_macos_version_line + ' -Werror=partial-availability'  # Build against latest SDK, but issue an error if using any API that would not work on the min OS version
+    env = os.environ.copy()
+    env['MACOSX_DEPLOYMENT_TARGET'] = min_macos_version
```
sbc100, line 117: Apparently you only need either the command line of the environment variable, not both:

https://stackoverflow.com/questions/25352389/what-is-the-difference-between-macosx-deployment-target-and-mmacosx-version-min

scripts/update_python.py
```diff
@@ -86,19 +88,36 @@ def build_python():
...
         check_call(['brew', 'install', 'openssl', 'pkg-config'])
-        os.remove('/usr/local/opt/openssl/lib/libssl.dylib')
-        os.remove('/usr/local/opt/openssl/lib/libcrypto.dylib')
-        os.environ['PKG_CONFIG_PATH'] = '/usr/local/opt/openssl/lib/pkgconfig/'
+        if platform.machine() == 'x86_64':
+            prefix = '/usr/local'
+            min_macos_version = '10.13'
+        elif platform.machine() == 'arm64':
+            prefix = '/opt/homebrew'
+            min_macos_version = '11.0'
```
sbc100, line 96: else assert?

sbc100 (approved):

sbc100:
> > What was the min version in the previous/current version of python for x86_64?
> 
> It would be whatever macOS version was installed on the system where the zips were created on.

I wonder if we can figure that out.. it would be good to know if we are increasing of decreasing the min requirments with this change.. maybe we inspect the binaries to figure it out (i'll give it a go)

sbc100:
Perhaps should also mention this in the  `Mac OS X / System Requirements` section of the README.md?

sbc100:
Looks like the previous version was 10.12:

```
$ llvm-objdump -p python3
Load command 9
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.12
      sdk 10.12
```

scripts/update_python.py
juj: That is true, but there is ambiguity in which method is preferred (https://stackoverflow.com/questions/25352389/what-is-the-difference-between-macosx-deployment-target-and-mmacosx-version-min highlights the cmdline flag no longer works, and https://github.com/devernay/xcodelegacy#using-the-older-sdks highlights the environment variable should be deprecated), and which method is supported on different compilers, so the general practice is to supply both. There is no harm in specifying them both for the build.

scripts/update_python.py
juj: the code will fail on the next line with these variables not defined, preferred not to add redundant lines (like the unnecessary `else:` in `if x: return foo else:` you often comment about)

juj:
macOS 10.13 was released in September 2017. Cross comparing system requirements at https://en.wikipedia.org/wiki/MacOS_High_Sierra vs https://en.wikipedia.org/wiki/MacOS_Sierra , there are no mac hardware that would not be able to update from macOS 10.12 to macOS 10.13 (and hardware from 2009 onwards is compatible), so not concerned about bumping the minimum version.

Emsdk on macOS requires system installed python in order to install Emsdk bundled python.

According to https://medium.com/@youngstone89/mac-os-x-sierra-python-installation-28db734a2a60 and https://stackoverflow.com/questions/41647424/how-to-install-updated-python-on-mac-os-sierra-with-brew, macOS 10.12 had Python 2.7.10 preinstalled, no Python 3.

According to https://stackoverflow.com/a/33175893 and https://stackoverflow.com/questions/51804203/atom-prioritizes-python-2-7-over-python-3-7-on-macos macOS 10.13 ships with the same Python 2.7.10 as well. So that will be the same minimum version that Emsdk needs to work with.

sbc100:
> macOS 10.13 was released in September 2017. Cross comparing system requirements at https://en.wikipedia.org/wiki/MacOS_High_Sierra vs https://en.wikipedia.org/wiki/MacOS_Sierra , there are no mac hardware that would not be able to update from macOS 10.12 to macOS 10.13 (and hardware from 2009 onwards is compatible), so not concerned about bumping the minimum version.
> 
> Emsdk on macOS requires system installed python in order to install Emsdk bundled python.
> 
> According to https://medium.com/@youngstone89/mac-os-x-sierra-python-installation-28db734a2a60 and https://stackoverflow.com/questions/41647424/how-to-install-updated-python-on-mac-os-sierra-with-brew, macOS 10.12 had Python 2.7.10 preinstalled, no Python 3.
> 
> According to https://stackoverflow.com/a/33175893 and https://stackoverflow.com/questions/51804203/atom-prioritizes-python-2-7-over-python-3-7-on-macos macOS 10.13 ships with the same Python 2.7.10 as well. So that will be the same minimum version that Emsdk needs to work with.

This is all fine with me.  I just want to make sure we track this stuff.

Thanks for investigating.

PR #754, opened 2021-03-11
Rename sdk-upstream-master SDK to sdk-upstream-main

sbc100:
Now that all the components (binaryen, emscripten and llvm) use `main`
as the brnach name is makes sense to give the SDK this name.

Keep backwards compat with the old name but issue a warning when its
used.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2588,7 +2588,7 @@ def exit_with_fastcomp_error():
 
 def expand_sdk_name(name, activating):
   if 'upstream-master' in name:
-    errlog('upstream-master SDK has been renames upstream-main')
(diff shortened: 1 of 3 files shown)


emsdk.py
```diff
@@ -2588,6 +2587,9 @@ def exit_with_fastcomp_error():
 
 
 def expand_sdk_name(name, activating):
+  if 'upstream-master' in name:
+    errlog('upstream-master SDK has been renames upstream-main')
```
kripken, line 2591:
```suggestion
    errlog('upstream-master SDK has been renamed upstream-main')
```

kripken (approved):

Issue #759, opened 2021-03-14
Generate html file with Bazel Emscripten toolchain


Nerdyvedi:
Hi,
I want to generate html file in addition to wasm and javascript "glue code". I am using bazel emscripten toolchain.

I want to do something like the following
`emcc hello.c -s WASM=1 -o hello.html
`
But not able to understand what changes I need to make in the Bazel build file to get the html file as well.

Thanks

walkingeyerobot:
Ok, this is now mostly done.

Right now the HTML minifier doesn't appear to work. While the npm package `html-minifier-terser` is correctly downloaded and installed, it runs it via `external/emscripten/emscripten/node_modules/.bin/html-minifier-terser`, which is a symlink to `node_modules/html-minifier-terser/cli.js`. This is correct, except `cli.js` has a bunch of relative paths, so running it via the symlink doesn't work. I don't yet know if this is an emscripten bug or a bazel specific bug. I think the latter is more likely, but it's also possible not a ton of people are using the html minifier and this has just gone unreported.

For now, just pass `linkopts = ["--oformat=html", "-s MINIFY_HTML=0"],`. I'll keep this bug open to track using the html minifier with the bazel toolchain.

sbc100:
@walkingeyerobot I assume we can now close this old issue?

PR #765, opened 2021-03-18
Bazel add html support

walkingeyerobot:
This gets us most of the way there for #759.

### DIFF
--- a/bazel/emscripten_toolchain/link_wrapper.py
+++ b/bazel/emscripten_toolchain/link_wrapper.py
@@ -52,13 +52,32 @@ parser.add_argument('--oformat')
 output_file = options.o
 oformat = options.oformat
 outdir = os.path.dirname(output_file)
-base_name_ext = os.path.basename(output_file)
-base_name = os.path.splitext(base_name_ext)[0]
+base_name = os.path.basename(output_file)
 
 # The output file name is the name of the build rule that was built.
 # Add an appropriate file extension based on --oformat.
-if oformat != None:
-  os.rename(output_file, output_file + '.' + oformat)
+if oformat is not None:
+  base_name_split = os.path.splitext(base_name)
+
+  # If the output name has no extension, give it the appropriate extension.
+  if not base_name_split[1]:
+    os.rename(output_file, output_file + '.' + oformat)
+
+  # If the output name does have an extension and it matches the output format,
+  # change the base_name so it doesn't have an extension.
+  elif base_name_split[1] == '.' + oformat:
+    base_name = base_name_split[0]
+
+  # If the output name does have an extension and it does not match the output
+  # format, change the base_name so it doesn't have an extension and rename
+  # the output_file so it has the proper extension.
+  # Note that if you do something like name your build rule "foo.js" and pass
+  # "--oformat=html", emscripten will write to the same file for both the js and
+  # html output, overwriting the js output entirely with the html.
+  # Please don't do that.
+  else:
+    base_name = base_name_split[0]
+    os.rename(output_file, os.path.join(outdir, base_name + '.' + oformat))
 
 files = []
 extensions = [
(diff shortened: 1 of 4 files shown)


bazel/emscripten_toolchain/link_wrapper.py
```diff
@@ -54,8 +43,41 @@
...
   sys.exit(1)
 
-js_name = os.path.basename(js_output)
-base_name = os.path.splitext(js_name)[0]
+# Parse the arguments that we gave to the linker to determine what the output
+# file is named and what the output format is.
+parser = argparse.ArgumentParser(add_help=False)
+parser.add_argument('-o')
+parser.add_argument('--oformat')
+(options, _) = parser.parse_known_args(param_file_args)
```
sbc100, line 51: Maybe `options = parser.parse_known_args(param_file_args)[0]`  (or at least drop the braces on the left hand side of the expression.

bazel/emscripten_toolchain/link_wrapper.py
```diff
@@ -54,8 +43,41 @@
...
+output_file = options.o
+oformat = options.oformat
+outdir = os.path.dirname(output_file)
+base_name = os.path.basename(output_file)
+
+# The output file name is the name of the build rule that was built.
+# Add an appropriate file extension based on --oformat.
+if oformat is not None:
```
sbc100, line 59: If oformat is None don't we want to default to "js" since that is what emcc will do?

How about `parser.add_argument('--oformat', default='js')` above and then you can remove this condition?

bazel/emscripten_toolchain/link_wrapper.py
```diff
@@ -25,20 +26,8 @@
...
-# google3-only(TODO(b/139440956): Default to False once the bug is fixed)
-replace_response_file = any(' ' in a for a in param_file_args)
-
-if not os.path.splitext(orig_output)[1]:
-  js_output = orig_output + '.js'
-  param_file_args[output_index] = js_output
-  replace_response_file = True
-
 # Re-write response file if needed.
-if replace_response_file:
```
sbc100, line 41: (unrelated) I wonder if we can completely remove this at some point... how does bazel + clang deal with this issue?

bazel/emscripten_toolchain/link_wrapper.py
```diff
@@ -54,8 +43,41 @@
...
+  # change the base_name so it doesn't have an extension.
+  elif base_name_split[1] == '.' + oformat:
+    base_name = base_name_split[0]
+
+  # If the output name does have an extension and it does not match the output
+  # format, change the base_name so it doesn't have an extension and rename
+  # the output_file so it has the proper extension.
+  # Note that if you do something like name your build rule "foo.js" and pass
+  # "--oformat=html", emscripten will write to the same file for both the js and
+  # html output, overwriting the js output entirely with the html.
```
sbc100, line 76: This sounds like there is still a bug in emscripten?

sbc100 (approved):

bazel/emscripten_toolchain/link_wrapper.py
walkingeyerobot: Done.

bazel/emscripten_toolchain/link_wrapper.py
walkingeyerobot: This check is a bit overly cautious, but I prefer to keep it in there. If oformat is None, I'd rather emcc handle what to do by default, just in case emcc ever changes. Also, in the crosstool we're setting `--oformat=js` as a default, so realistically this won't ever be None.

bazel/emscripten_toolchain/link_wrapper.py
walkingeyerobot: There's an internal P3 bug filed with the bazel folks that we're tracking.

bazel/emscripten_toolchain/link_wrapper.py
walkingeyerobot: Possibly. Emscripten could ensure it's only writing to files it just created and throw an error. I can file a bug for later.

walkingeyerobot: Filed https://github.com/emscripten-core/emscripten/issues/13729 for later

Issue #696, opened 2021-02-05
Support loading as external Bazel workspace


attilaolah:
Support loading `@emsdk` as an external workspace, instead of merging it into the main workspace.

TODOs:

- [ ] Move `http_archive()` calls from `baze/WORKSPACE` to some `.bzl` file which is `load()` ed from the `WORKSPACE` file.
- [ ] Update `bazel/README` to explain how to load the workspace as external.
- [ ] Add a test that loads `@emsdk` as an external workspace.

sbc100:
Thanks.   @walkingeyerobot might have opinions.

walkingeyerobot:
Sounds great to me. I wasn't aware this was an option. I'll try to get to this soon if no one beats me to it.

sbc100:
I assume this has now be done?   @DoDoENT can we close this?

DoDoENT:
Bazel 8 has deprecated workspaces and Bazel 9 fully removed support for them. Only Bzlmod is now supported. I think this can be closed.

Issue #650, opened 2020-11-12
Make bazel toolchain more usable


bianpengyuan:
Right now bazel toolchain is not very user friendly since
* It requires copy all tool chain files, which is hard to maintain.
* It introduces bazel dependencies from rule_nodejs, which is unnecessary for cpp target.
* It hard codes to use linux binary, which will break on other platforms.

walkingeyerobot:
Thanks for the feedback! I'm happy to look into this, but I could use a few more details.

> It requires copy all tool chain files, which is hard to maintain.
1) Bazel has to get a copy of emscripten from somewhere. I've found this is pretty easy to set up in clean environments. What would you prefer here?

> It introduces bazel dependencies from rule_nodejs, which is unnecessary for cpp target.
2) As Emscripten uses nodejs and node modules, depending on the bazel nodejs rules seems like a reasonable dependency. Are you having trouble with using the toolchain as a result? Can you elaborate more on why you dislike this dependency?

> It hard codes to use linux binary, which will break on other platforms.
3) This is clearly a deficiency. I agree it should be fixed. I'll work on adding mac / windows support.

sbc100:
I assume all these issues have been addressed in the intervening years?    @DoDoENT @walkingeyerobot ?

DoDoENT:
> I assume all these issues have been addressed in the intervening years? [@DoDoENT](https://github.com/DoDoENT) [@walkingeyerobot](https://github.com/walkingeyerobot) ?

Yes. All these are either addressed or no longer relevant.

PR #766, opened 2021-03-19
Use emsdk as external bazel dependency

celentes:
* Makes provided bazel rules look up @emsdk workspace instead of local workspace
* Uses system-specific emscripten binaries instead of defaulting to linux
* Provides macros for loading emsdk dependencies (nodejs and emscripten binaries)
* Unhardcodes paths in bazel rules and .sh wrappers
* `update_bazel_workspace.sh` now updates `revisions.bzl`
* `emscripten_deps()` can be fed with specific emscripten version
* Adds external usage test

Addresses #650 and #696


sbc100:
I'm not too famiilar with this stuff but it looks like there are a few different things going on here.  If its possible to split into smaller changes that would great (maybe its isn't?).

celentes:
Most of the changes are interlinked, I think it was possible to stop early on after externalising @emsdk, but after that it was a rollercoaster.
Specifically, unhardcoding paths in bazel setup would be a lot of repeated changes if it was to be split up.

celentes:
I could split off `revisions.bzl` shenanigans, but it's a relatively minor change. Will stop myself in the future.

celentes:
Have renamed `test-bazel` to `test-bazel-linux` as there's now two of them

bazel/emscripten_toolchain/crosstool.bzl
```diff
@@ -1006,6 +1007,22 @@ def _impl(ctx):
...
     crosstool_default_env_sets = [
+        # Globals
+        env_set(
+            actions = all_compile_actions +
+                      all_link_actions +
+                      [ACTION_NAMES.cpp_link_static_library],
+            env_entries = [
+                env_entry(
+                    key = "EM_BIN_PATH",
+                    value = ctx.attr.emscripten_binaries.label.workspace_root
```
walkingeyerobot, line 1018: use `emscripten_dir` here

bazel/emscripten_toolchain/crosstool.bzl
```diff
@@ -1042,49 +1059,42 @@ def _impl(ctx):
     features.append(crosstool_default_flags_feature)
 
     cxx_builtin_include_directories = [
-        "external/emscripten/emscripten/cache/sysroot/include/c++/v1",
-        "external/emscripten/emscripten/cache/sysroot/include/compat",
-        "external/emscripten/emscripten/cache/sysroot/include",
-        "external/emscripten/lib/clang/13.0.0/include",
+        emscripten_dir + "emscripten/cache/sysroot/include/c++/v1",
```
walkingeyerobot, line 1062: these don't look correct. I think the string needs to begin with a `/`. i.e.: `emscripten_dir + "/emscripten/cache/sysroot/include/c++/v1",`

walkingeyerobot (changes requested):
I'll probably have more comments as I continue to test.

bazel/emscripten_toolchain/crosstool.bzl
celentes: Good catch, didn't notice it

celentes:
Figured out how to do older versions of binaries. But that's material for the next PR.

bkotsopoulossc:
This is an awesome change! I've been playing with pulling in this branch, and it seems to work well. Really appreciate the work.

I have two things to raise, that maybe you would be interested in?

The first is, we are using the `--incompatible_enable_cc_toolchain_resolution` flag in our project, which does toolchain resolution at the [platform](https://docs.bazel.build/versions/master/platforms.html) level. It seems like wasm was added as a 'cpu' at the platform level already ([link](https://github.com/bazelbuild/platforms/pull/12/files)). So it seems possible to create a `toolchain` like
```
toolchain(
    name = "cc-toolchain-wasm",
    target_compatible_with = ["@platforms//cpu:wasm64"],
    toolchain = "@emsdk//emscripten_toolchain:cc-compiler-wasm",
    toolchain_type = "@bazel_tools//tools/cpp:toolchain_type",
)
``` 
And then maybe expose a function in this bazel repo to call `register_toolchains`? Is there interest in supporting something like that?

The second is, because the bazel workspace is in a subfolder, but all of the bazel target paths are relative to the bazel root, to get the paths to resolve I used a strip_prefix to kind of move the bazel subfolder to the root dir. Is this the right way of handling this?

```
emsdk_version = "dbda15d1c07ba5c434e63a8435913ecfae1bdf69"
http_archive(
    name = "emsdk",
    strip_prefix = "emsdk-%s/bazel" % emsdk_version,
    type = "zip",
    url = "https://github.com/emscripten-core/emsdk/archive/%s.zip" % emsdk_version,
    sha256 = "928e0afff70a7afe383038bfed4d7e3ed791ef8744a82966f9f14d11d5d585d1",
)
```

celentes:
@bkotsopoulossc 
I was gonna have that in this PR but stopped myself, it's enough changes for now.

To implement this, you need to do the toolchain as you said, then register it in workspace (ideally provided in deps) and also add platform in transition in `wasm_cc_binary.bzl`

celentes:
Another problematic bit is that `@platforms//cpu:wasm64` is not exposed by default, needs latest `@platforms` pulled

celentes:
And yes, the trick with strip_prefix is how i am using it in externals right now :)

bkotsopoulossc:
Makes total sense about keeping the scope of this PR down. I wonder if there is an interest in trying to support the platforms use case later on though?

celentes:
Well, bazel docs blatantly state that this is the future and all new crossplatform jazz should be developed with that in mind. The reason why cc rules are behind is because some platforms (android, primarily) does not support it yet

sbc100:
> Another problematic bit is that `@platforms//cpu:wasm64` is not exposed by default, needs latest `@platforms` pulled

Wait are you trying to actually use wasm64?  That is something that doesn't really exist yet, at least not at the emscripten level.  Do you mean wasm32?

celentes:
I think it's just a platforms flag. It's not "real" architecture, it's more of a label. Should be done against `wasm32` once we get to this

bkotsopoulossc:
I didn't actually try either, but didn't know if it made a difference since they seem more like labels like you mentioned? Anyways, good to know about using `wasm32`, thanks

celentes:
@bkotsopoulossc 
Another problem with automatic toolchain resolution and bazel-emscripten is that if you just run stuff with resolved toolchain, it creates tarballs instead of correct files, which need to be unpacked prior to using them first.
This is caused by bazel having opinions on which files you can and cannot produce with cc rules: https://github.com/bazelbuild/bazel/issues/2298

So until that gets resolved, you do need to use `wasm_cc_binary` rule to unpack them, and transition resolves the toolchain correctly without any extra input like platforms, cpu type etc.

walkingeyerobot (approved):

celentes:
so because test-bazel has been renamed, circleci expects it to pass

should I rename it back and have test-bazel and test-bazel-mac instead?

sbc100:
No thats OK, we just need to teach github about these new test names.

sbc100:
I temporarly removed "test-bazel" from the list of required steps to submit.

@walkingeyerobot if you are happy with this change I an land with with force

sbc100:
Oh looks like it worked on its own.

celentes:
Cheers!

sbc100:
Damit.. I closed it by accident.

I'm not sure what you want the final commit message to look like.  I'll leave that to you and/or @walkingeyerobot .

sbc100:
Are you good with the current PR title and description for the squashed change?

celentes:
` Use emsdk as external bazel dependency #766 ` works for me

celentes:
Does what it says on the tin. I'm gonna add a readme update tomorrow

bkotsopoulossc:
Sorry to revive this, any further thoughts on platform-based toolchain resolution? RE: https://github.com/emscripten-core/emsdk/pull/766#issuecomment-806194696

PR #767, opened 2021-03-20
Simplify the notation for installing by hash to just "emsdk install HASH"

kripken:
Previously this had to be

```
emsdk install sdk-releases-upstream-HASH
```

The only thing preventing using just the hash was that there was no
default for the backend, so defaulting to upstream fixes this.  And then
we can do

```
emsdk install HASH
```

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2767,6 +2767,8 @@ def expand_sdk_name(name, activating):
           backend = 'fastcomp'
       return 'sdk-releases-%s-%s-64bit' % (backend, release_hash)
     elif len(version) == 40:
+      if backend is None:
+        backend = 'upstream'
       global extra_release_tag
       extra_release_tag = version
       return 'sdk-releases-%s-%s-64bit' % (backend, version)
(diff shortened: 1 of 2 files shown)


emsdk.py
```diff
@@ -2767,6 +2767,8 @@ def expand_sdk_name(name, activating):
           backend = 'fastcomp'
       return 'sdk-releases-%s-%s-64bit' % (backend, release_hash)
     elif len(version) == 40:
+      if backend is None:
+        backend = 'upstream'
```
sbc100, line 2771: Just set `backend = 'upstream'` when it initialized on ling 2748?

sbc100 (approved):

emsdk.py
kripken: We can't do that, as we need to track if the user provided a backend. If not, we autodetect it based on the version on lines 2764-7. That code would break with your proposed change.

sbc100 (approved):

PR #770, opened 2021-03-22
List "Release" as the default build option

### LINKED ISSUE #760: Is the default still RelWithDebInfo?
https://github.com/emscripten-core/emsdk/blob/9c65611d96e68081f7c146012f5ede31e7de4d73/emsdk.py#L2806-L2809

https://github.com/emscripten-core/emsdk/search?q=+relwithdebinfo&type=commits

truboxl:
This was set back in commit cea44f475af10c6fde85d4906111a39818e3f134

Fixes #760

A minor change that clears up the confusion when building / packaging emscripten, the default build option does not match up with what was listed in `./emsdk.py`


sbc100 (approved):
Thanks!

PR #774, opened 2021-03-25
Update bazel readme

celentes:


celentes:
As promised, updates to readme post bazel setup changes.

celentes:
Here's a notion to think about:
Because bazel wants to download archives from a particular commit/tag, updating readme post changes has to go through a separate commit post-merge (as we don't know the url/hash/sha prior to merging).
The canonical alternative to that is point to the latest upstream release message like [this](https://github.com/bazelbuild/rules_foreign_cc/releases/tag/0.2.0). That way it's no longer part of the git/bazel conundrum, but the overhead is that it puts an extra step in release procedure.

Would that be acceptable? We can provide a script to generate the message.

walkingeyerobot (approved):

celentes:
I'm gonna figure out a way to automate it later on.

PR #776, opened 2021-03-27
Improve bazel example

jinfwhuang:
Added emscripten bindings to bazel examples.

### DIFF
--- a/bazel/README.md
+++ b/bazel/README.md
@@ -25,13 +25,16 @@ emsdk_emscripten_deps()
 
 Put the following lines into your `.bazelrc`:
 ```
-build:wasm --crosstool_top=//emscripten_toolchain:everything
+build:wasm --crosstool_top=@emsdk//emscripten_toolchain:everything
 build:wasm --cpu=wasm
 build:wasm --host_crosstool_top=@bazel_tools//tools/cpp:toolchain
 ```
 
-Simply pass `--config=wasm` when building a normal `cc_binary`. The result of
-this build will be a tar archive containing any files produced by emscripten.
+Simply pass `--config=wasm` when building a normal `cc_binary`, e.g.
+```
+bazel build --config=wasm :hello-world
+```
+The result of this build will be a tar archive containing any files produced by emscripten.
 
 ### Using wasm_cc_binary
 First, write a new rule wrapping your `cc_binary`.
--- a/bazel/test_external/.bazelrc
+++ b/bazel/test_external/.bazelrc
@@ -1,0 +1,3 @@
+build:wasm --crosstool_top=@emsdk//emscripten_toolchain:everything
+build:wasm --cpu=wasm
+build:wasm --host_crosstool_top=@bazel_tools//tools/cpp:toolchain
--- a/bazel/test_external/.gitignore
+++ b/bazel/test_external/.gitignore
@@ -1,2 +1,1 @@
 bazel-*
-dist/
--- a/bazel/test_external/.gitignore
+++ b/bazel/test_external/.gitignore
@@ -1,0 +1,2 @@
+bazel-*
+dist/
--- a/bazel/test_external/BUILD
+++ b/bazel/test_external/BUILD
@@ -3,6 +3,9 @@ load("@emsdk//emscripten_toolchain:wasm_rules.bzl", "wasm_cc_binary")
 cc_binary(
     name = "hello-world",
     srcs = ["hello-world.cc"],
+    linkopts = [
+        "--bind",
+    ],
 )
 
 wasm_cc_binary(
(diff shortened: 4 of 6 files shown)


bazel/README.md
```diff
@@ -25,13 +25,16 @@ emsdk_emscripten_deps()
 
 Put the following lines into your `.bazelrc`:
 ```
-build:wasm --crosstool_top=//emscripten_toolchain:everything
+build:wasm --crosstool_top=@emsdk//emscripten_toolchain:everything
```
jinfwhuang, line 28: This was a bug.

bazel/test_external/BUILD
```diff
@@ -3,6 +3,9 @@ load("@emsdk//emscripten_toolchain:wasm_rules.bzl", "wasm_cc_binary")
 cc_binary(
     name = "hello-world",
     srcs = ["hello-world.cc"],
+    linkopts = [
+        "--bind",
```
jinfwhuang, line 7: Required for embind.

https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html

bazel/test_external/.gitignore
```diff
@@ -0,0 +1,2 @@
+bazel-*
+dist/
```
sbc100, line 2: missing newline

bazel/test_external/hello-world.cc
```diff
@@ -1,6 +1,24 @@
...
+#ifdef __EMSCRIPTEN__
+EMSCRIPTEN_BINDINGS(aa) {
+  function("sayHello", &sayHello);
+}
+#else
 int main(int argc, char** argv) {
-  std::cout << "hello world!" << std::endl;
+  std::cout << "start main function" << std::endl;
+  sayHello();
+  std::cout << "end main function" << std::endl;
```
sbc100, line 21: Any not include a main function in the program in both cases?   This change seem to turn this example from the executable into a library.  Any particular reason for doing that?

Don't we actually run the result of building this?  (under node?)   We probably should.

bazel/test_external/BUILD
sbc100: Why is this not formatted on one line like `srcs` above?  Is this the bazel style?

jinfwhuang:
```
linkopts = ["--bind"],
```
works as well. Both comply with `buildifier`.

bazel/test_external/hello-world.cc
jinfwhuang: Both are fine. The example illustrates that we can use `EMSCRIPTEN_BINDINGS` optionally.

jinfwhuang:
Sure. The key addition to the example is that it demonstrates how to pass opt flag to the compiler to bind c functions to javascript. These should be helpful addition to the example for a majority of the users.

I highly question that this would make this more difficult to maintain.

bazel/test_external/hello-world.cc
jinfwhuang: I updated it so it looks more like what you are looking for.

jinfwhuang:
Remove `test_external/README.md`

bazel/test_external/.bazelrc
```diff
@@ -0,0 +1,3 @@
+build:wasm --crosstool_top=@emsdk//emscripten_toolchain:everything
```
jinfwhuang, line 1: without these, the example would not work for `:hello-world`.

sbc100:
> Sure. The key addition to the example is that it demonstrates how to pass opt flag to the compiler to bind c functions to javascript. These should be helpful addition to the example for a majority of the users.
> 
> I highly question that this would make this more difficult to maintain.
> 
> * Without the .bazelrc, the command `bazel build --config=wasm :hello-world` would not work.

What about keeping hello world as is and adding a new test (hello_embind or test_embind?) that adds this more complex stuff?

bazel/test_external/.bazelrc
walkingeyerobot: I think it's ok for the example to not work with `--config=wasm` right out of the box; we'd rather promote the use of `wasm_cc_binary` as it produces more useful output anyway. Also, the readme covers adding the appropriate lines to your own .bazelrc file, which would enable `--config=wasm` builds to function. The big thing here I want to avoid is having this file be identical to bazel/bazelrc and having to keep them both in sync.

jinfwhuang:
> > Sure. The key addition to the example is that it demonstrates how to pass opt flag to the compiler to bind c functions to javascript. These should be helpful addition to the example for a majority of the users.
> 
> Passing linkopts using bazel is not unique to this toolchain. This is something bazel documentation covers just fine. Similarly, using embind with and without this toolchain is the same, and emscripten's embind documentation covers how to use this.
> 
> > I highly question that this would make this more difficult to maintain.
> 
> The more duplication of documentation there is, the more places that need to be updated when something changes, and the more likely it is that they become out of sync, confusing users. I would be in favor of linking to other documentation instead.

@walkingeyerobot So you want to keep the example to be broken with respect to  with the example `hello-world` to stay broken?

sbc100:
Closing since this is likely very much out of date at this point.

Feel free to re-open if you disagree.

PR #791, opened 2021-04-19
Remove the use of a custom temp directory

juj:
Remove the use of a custom temp directory. That is a bad practice, we should reuse system temp directory instead, that is what it is there for.


sbc100 (approved):
Yes!

PR #792, opened 2021-04-19
Add option --override-repository

juj:
Add option --override-repository to allow controlling where git clones happen from.

With this method we can integrate emsdk builds to our CI.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1710,7 +1710,7 @@ class Tool(object):
 
     # Cache the name ID of this Tool (these are read very often)
     self.name = self.id
-    if len(self.version) > 0:
+    if self.version:
       self.name += '-' + self.version
     if hasattr(self, 'bitness'):
       self.name += '-' + str(self.bitness) + 'bit'
--- a/emsdk.py
+++ b/emsdk.py
@@ -182,6 +182,32 @@ if POWERSHELL:
 else:
   EMSDK_SET_ENV = os.path.join(emsdk_path(), 'emsdk_set_env.bat')
 
+
+# Parses https://github.com/emscripten-core/emscripten/tree/d6aced8 to a pair (https://github.com/emscripten-core/emscripten, d6aced8)
+def parse_github_url_and_refspec(url):
+  if not url:
+    return ('', '')
+
+  if url.endswith(('/tree/', '/tree', '/commit/', '/commit')):
+    raise Exception('Malformed git URL and refspec ' + url + '!')
+
+  if '/tree/' in url:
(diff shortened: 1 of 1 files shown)


juj:
Looks like flake8 is run using python2?

sbc100:
Yes, emsdk, unlike emscripten itself still needs to support python2 I believe so that it can bootstrap itself successfully on macOS where python2 comes pre-installed in the path but not python3.

emsdk.py
```diff
@@ -2900,6 +2936,14 @@ def extract_bool_arg(name):
       return True
     return False
 
+  def extract_string_arg(name):
+    nonlocal args
+    for i in range(len(args)):
+      if args[i] == name:
+        value = args[i + 1]
+        args = args[:i] + args[i + 2:]
```
sbc100, line 2944: Rather then re-assigning `args` you can maybe avoid the `nonlocal` thing by doing `del args[i:i+1]` which is in-place.

emsdk.py
```diff
@@ -182,6 +182,32 @@ def emsdk_path():
...
+      raise Exception('Malformed git URL and refspec ' + url + '!')
+    return url.split('/commit/')
+  else:
+    return (url, 'main')  # Assume the default branch is main in the absence of a refspec
+
+
+emscripten_config_directory = os.path.expanduser("~/")
+# If .emscripten exists, we are configuring as embedded inside the emsdk directory.
+if os.path.exists(os.path.join(emsdk_path(), '.emscripten')):
+  emscripten_config_directory = emsdk_path()
```
sbc100, line 209: emscripten_config_directory doesn't looked like it is used anywhere in this patch. 

Also there already exists `dot_emscripten_path()` function in this file and we transitioned to only supporting embedded configs.   It doesn't stop you using your own config file, but it just means that emsdk itself will only ever write an embedded config.

juj: Ops sorry, `emscripten_config_directory` was a leftover from the previous version of this code, removed it now.

juj:
Updated according to review, ptal.

emsdk.py
```diff
@@ -1684,7 +1709,9 @@ def __init__(self, data):
       setattr(self, key, value)
 
     # Cache the name ID of this Tool (these are read very often)
-    self.name = self.id + '-' + self.version
+    self.name = self.id
+    if len(self.version) > 0:
```
sbc100, line 1713: Just `if self.version:` is probably fine here?

sbc100 (approved):
lgtm % test failiures.

PR #793, opened 2021-04-19
Fix issue with npm install installing dev packages

juj:
Fix issues with npm install of google-closure-compiler installing all (unwanted) dev packages as well.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1411,7 +1411,7 @@ def emscripten_npm_install(tool, directory):
     print('Running post-install step: npm install', closure_compiler_native)
     try:
       subprocess.check_output(
-        [npm, 'install', closure_compiler_native],
+        [npm, 'install', '--production', '--no-optional', closure_compiler_native],
         cwd=directory, stderr=subprocess.STDOUT, env=env,
         universal_newlines=True)
     except subprocess.CalledProcessError as e:


emsdk.py
```diff
@@ -1383,7 +1383,7 @@ def emscripten_npm_install(tool, directory):
   print('Running post-install step: npm ci ...')
   try:
     subprocess.check_output(
-        [npm, 'ci', '--production', '--no-optional'],
+        [npm, 'ci', '--no-package-lock', '--omit=dev', '--omit=optional'],
```
sbc100, line 1386: I guess `--no-package-lock` means we no longer get pinned versions but just the latest versions from package.json?

Maybe that isn't a big problem and `package.json` already specifies constrained-enough version?

sbc100:
Hmm.. I seem to remember that we transitioned to bundling the npm packages in the uploaded archive which should make this step no longer needed (at least for recent versions of the emsdk).  I guess we never finished the transition.   

Also does the version of node that we bundle already deprecate these options?  If so now did this not cause our tests to fail already?   Or are they in the process of being deprecated in the currently bundled version maybe?

juj:
> Hmm.. I seem to remember that we transitioned to bundling the npm packages in the uploaded archive

Not sure I understand.. this step will be done for the compile-from-git-source installations, where we don't use precompiled archives?


> Also does the version of node that we bundle already deprecate these options?

Yes, this is using the version of node.js that is bundled.

>  If so now did this not cause our tests to fail already?

Is there a test to verify that only production packages are installed, and nothing more?

juj:
Actually, now I realize there is a simpler way to achieve this, so revised to use that instead. It will still use deprecated flags, but since npm is so brittle and it looks like multiple `--omit=` flags have some quirks(?), I'll stick to the `--production` flag.

sbc100:
> > Hmm.. I seem to remember that we transitioned to bundling the npm packages in the uploaded archive
> 
> Not sure I understand.. this step will be done for the compile-from-git-source installations, where we don't use precompiled archives?
> 

At yes, sorry, I see this is not used at all in the normal emsdk install path.

sbc100 (approved):

PR #797, opened 2021-04-23
Avoid setting `EM_CACHE` unless we really need to

sbc100:
This avoid polluting the global environment which makes
side-by-side installational of different emscripten version
harder.

See https://github.com/emscripten-core/emscripten/pull/13954

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2171,6 +2171,14 @@ def find_tot_sdk():
   return 'sdk-releases-upstream-%s-64bit' % (extra_release_tag)
 
 
+def parse_emscripten_version(emscripten_root):
+  version_file = os.path.join(emscripten_root, 'emscripten-version.txt')
+  with open(version_file) as f:
+    version = f.read().strip()
+    version = version.strip('"').split('.')
+    return [int(v) for v in version]
+
+
 # Given a git hash in emscripten-releases, find the emscripten
 # version for it. There may not be one if this is not the hash of
 # a release, in which case we return None.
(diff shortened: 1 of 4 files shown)


emsdk.py
```diff
@@ -2643,10 +2651,26 @@ def get_env_vars_to_add(tools_to_activate, system, user):
...
+      # - 'CACHE' config started being honored in 1.39.16
+      #   https://github.com/emscripten-core/emscripten/pull/11091
+      # - Default to embedded cache started in 1.39.16
+      #   https://github.com/emscripten-core/emscripten/pull/#11126
+      #
+      # Since setting EM_CACHE in the environment effects the entire machine
+      # we want to avoid this except when installing these older emscripten
+      # versions that really need it.
+      version = parse_emscripten_version(config['EMSCRIPTEN_ROOT'])
+      if version <= [1, 39, 16]:
```
kripken, line 2670:
```suggestion
      if version < [1, 39, 16]:
```
Seems like we need this in earlier versions, not this exact one?

sbc100:
I verified this locally by installing `sdk-fastcomp-1.38.33` and checking that it does at `EM_CONFIG` to the env whereas for newer versions it doesn't get added.

sbc100:
The other nice thing about this change is that folks like me who regularly switch between emsdk and non-emsdk setups have one less thing to do when we switch.

After this all I really need to do disable emsdk to to run:

```
$ unset EM_CONFIG
```

I no longer need to run `$ unset EM_CACHE` too.

kripken (approved):

juj (approved):

PR #801, opened 2021-04-27
Unset emsdk-related environment variables from inactive tools

sbc100:
When we deactivate a tool we also want to remove its environment
variables.   One driver for this is that modern sdks don't set
`EM_CACHE` whereas old ones did and we want to make sure that
`EM_CACHE` gets unset when folks upgrade (and then re-set if
they downgrade).  See #797.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2684,6 +2684,18 @@ def construct_env(tools_to_activate, system, user):
   return construct_env_with_vars(get_env_vars_to_add(tools_to_activate, system, user))
 
 
+def unset_env(key):
+  if POWERSHELL:
+    return 'Remove-Item env:%s\n' % key
+  if CMD:
+    return 'set %s=\n' % key
+  if CSH:
+    return 'unsetenv %s;\n' % key
+  if BASH:
+    return 'unset %s;\n' % key
+  assert False
+
+
 def construct_env_with_vars(env_vars_to_add):
   env_string = ''
   if env_vars_to_add:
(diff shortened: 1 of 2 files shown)


sbc100:
I should probably write test for this specifically around older SDKs and `EM_CACHE`...

sbc100:
In the long run, this change (combined with existing #797) should allow unity to move away from passing `--cache` to emcc since the point of doing that is to override the `EM_CACHE` environment variable set by emsdk.

test/test.sh
```diff
@@ -13,9 +13,23 @@ source ./emsdk_env.sh
...
+# Install an older version of the SDK that requires EM_CACHE to be
+# set in the environment, so that we can test it is later removed
+./emsdk install sdk-fastcomp-3b8cff670e9233a6623563add831647e8689a86b
+./emsdk activate sdk-fastcomp-3b8cff670e9233a6623563add831647e8689a86b
+source ./emsdk_env.sh
+which emcc
+emcc -v
+test -n "$EM_CACHE"
+
+# Install the latest version of the SDK which is the expeced precondition
```
kripken, line 25:
```suggestion
# Install the latest version of the SDK which is the expected precondition
```

emsdk.py
```diff
@@ -2684,43 +2684,55 @@ def construct_env(tools_to_activate, system, user):
...
+    if 'EMSDK_PYTHON' in env_vars_to_add:
+      # When using our bundled python we never want the user's
+      # PYTHONHOME or PYTHONPATH
+      # See https://github.com/emscripten-core/emsdk/issues/598
+      env_string += unset_env('PYTHONHOME')
+      env_string += unset_env('PYTHONPATH')
+
+  # Remove any environment variables that might have been set by old or
+  # inactive tools/sdks.
+  temp_keys = set(['EMSDK_POWERSHELL', 'EMSDK_CSH', 'EMSDK_CMD', 'EMSDK_BASH', 'EMSDK_NUM_CORES', 'EMSDK_TTY'])
```
kripken, line 2729: I do not follow what a "temp key" is. Temporary in what sense?

sbc100: I renamed to `ignore_keys` and added more comments.

juj:
lgtm, though this maybe a good item to throw a notification to mailing list, since this will visibly change the semantics for users.

Issue #805, opened 2021-04-29
switch primary branch from `master` to `main`


sbc100:
All the other repos in our ecosystem have made the switch now so its really just catchup and consistency now.

kripken:
This will be the most complex branch change we've done yet, I think, because checking out the emsdk from git is the main way users use it.

Some thoughts on a possible process:

* Add a `main` branch, forked from `master`. Do **not** remove `master` yet.
* Update docs to `main`, continue development there.
* Add a single new change to `master`, a warning shown on every invocation to switch to `main` (assuming that emsdk is in a git checkout, and the branch is still `master`). Without this, people will do `git pull` and silently not get updates.
* Some long time later, change that warning to an error. That will be noticed by CI and other things, that missed the warning.
* Some long time later, remove `master`.

Thoughts?

sbc100:
Sounds like a good plan.

truboxl:
Some commits got left out...
https://github.com/emscripten-core/emsdk/compare/main...master

sbc100:
At thanks.  Looks like it was just one commit.   The other one there was deliberate commit to master to add the warning.    

I will cherry-pick the other.. thanks!

kripken:
Looks like it's been over 2 weeks since the switch. I think no major issues have come up. Perhaps it is time to move to the next step, which is to make the warning on the master branch an error?

People using master but not seeing the warning are not getting broken, but they also are not getting new versions. So it may be good to do that sooner.

sbc100:
Sounds good to me.

skejeton:
Unfortunately the error message is incorrect
![image](https://user-images.githubusercontent.com/27076041/145492651-07e6527e-5e52-4f07-a2c0-a6cd4588b8c6.png)

kripken:
@skejeton 

Most likely this can be fixed by doing `git pull`, but if that doesn't work, what does `git show HEAD` print for you? And `git remote -v`?

sbc100:
The migration was completed back in 2021 so closing this out.  The `master` branch still exists and produces an error if used.

PR #809, opened 2021-05-03
Switch primary branch from `master` to `main`

sbc100:
See: #805


sbc100:
As per the plan in #805 we are not going to remove the actual master branch for some time.

kripken (approved):

PR #810, opened 2021-05-03
Add warning to those using `master` branch

sbc100:
See #805

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2842,6 +2842,18 @@ def expand_sdk_name(name, activating):
 
 
 def main(args):
+
+  if is_emsdk_sourced_from_github():
+    # This code only exists on the master branch
+    errlog('****')
+    errlog('Warning: You appear to be using the `master` branch of emsdk.')
+    errlog('We recently made the switch to using `main`')
+    errlog('In order to continue to receive updates you will need to make the switch locally too.')
+    errlog('For normal clones without any local branches simply running the following command should be enough:')
+    errlog('  `git checkout main`')
+    errlog('For more information see https://github.com/emscripten-core/emsdk/issues/805')
+    errlog('****')
+
   if not args:
     errlog("Missing command; Type 'emsdk help' to get a list of commands.")
     return 1


emsdk.py
```diff
@@ -2842,6 +2842,18 @@ def expand_sdk_name(name, activating):
...
 def main(args):
+
+  if is_emsdk_sourced_from_github():
+    # This code only exists on the master branch
+    errlog('****')
+    errlog('Warning: You appear to be using the `master` branch of emsdk.')
+    errlog('We recently made the switch to using `main`')
+    errlog('In order to continue to receive updates will need to make the switch locally too.')
+    errlog('For trivial checkouts simply running the following command should be enough:')
+    errlog('  `git checkout origin/main`')
```
kripken, line 2853: why `origin/main` instead of `main`?

kripken (commented):
Looks like this PR is targeted to `main` - should be `master`.

emsdk.py
sbc100: `main` does not exist after a fetch.. only `origin/main` exists at that point..

I guess `git checkout` does automatically create local branches based on remote ones do `git check main` will technically work.   Maybe I'm biased because I tend to always qualify my banch names when I want to refer to a remote branch like this.

kripken: Yes, it will automatically create a local branch for it, that's why I think it's ok to do. It's what I always depend on myself.

Plain `main` seems closer to plain `master`. Also it will be closer to what a user gets on a fresh clone. That's why I think it's simpler.

sbc100: Yeah, I think you are right.   

I'm an anomaly because I never have local branch called `master` or `main` since I use `git rebase-update` all the time and it (quite rightly I guess) deletes any branch that are the same as their upstream branch (i.e. it assumed they have been merged).    Local `main` has no point to exist in this workflow as `origin/main` is the upstream target of any local branches.

emsdk.py
```diff
@@ -2842,6 +2842,18 @@ def expand_sdk_name(name, activating):
 
 
 def main(args):
+
+  if is_emsdk_sourced_from_github():
+    # This code only exists on the master branch
+    errlog('****')
+    errlog('Warning: You appear to be using the `master` branch of emsdk.')
+    errlog('We recently made the switch to using `main`')
+    errlog('In order to continue to receive updates will need to make the switch locally too.')
```
kripken, line 2851:
```suggestion
    errlog('In order to continue to receive updates you will need to make the switch locally too.')
```

emsdk.py
```diff
@@ -2842,6 +2842,18 @@ def expand_sdk_name(name, activating):
...
+    errlog('For trivial checkouts simply running the following command should be enough:')
+    errlog('  `git checkout main`')
+    errlog('For more complex use cases see https://github.com/emscripten-core/emsdk/issues/805')
+    errlog('****')
```
kripken, line 2855: I don't follow these lines. What is a "trivial checkout"? And more complex cases do not seem to be mentioned in that link?

sbc100: I meant if you don't have any local branches .. I can clarity.

kripken (approved):

PR #811, opened 2021-05-04
2.0.19

sbc100:


kripken (dismissed):

sbc100:
Also added LTO build so folks can try it out.

kripken:
Aside from the test failures this lgtm to land. (Even before the LTO one finishes all tests - we can remove it later if we want.)

kripken:
Actually, are the lto and non-lto versions identical in the DEPS they are built from?

sbc100:
> Actually, are the lto and non-lto versions identical in the DEPS they are built from?

Yup that is my understanding

kripken:
Yes, I verified they are identical now, sorry. I was misreading it before.

PR #823, opened 2021-05-21
Error on the master branch. See #805

kripken:
Behavior is:
```
$ ./emsdk install 2.0.0
****
Error: You appear to be using the `master` branch of emsdk.
We recently made the switch to using `main`
In order to continue to receive updates you will need to make the switch locally too.
For normal clones without any local branches simply running the following command should be enough:
  `git checkout main`
For more information see https://github.com/emscripten-core/emsdk/issues/805
****
$ echo $?
1
```

See #805

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2818,13 +2818,14 @@ def main(args):
   if is_emsdk_sourced_from_github():
     # This code only exists on the master branch
     errlog('****')
-    errlog('Warning: You appear to be using the `master` branch of emsdk.')
+    errlog('Error: You appear to be using the `master` branch of emsdk.')
     errlog('We recently made the switch to using `main`')
     errlog('In order to continue to receive updates you will need to make the switch locally too.')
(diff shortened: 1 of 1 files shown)


sbc100 (approved):

kripken:
CI fails on the intended error as expected. I think this is one of the few cases where landing with errors is actually the right thing...

Any thoughts on the timing? As this is Friday, may we should wait until Monday to land it.

sbc100:
> CI fails on the intended error as expected. I think this is one of the few cases where landing with errors is actually the right thing...

Don't you need to update the test scripts to expect the error?   As it stands the failure is unexpected and prevents further test execution... Or are you saying that somehow after landing this change future PR's wont fail?  Maybe I misunderstand what you are saying?

> 
> Any thoughts on the timing? As this is Friday, may we should wait until Monday to land it.

sbc100:
Oh wait I see.. this is landing on master.. in that case yes please ignore failures.. we don't really need/want CI on master

PR #833, opened 2021-06-05
bugfix: allow to install emscripten only

fs-eire:
add a check in function `expand_sdk_name()`: if name is a known tool or sdk, skip the expand.

this fixes the wrong behavior of emsdk:

The command `emsdk install releases-upstream-<commit>-64bit` should install emscripten only, but `expand_sdk_name()` rewrite the args so it behaves same as `emsdk install sdk-releases-upstream-<commit>-64bit`.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2770,6 +2770,8 @@ def exit_with_fastcomp_error():
 
 
 def expand_sdk_name(name, activating):
+  if find_tool(name) or find_sdk(name):
+    return name
   if 'upstream-master' in name:
     errlog('upstream-master SDK has been renamed upstream-main')
     name = name.replace('upstream-master', 'upstream-main')
@@ -2984,13 +2986,13 @@ def main(args):
     global TTY_OUTPUT
     TTY_OUTPUT = False
 
+  load_dot_emscripten()
+  load_sdk_manifest()
+
   # Replace meta-packages with the real package names.
   if cmd in ('update', 'install', 'activate'):
     activating = cmd == 'activate'
     args = [expand_sdk_name(a, activating=activating) for a in args]
-
-  load_dot_emscripten()
-  load_sdk_manifest()
 
   # Apply any overrides to git branch names to clone from.
   forked_url = extract_string_arg('--override-repository')


fs-eire:
looks like this approach does not work because it breaks the extra_release_tag scenario. replace this by #834, an alternative solution.

sbc100:
My understanding is that anything with "releases" in the name installs the entire SDK and always has done.

What makes you think that `emsdk install releases-upstream-<commit>-64bit` should install emscripten only?

The only way I know of to install emscripten (at least any modern version of emscripten) on its own is via `emscripten-main-64bit` which will install the latest main branch of emscripten.

sbc100:
I guess there is the "tool",  `releases-upstream-77b065ace39e6ab21446e13f92897f956c80476a-64bit` and the "sdk"   `sdk-releases-upstream-77b065ace39e6ab21446e13f92897f956c80476a-64bit`,

but the tool in question is build from the emscripten-releases repo and it includes llvm and binaryn and emscripten all in a single tarball.  We have been shipping emsdk in this way for several years now.

fs-eire:
> I guess there is the "tool", `releases-upstream-77b065ace39e6ab21446e13f92897f956c80476a-64bit` and the "sdk" `sdk-releases-upstream-77b065ace39e6ab21446e13f92897f956c80476a-64bit`,
> 
> but the tool in question is build from the emscripten-releases repo and it includes llvm and binaryn and emscripten all in a single tarball. We have been shipping emsdk in this way for several years now.

There is no concern on the emscripten package. the concern is about nodejs, python and java installed with emscripten.

I want to install emscripten by emsdk, without installing nodejs, python and java. there are multiple reasons why I want to do this: node.js 14.15.5 has a security alert with its NPM dependency y18n 4.0.0 (although I believe this is not in use), and for Java there is a license requirement for our code base so openjdk is preferred.

The problem is, apparently emsdk is designed to support installing a single tool. for example, we can use emsdk to install node, python or java separately. The reason why it does not work is a bug in the code which rewrite "releases-*" into "sdk-releases-*".

as explained in #834, this may not be very easy to fix, so #834 is like a workaround instead of a solid fix.

PR #834, opened 2021-06-05
bugfix: allow to install SDK binaries alone

fs-eire:
an alternative approach to #833.

> this fixes the wrong behavior of emsdk:

> The command `emsdk install releases-upstream-<commit>-64bit` should install emscripten only, but `expand_sdk_name()` rewrite the args so it behaves same as `emsdk install sdk-releases-upstream-<commit>-64bit`.

as #833 broke extra_release_tag, this change uses a more straightforward way: if name starts with "releases-", we treat this as a tool instead of sdk.

The current code is bug prone (for example, if a tool's name is exactly length of 40, it will be treated as a "version" of SDK), but I don't think there is a easy fix as several places depends on global variable extra_release_tag. the ideal fix is to have a strict definition of whether a name is a tool or sdk, and perform a check based on that. considering backward compatibility this may be a hard work.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2812,6 +2812,7 @@ def expand_sdk_name(name, activating):
     fullname = fullname.replace('-fastcomp', '')
     backend = 'fastcomp'
   version = fullname.replace('sdk-', '').replace('releases-', '').replace('-64bit', '').replace('tag-', '')
+  sdk = 'sdk-' if not name.startswith('releases-') else ''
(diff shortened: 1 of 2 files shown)


sbc100 (commented):
lgtm!

Can you re-title the PR though, and also rebase?

This is not about install emscripten on its own but about installing the SDK binaries alone (which includes llvm and binaryen).

sbc100 (approved):

PR #837, opened 2021-06-08
Improve emscripten-releases-tags.txt to support mutliple aliases

sbc100:
Also, improve reporting of version resolution. e.g.:

```
$ ./emsdk  install sdk-latest
Resolving SDK alias 'latest' to '2.0.23'
Resolving SDK version '2.0.23' to 'sdk-releases-upstream-77b065ace39e6ab21446e13f92897f956c80476a-64bit'
Installing SDK 'sdk-releases-upstream-77b065ace39e6ab21446e13f92897f956c80476a-64bit'..
...
```

Should help with #836.

### DIFF
--- a/emscripten-releases-tags.txt
+++ b/emscripten-releases-tags.txt
@@ -1,5 +1,13 @@
 {
-  "latest": "2.0.23",
+  "aliases": {
+    "latest": "2.0.23",
+    "latest-sdk": "latest",
+    "latest-64bit": "latest",
+    "sdk-latest-64bit": "latest",
+    "latest-upstream": "latest",
+    "latest-clang-upstream": "latest",
+    "latest-releases-upstream": "latest"
+  },
   "releases": {
     "2.0.23": "77b065ace39e6ab21446e13f92897f956c80476a",
     "2.0.23-lto": "3f6dbb899f61fab52e4574beb4f7c8382658aa20",
--- a/emsdk.py
+++ b/emsdk.py
@@ -2166,18 +2166,27 @@ def is_os_64bit():
   return platform.machine().endswith('64')
 
 
-def find_latest_releases_version():
+def find_latest_version():
+  return resolve_sdk_aliases('latest')
+
+
+def find_latest_hash():
+  version = find_latest_version()
   releases_info = load_releases_info()
-  return releases_info['latest']
+  return releases_info['releases'][version]
 
 
-def find_latest_releases_hash():
+def resolve_sdk_aliases(name, verbose=False):
   releases_info = load_releases_info()
-  return releases_info['releases'][find_latest_releases_version()]
+  while name in releases_info['aliases']:
+    if verbose:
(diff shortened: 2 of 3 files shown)


sbc100:
@RReverser

emsdk.py
```diff
@@ -2775,12 +2784,10 @@ def expand_sdk_name(name, activating):
     name = name.replace('upstream-master', 'upstream-main')
   if name in ('latest-fastcomp', 'latest-releases-fastcomp', 'tot-fastcomp', 'sdk-nightly-latest'):
     exit_with_fastcomp_error()
-  if name in ('latest', 'sdk-latest', 'latest-64bit', 'sdk-latest-64bit'):
-    # This is effectly the default SDK
-    return str(find_latest_releases_sdk('upstream'))
-  elif name in ('latest-upstream', 'latest-clang-upstream', 'latest-releases-upstream'):
-    return str(find_latest_releases_sdk('upstream'))
-  elif name in ('tot', 'sdk-tot', 'tot-upstream'):
+  if name in ('latest', 'sdk-latest', 'latest-64bit', 'sdk-latest-64bit', 'latest-upstream', 'latest-clang-upstream', 'latest-releases-upstream'):
```
RReverser, line 2787: I wonder if this could be part of the new `aliases` mechanism too - that is, allow alias to point to other aliases and make function resolve recursively.

The depth would always be only 1 or 2, but it would allow to encode `"sdk-latest": "latest"` etc. as regular alises in the same config.

Or maybe I'm overcomplicating things?

emscripten-releases-tags.txt
```diff
@@ -1,5 +1,7 @@
 {
```
RReverser, line 1: Btw, not that it matters, but why does this file use `.txt` extension rather than JSON?

kripken (approved):

emscripten-releases-tags.txt
sbc100: You are right.. we should rename them

emsdk.py
sbc100: I thought of that too.. but for some reason I found that to be ugly... I'll revisit.

kripken:
Interesting, what does `sbc100 dismissed kripken’s stale review via 04cf92b 25 minutes ago` mean? I've never seen that on github before. Did you do that manually or did github automatically do it due to a force-push?

kripken:
(And where is my review...? It can't be viewed. I think I approved this, but that is gone too...)

kripken (approved):
lgtm but I'd check locally/manually that the old aliases work (as I don't think we have automation for them all), to check for a typo.

sbc100:
> (And where is my review...? It can't be viewed. I think I approved this, but that is gone too...)

I certainly didn't intentionally dismiss or delete anyting.   I wonder if github changes the behaviour of force push.. (which I use a *lot*).

PR #838, opened 2021-06-09
Error out on attempt to activate a missing tools

sbc100:
Previously if a tool (any part of an SDK) was not installed
we would issue a warning and continue to active without returning
non-zero.

This meant doing `emsdk install 2.0.0 && emsdk activate latest`
would appear to be work aside from the warning messages about
latest not being installed.

This is especially annoying since we dropped support for side
by side SDK installations.  The following sequence is no longer
valid and we want to make that clear by erroring out:

```
$ emsdk install 2.0.1
$ emsdk install 2.0.2
$ emsdk activate 2.0.1
```

Since 2.0.2 replaces 2.0.1 on the filesystem the active here
could fail hard rather than just warning.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2497,19 +2497,6 @@ def can_simultaneously_activate(tool1, tool2):
   return tool1.id != tool2.id
 
 
-def remove_nonexisting_tools(tool_list, log_errors=True):
-  i = 0
-  while i < len(tool_list):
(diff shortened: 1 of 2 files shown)


RReverser (approved):

kripken (dismissed):
lgtm with a test

PR #839, opened 2021-06-09
2.0.24

sbc100:


sbc100:
Since Derek is OOO I suggest we just do a non-LTO release for now.  We can add an LTO version of this release later.. and make it the default in a future release?

kripken (approved):
Plan sgtm.

kripken:
However, looking at the roller, the emscripten one seems to be broken?

sbc100:
> However, looking at the roller, the emscripten one seems to be broken?

Does that matter?  It shouldn't stop use releases should it?

sbc100:
Actually maybe if we are 17 behind that is a lot.. let me check.

kripken:
It doesn't block in general, yeah, but weren't there changes you wanted to roll in for the release? Or did they make it before the breakage?

sbc100:
Uploaded new revision now that the roller is passing again

kripken:
It looks like the asan test suite started to fail right on that commit, so perhaps we should verify if that is a test suite problem or something more serious first? I can take a look in a bit.

sbc100:
Looks like it was #14399 that starting asan.test_utf8_textdecoder failing.

sbc100:
If its just a bug in the test I suggest we go ahead with this release.. if its a bug in the library code I guess we need to wait for a fix.

kripken:
I verified it is a bug in the test. The bug pre-existed before my commit, but the commit unsurfaced it. Anyhow, lgtm for this PR, nothing to worry about, there is no library code bug.

PR #843, opened 2021-06-16
Point zsh and csh users to the correct startup script

sbc100:
Fixes: https://github.com/emscripten-core/emscripten/issues/14446

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1663,8 +1663,14 @@ JS_ENGINES = [NODE_JS]
       print('    ' + p)
     print('- This can be done for the current shell by running:')
     print('    source "%s"' % emsdk_env)
-    print('- Configure emsdk in your bash profile by running:')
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -1663,8 +1663,14 @@ def generate_dot_emscripten(active_tools):
       print('    ' + p)
     print('- This can be done for the current shell by running:')
     print('    source "%s"' % emsdk_env)
-    print('- Configure emsdk in your bash profile by running:')
-    print('    echo \'source "%s"\' >> $HOME/.bash_profile' % emsdk_env)
+    print('- Configure emsdk in your shell startup scripts by running:')
+    shell = os.environ.get('SHELL', '')
+    if 'zsh' in shell:
```
kripken, line 1668: should these perhaps be `shell.endswith('/zsh')`? but I guess it's almost always `/bin/X` anyhow...

kripken (approved):

PR #849, opened 2021-06-28
Support linking with `-o filename`

attilaolah:
When linking with `-o filename` (such as in various CMake build checks), the parameter passed to the linker is a temporary file, and it is passed as a bare filename (i.e. relative path without a `'/'`). In such cases, `outdir` would have been the empty string, and the final `tar` command would fail (actually the call to `subprocess.check_call(…)` is what fails).


attilaolah:
@walkingeyerobot please take a look.

walkingeyerobot (approved):

PR #904, opened 2021-09-28
bazel: pass -g3 instead of -g4 for wasm_asan feature

### LINKED ISSUE #903: wasm_asan bazel feature is broke.
In the file: 

  bazel/emscripten_toolchain/crosstool.bzl
  
The following code sets the command line arguments for the feature "wasm_asan".

```skylark
flag_set(
            actions = all_compile_actions +
                      all_link_actions,
            flags = [
                "-g4",
                "-fsanitize=address",
                "-O1",
                "-DADDRESS_SANITIZER=1",
                "-fno-omit-frame-pointer",
            ],
            features = ["wasm_asan"],
        ),
        
```
When the feature is enabled, the build reports:

  emcc: error: please replace -g4 with -gsource-map [-Wdeprecated] [-Werror]

It appears the fix is to just swap the -g4 in the flags with -gsource-map

It appears as if any automated testing of the bazel rules and toolchains did not pick this up.

walkingeyerobot:
fixes #903

### DIFF
--- a/bazel/emscripten_toolchain/crosstool.bzl
+++ b/bazel/emscripten_toolchain/crosstool.bzl
@@ -619,7 +619,7 @@ def _impl(ctx):
             actions = all_compile_actions +
                       all_link_actions,
             flags = [
-                "-g4",
+                "-g3",
                 "-fsanitize=address",
                 "-O1",
                 "-DADDRESS_SANITIZER=1",


sbc100 (approved):
Why not just `-g`?

daveb68 (commented):
I should have mentioned this in the original issue report.  

-g4 is also used in other places in this file.   For example, 12 lines below with the wasm_ubsan feature.  I did not test this, but assume it would have the exact same problem.

PR #910, opened 2021-10-13
Add embind example to Bazel docs

kjlubick:
This also tidies up a few things in the README.


sbc100:
In the circleci UI there is a "Rerun" drop down in the top right.  If you don't see it I guess it could be some kind of permission isssue?

I selected "Re-run from failed"

PR #915, opened 2021-10-19
2.0.32

sbc100:


kripken:
The test failure may be real. Did we add a new `libstubs` library, but are not building it into the precompiled libs that ship with the emsdk perhaps?

PR #920, opened 2021-10-27
Forward Bazel platforms from CLI onward

kjlubick:
I was very confused why the platforms I was specifying on the CLI were not getting passed through the `wasm_cc_binary` rule. For the Skia project, we make use of [platforms and constraint values](https://skia-review.googlesource.com/c/skia/+/463517/4/bazel/common_config_settings/BUILD.bazel#51) to define things like WebGL vs WebGPU support, so we those to get passed through to our cc_library rules and such, so we can build the correct supporting code.

If there was a good reason to use empty list for platform, that should probably be documented. If we, the Skia team, should not be using platforms like that, I'd like to hear what to use instead.

### DIFF
--- a/bazel/emscripten_toolchain/wasm_cc_binary.bzl
+++ b/bazel/emscripten_toolchain/wasm_cc_binary.bzl
@@ -6,6 +6,7 @@ def _wasm_transition_impl(settings, attr):
 
     features = list(settings["//command_line_option:features"])
     linkopts = list(settings["//command_line_option:linkopt"])
+    platforms = list(settings["//command_line_option:platforms"])
 
     if attr.threads == "emscripten":
         # threads enabled
@@ -32,7 +33,7 @@ def _wasm_transition_impl(settings, attr):
         "//command_line_option:features": features,
         "//command_line_option:dynamic_mode": "off",
         "//command_line_option:linkopt": linkopts,
-        "//command_line_option:platforms": [],
+        "//command_line_option:platforms": platforms,
         "//command_line_option:custom_malloc": "@emsdk//emscripten_toolchain:malloc",
     }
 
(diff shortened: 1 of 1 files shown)


trybka:
This is a tricky one. In some sense, the platform should be set to wasm (whatever your wasm platform is) since everything under the wasm_cc_binary would be thought of to be building for wasm. 

No other platform makes sense here. We could add a hook to set the appropriate wasm platform for your repo, but I think it would be incorrect to pass the current platform on through.

kjlubick:
Is it the case that I'm misusing platforms then to turn on and off things like "which gpu backend to use" or "Do you want to include the JPEG codec"? Should I use some other bazel-ism?

The reason I went with platforms and constraint_settings was that they work with select(). I wasn't sure how to use something like Bazel "features" with selects() (which I notice *are* passed along in the build).

trybka:
[Platforms](https://docs.bazel.build/versions/main/platforms.html), to my understanding, are intended to encode features and capabilities of the logical target system itself. They are a bag of `constraint_settings`, and I think the intent is largely to use generic `constraint_settings` (at least in rules like `wasm_cc_binary`).

If the decision to use JPEG codecs were based on some capability of the `platform` (e.g. has SIMD support) then you likely want to `select()` on that rather than encode your build choices in the platform itself. 

If you envision wanting to build, say, Wasm with and without JPEG codecs, then it would likely make more sense as a `build_setting` (which is also `select()`-able).

(see https://docs.bazel.build/versions/main/skylark/config.html#build-settings-and-select)

kjlubick:
Thanks for the detailed response. I'll re-work my rules to use build_settings instead of platforms. This PR need not be merged.

PR #929, opened 2021-11-13
Bazel windows support

### LINKED ISSUE #814: Bazel toolchain does not work on Windows
The Bazel toolchain does not work on Windows. There are, as far as I can tell, two reasons for this, although there may be more:
1. The Windows URLs are incorrect. The actual files end in `.zip` (e.g. https://storage.googleapis.com/webassembly/emscripten-releases-builds/win/89202930a98fe7f9ed59b574469a9471b0bda7dd/wasm-binaries.zip) (Thanks, @walkingeyerobot, for pointing me to them!) but they're [listed as `.tar.bz2`](https://github.com/emscripten-core/emsdk/blob/main/bazel/emscripten_deps.bzl#L54) in `emscripten_deps.bzl`. This can be fixed by changing the `type` argument passed to `http_archive` to `"zip"` and the formatting of the [`emscripten_url` variable](https://github.com/emscripten-core/emsdk/blob/main/bazel/emscripten_deps.bzl#L22) to allow for replacing the file extension with `.zip`.
2. The bash scripts used for Mac and Linux don't run on Windows. The fix for this is likely more difficult since it probably requires equivalent powershell scripts to be written (or maybe there's some way to have Bazel download bash for Windows).

Here's the output of building the hello world demo after applying the fix described in 1. Running in Git bash.
```
$ ~/AppData/Local/Yarn/bin/bazel build :hello-world-wasm
INFO: Analyzed target //hello-world:hello-world-wasm (202 packages loaded, 18554 targets configured).
INFO: Found 1 target...
ERROR: C:/users/matthew/projects/emsdk/bazel/hello-world/BUILD:4:10: Compiling hello-world/hello-world.cc failed: (Exit -1
): emcc.sh failed: error executing command external/emsdk/emscripten_toolchain/emcc.sh --sysroot=external/emscripten_bin_w
in/emscripten/cache/sysroot -fdiagnostics-color -fno-exceptions -fno-strict-aliasing -funsigned-char ... (remaining 32 arg
ument(s) skipped). Note: Remote connection/protocol failed with: execution failed
Action failed to execute: java.io.IOException: ERROR: src/main/native/windows/process.cc(202): CreateProcessW("C:\users\ma
tthew\_bazel_matthew\2zysnvbj\execroot\emsdk\external\emsdk\emscripten_toolchain\emcc.sh" --sysroot=external/emscripten_bi
n_win/emscripten/cache/sysroot -fdiagnostics-color -fno-exceptions -fno-strict-aliasing -funsigned-char -no-canonical-pref
ixes -std=gnu++17 -nostdinc -nostdinc++ -fomit-frame-pointer -O2 -Wall -iquote . -iquote bazel-out/wasm-fastbuild-ST-95dcf
5f511a2/bin -iquote external/emsdk -iquote bazel-out/wasm-fastbuild-ST-95dcf5f511a2/bin/external/emsdk -iwithsysroot/(...)
): %1 is not a valid Win32 application.
 (error: 193)
Target //hello-world:hello-world-wasm failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 146.715s, Critical Path: 0.11s
INFO: 8 processes: 7 internal, 1 local.
FAILED: Build did NOT complete successfully
```

zaucy:
closes #814 

* Updated `scripts/update_bazel_workspace.sh` to calculate sha from `.zip` archive for windows
* Adds `.bat` file equivalent scripts for use on windows
* Removes use of `cp` and `file` commands in favour of python modules
  * Introduces [`filetype`](https://pypi.org/project/filetype/) dependency for `wasm_binary.py` instead of relying on `file`


sbc100 (commented):
Nice one!

Mostly makes sense to me.  The `pip install` thing seems a little heavy weight, but if its just one once on first use I guess its fine.

Will leave the final lgtm to @walkingeyerobot

bazel/emscripten_toolchain/crosstool.bzl
```diff
@@ -1099,12 +1105,22 @@ def _impl(ctx):
...
     attrs = {
         "cpu": attr.string(mandatory = True, values = ["asmjs", "wasm"]),
         "em_config": attr.label(mandatory = True, allow_single_file = True),
         "emscripten_binaries": attr.label(mandatory = True),
+        "is_windows": attr.bool(mandatory = True),
     },
     provides = [CcToolchainConfigInfo],
 )
+
+def emscripten_cc_toolchain_config_rule(**kwargs):
```
trybka, line 1119: this seems unnecessary to hide in a macro and somewhat obscures its purpose. 

could you just make an attribute `script_extension` which is set in BUILD.bazel and uses a similar select to set it?

zaucy: Sorry if I'm misunderstanding you, but wouldn't having a `script_extension` attribute put the burden of setting it on the user? The extension used (`.bat` vs `.sh`) is more or less an internal detail and less of a useful configuration option in my eyes.

Do you know of a way to set the attributes default similar to what the select is doing? That could be a nice middle ground. AFAIK attributes don't allow defaults to be set with select (https://github.com/bazelbuild/bazel/issues/287).

This strategy for checking for windows is used in similar fashion in [rules_nodejs](https://github.com/bazelbuild/rules_nodejs) and [bazel skylib](https://github.com/bazelbuild/bazel-skylib). e.g. https://github.com/bazelbuild/rules_nodejs/blob/aa09b5796faaf2470b84246a8f00e6e27181c63b/internal/common/copy_to_bin.bzl#L40-L65

trybka: > wouldn't having ... attribute put the burden of setting on the user?

I don't think so--unless they typically modify the BUILD.bazel that also ships in this repo? 

You're right, attributes can't be set with a select default. I think the "correct" way to handle this would be to make this part of a `toolchain()` that bazel selects for us, but I'm not going to push for that. 

Given that this is how rules_nodejs and bazel skylib do it though, I'm content to keep it consistent, so this is fine as-is.

zaucy: I took a look again and  realized use of `emscripten_cc_toolchain_config_rule` would be unlikely from users. I added the `script_extension` per your suggestion.

trybka: LGTM thanks

zaucy:
@walkingeyerobot I've rebased against the latest main branch and added a (untested) circle CI job for windows with bazel.

I didn't change `link_wrapper.py` to use the tarfile module only because I didn't have any trouble with it on my local windows machine, but I can make that change.

walkingeyerobot (approved):

zaucy:
I couldn't seem to find if there was a `python3.exe` somewhere in the windows vm. I was planning on looking again this weekend sometime.

zaucy:
@walkingeyerobot The bazel CI is passing now. I used `py -3` instead of `python3` which should be available for most windows python users. I also removed a hardcoded `python3` subprocess call and used `sys.executable` instead.

Hope those changes are acceptable :)

walkingeyerobot (approved):

PR #930, opened 2021-11-14
Optionally keep the downloaded archives

jmglogow:
I have been switching Emscripten versions quite a few times, and emsdk always downloads the binaries. So this adds the environment setting EMSDK_KEEP_DOWNLOADS, which keeps the downloaded archives around.

This also includes a documentation commit for some other EMSDK_* env-vars and one minor typo fix, which I didn't test, but seemed obviously broken.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -143,6 +143,9 @@ BUILD_FOR_TESTING = False
 # Other valid values are 'ON' and 'OFF'
 ENABLE_LLVM_ASSERTIONS = 'auto'
 
+# If true, keeps the downloaded archive files.
+KEEP_DOWNLOADS = bool(os.getenv('EMSDK_KEEP_DOWNLOADS'))
+
 
 def os_name():
   if WINDOWS:
@@ -1537,13 +1540,8 @@ def download_and_unzip(zipfile, dest_dir, download_even_if_exists=False,
   if not download_even_if_exists and num_files_in_directory(dest_dir) > 0:
     print("The contents of file '" + zipfile + "' already exist in destination '" + dest_dir + "', skipping.")
     return True
-  # Otherwise, if the archive must be downloaded, always write into the
-  # target directory, since it may be a new version of a tool that gets
-  # installed to the same place (that is, a different download name
-  # indicates different contents).
-  download_even_if_exists = True
 
-  received_download_target = download_file(url, zips_subdir, download_even_if_exists, filename_prefix)
+  received_download_target = download_file(url, zips_subdir, not KEEP_DOWNLOADS, filename_prefix)
   if not received_download_target:
     return False
   assert received_download_target == download_target
@@ -1554,9 +1552,9 @@ def download_and_unzip(zipfile, dest_dir, download_even_if_exists=False,
   if clobber:
     remove_tree(dest_dir)
   if zipfile.endswith('.zip'):
-    return unzip(download_target, dest_dir, unpack_even_if_exists=download_even_if_exists)
+    return unzip(download_target, dest_dir, unpack_even_if_exists=True)
   else:
-    return untargz(download_target, dest_dir, unpack_even_if_exists=download_even_if_exists)
+    return untargz(download_target, dest_dir, unpack_even_if_exists=True)
 
 
 def to_native_path(p):
@@ -2068,6 +2066,8 @@ class Tool(object):
     return True
 
   def cleanup_temp_install_files(self):
+    if KEEP_DOWNLOADS:
+        return
     url = self.download_url()
     if url.endswith(ARCHIVE_SUFFIXES):
       download_target = get_download_target(url, zips_subdir, getattr(self, 'zipfile_prefix', ''))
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -1525,7 +1528,7 @@ def build_binaryen_tool(tool):
   return success
 
 
-def download_and_unzip(zipfile, dest_dir, download_even_if_exists=False,
+def download_and_unzip(zipfile, dest_dir, unpack_even_if_exists=False,
```
sbc100, line 1531: Maybe not worth renaming this variable.  Its used to avoid both download and unpacking, so I'm not sure one name is better than than the other.    Maybe we could avoid the verb and just call it `ignore_existing` ?

emsdk.py
```diff
@@ -2026,14 +2024,12 @@ def install_tool(self):
...
       # TODO: explain the vs-tool special-casing
-      download_even_if_exists = (self.id == 'vs-tool')
       # The 'releases' sdk is doesn't include a verion number in the directory
       # name and instead only one version can be install at the time and each
       # one will clobber the other.  This means we always need to extract this
       # archive even when the target directory exists.
-      download_even_if_exists = (self.id == 'releases')
       filename_prefix = getattr(self, 'zipfile_prefix', '')
-      success = download_and_unzip(url, self.installation_path(), download_even_if_exists=download_even_if_exists, filename_prefix=filename_prefix)
+      success = download_and_unzip(url, self.installation_path(), (self.id == 'vs-tool') or (self.id == 'releases'), filename_prefix=filename_prefix)
```
sbc100, line 2032: Why no keep the sperate variable?  I think its make the code a little clearer, and to explains the `TODO: explain the vs-tool special-casing` above which.  Why do you special vs-tool and releases? 

Maybe just change the `=` to and `|=` in  `download_even_if_exists = (self.id == 'releases')` above?

sbc100: I guess this is one on the bugfixes that can be landed separately?  I also wonder if we should keep ths vs-tool special casing around given that is seems to have been inactive for a while now. 

In any case you split this change out into a separate PR?

emsdk.py
```diff
@@ -2265,7 +2263,7 @@ def update_emsdk():
   if is_emsdk_sourced_from_github():
     errlog('You seem to have bootstrapped Emscripten SDK by cloning from GitHub. In this case, use "git pull" instead of "emsdk update" to update emsdk. (Not doing that automatically in case you have local changes)')
     sys.exit(1)
-  if not download_and_unzip(emsdk_zip_download_url, emsdk_path(), download_even_if_exists=True, clobber=False):
+  if not download_and_unzip(emsdk_zip_download_url, emsdk_path(), True, clobber=False):
```
sbc100, line 2266: Its it not clear to keep the named parameter?    Especially for literal param like `True` and `False` I think it helps readability here to keep the name,

sbc100 (commented):
I like the intent here. I have often wished that zip archives would be cached and even tried to suggest a machanism a while back.

I think it would be good split out the seperate bug fixed into their own PRs.  Also I find the logic of `unpack_even_if_exists` and `download_even_if_existing` I little complex/convoluted.   I'm hoping we can simplify this somewhat.

I wonder if we need the `unpack_even_if_exists` arguments to untargz or unzip at all?  Are there any cases where we call these functions but we want them to early return?  I think maybe not.  (This could also be a separate change).

emsdk.py
jmglogow: I just kept the old name, which cuts this patch size a lot. Since the `download_even_if_exists` was always set to `True` originally, this is now really a `unpack_even_if_exists`, as the download now just depends on `KEEP_DOWNLOADS`, but the `unpack_even_if_exists` now isn't always True anymore.

jmglogow (commented):
Ok - seems one can't add a comment on line not changed by a commit, so lines 1554 - 1557 are:

    if zipfile.endswith('.zip'):
      return unzip(download_target, dest_dir, unpack_even_if_exists=download_even_if_exists)
    else:
      return unzip(download_target, dest_dir, unpack_even_if_exists=download_even_if_exists)

Maybe `download_even_if_exists` should now be True, like it was before?

jmglogow:
> Ok - seems one can't add a comment on line not changed by a commit, so lines 1554 - 1557 are:
> 
> ```
> if zipfile.endswith('.zip'):
>   return unzip(download_target, dest_dir, unpack_even_if_exists=download_even_if_exists)
> else:
>   return unzip(download_target, dest_dir, unpack_even_if_exists=download_even_if_exists)
> ```
> 
> Maybe `download_even_if_exists` should now be True, like it was before?

So I just pushed a rebase including this change, I think setting `unpack_even_if_exists=True` is more like the previous code. The end result is no difference, as the variables `unpack_even_if_exists` and `download_even_if_exists` trigger the same test, which returns early for `download_and_unzip` the same way then `unzip` and `untargz`:

    if not unpack_even_if_exists and num_files_in_directory(dest_dir) > 0:

So `unpack_even_if_exists=False` from `download_and_unzip` can just happen, if that branch wasn't taken anyway. YMMV.

jmglogow:
> I wonder if we need the `unpack_even_if_exists` arguments to untargz or unzip at all? Are there any cases where we call these functions but we want them to early return? I think maybe not. (This could also be a separate change).

Currently there are no other callers of `unzip` and `untargz` then `download_and_unzip` and with my adaption to set `unpack_even_if_exists=True` (which is the same code then before) there is no need for that variable. It's a bit like the

    assert received_download_target == download_target

in `download_and_unzip`. Nothing ever uses this path currently, so it could return just a boolean. And both `download_and_unzip` and `download_file` call `get_download_target`. The `download_target` is just used for that assert…

jmglogow:
Is there anything else missing to get this merged?

I just went over the comments and except for cleaning up more code in additional patches, which isn't part of this PR, there is nothing I could find.

sbc100 (approved):

PR #931, opened 2021-11-14
Remove all vs-tool references

jmglogow:
Untested, but the original code looked wrong.


sbc100:
Actually I did a little digging it looks like vs-tool was removed back in #395.  Can you just remove this line completely instead?

sbc100:
(All references to vs-tools can be removed).

sbc100 (approved):

PR #936, opened 2021-11-19
Always output a tarball from cc_binary to simplify logic.

walkingeyerobot:
This will change the result of --config=wasm builds that were previously outputting a single file.

Helps prepare for windows support (#814) and removes the need for #929 to add an external python dep.

### DIFF
--- a/bazel/emscripten_toolchain/link_wrapper.py
+++ b/bazel/emscripten_toolchain/link_wrapper.py
@@ -153,15 +153,11 @@ if os.path.exists(wasm_base + '.debug.wasm') and os.path.exists(wasm_base):
         wasm_base,
         '--add-section=external_debug_info=debugsection.tmp'])
 
-# If we have more than one output file then create tarball
-if len(files) > 1:
+# If we have one or more output files then create a tarball
+if len(files) >= 1:
   cmd = ['tar', 'cf', 'tmp.tar'] + files
   subprocess.check_call(cmd, cwd=outdir)
   os.rename(os.path.join(outdir, 'tmp.tar'), output_file)
-elif len(files) == 1:
-  # Otherwise, if only have a single output than move it to the expected name
-  if files[0] != os.path.basename(output_file):
-    os.rename(os.path.join(outdir, files[0]), output_file)
 else:
   print('emcc.py did not appear to output any known files!')
   sys.exit(1)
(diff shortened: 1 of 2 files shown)


bazel/emscripten_toolchain/link_wrapper.py
```diff
@@ -153,15 +153,11 @@
         wasm_base,
         '--add-section=external_debug_info=debugsection.tmp'])
 
-# If we have more than one output file then create tarball
-if len(files) > 1:
+# If we have one or more output files then create a tarball
+if len(files) >= 1:
```
sbc100, line 157: Can we just drop this `if/else` completely.  If you a worried that would loose coverage perhaps just `assert(len(files))` here?

sbc100:
Its maybe a little sad that `--config=wasm` a little less directly useful now since users will always need to untar the wasm file.     I guess it will push more folks towards using wasm_cc_binaryen?

bazel/emscripten_toolchain/link_wrapper.py
walkingeyerobot: Yeah, I would prefer to maintain the check. I don't expect it to be triggered, but we're not losing anything by keeping it, so I figure it's better to keep it.

sbc100: OK, but then at least re-write it to use early-return style and avoid the indentation:

```
if not len(files):
  print('emcc.py did not appear to output any known files!')
  sys.exit(1)

.. the rest here without any indentation
```

trybka:
> cc @trybka

LGTM (with sbc100's suggested tweak)

bazel/emscripten_toolchain/link_wrapper.py
walkingeyerobot: Done

sbc100:
Looks like an unrelated binaryen failure:
```
root/project/binaryen/main/src/passes/GlobalRefining.cpp: In member function 'virtual void wasm::{anonymous}::GlobalRefining::run(wasm::PassRunner*, wasm::Module*)':
/root/project/binaryen/main/src/passes/GlobalRefining.cpp:57:27: error: unused variable 'func' [-Werror=unused-variable]
     for (auto& [func, info] : analysis.map) 
```

sbc100:
I think you just need to rebase (or merge), no?

sbc100 (approved):

PR #940, opened 2021-11-26
Canonical optimization levels in bazel toolchain

bkotsopoulossc:
Bazel has three compilation modes, fastbuild, dbg and opt. Docs [here](https://docs.bazel.build/versions/main/user-manual.html#flag--compilation_mode).

Copying the definitions here for reference:
```
- fastbuild means build as fast as possible: generate minimal debugging information (-gmlt -Wl,-S), and don't optimize. This is the default. Note: -DNDEBUG will not be set.
- dbg means build with debugging enabled (-g), so that you can use gdb (or another debugger).
- opt means build with optimization enabled and with assert() calls disabled (-O2 -DNDEBUG). Debugging information will not be generated in opt mode unless you also pass --copt -g.
```

Currently in the emscripten bazel toolchain, dbg seems to be doing the right thing (`-O0 -g`). But fastbuild is optimizing at -O2 when it is supposed to be -O0. And opt is going up to -O3 instead of -O2. 

I believe what we want to achieve is the following:
`-c dbg` -> `-O0 -g` (-g implies -g3)
`-c fastbuild` or nothing, since fastbuild is the default -> `-O0 -g0`
`-c opt` -> `-O2 -g0 -NDEBUG`

`-c opt --features=optimized_for_speed`  -> `-O3 -g0 -NDEBUG`
`-c opt --features=optimized_for_size`  -> `-Oz -g0 -NDEBUG`
`-c opt --features=full_debug_info` -> `-O2 -g -NDEBUG`
etc etc.

Thoughts @sbc100 @celentes @walkingeyerobot?

### DIFF
--- a/bazel/emscripten_toolchain/crosstool.bzl
+++ b/bazel/emscripten_toolchain/crosstool.bzl
@@ -572,7 +572,7 @@ def _impl(ctx):
         flag_set(
             actions = all_compile_actions +
                       all_link_actions,
-            flags = ["-O3"],
+            flags = ["-O2"],
             features = ["opt"],
         ),
         # Users can override opt-level with semantic names...
@@ -598,7 +598,7 @@ def _impl(ctx):
         flag_set(
             actions = all_compile_actions +
                       all_link_actions,
-            flags = ["-O2"],
+            flags = ["-O0"],
             features = ["fastbuild"],
         ),
 
--- a/bazel/emscripten_toolchain/crosstool.bzl
+++ b/bazel/emscripten_toolchain/crosstool.bzl
@@ -575,6 +575,7 @@ def _impl(ctx):
             flags = ["-O2"],
             features = ["opt"],
         ),
+
         # Users can override opt-level with semantic names...
         flag_set(
             actions = all_compile_actions +


sbc100:
Is there any point in passing `-g0` or `-O0` since they are both the default?   Seems like just omiting them would be the simplest thing to do.

bkotsopoulossc:
@sbc100 I agree on not passing -g0. I'm also okay with omitting -O0 but it felt nicer to me to be explicit there

sbc100:
> @sbc100 I agree on not passing -g0. I'm also okay with omitting -O0 but it felt nicer to me to be explicit there

Lets just go with whatever the default bezel toolchain does for desktop builds.    I don't know what it does but we have no reason to do anything different.

bkotsopoulossc:
> > @sbc100 I agree on not passing -g0. I'm also okay with omitting -O0 but it felt nicer to me to be explicit there
> 
> Lets just go with whatever the default bezel toolchain does for desktop builds. I don't know what it does but we have no reason to do anything different.

I imagine it might be slightly different on different platforms and might be tricky to track down, but I’ll take a look at what it’s doing on my Mac toolchain, in terms of explicitly setting defaults vs omitting them, and I’ll update the PR. But in general, with this change it should match the main flags

bkotsopoulossc:
Checked on the default Mac toolchain.

- Looks like `-c dbg` sets just `-g` and doesn't touch `-O`.
- Looks like `-c fastbuild` sets `-O0` explicitly and doesn't touch `-g`

It seems arbitrary and I don't think it really matters whether we omit defaults or explicitly set them. As long as we do the right thing for the three modes, that seems like a good improvement

walkingeyerobot (approved):

bkotsopoulossc:
@walkingeyerobot it seems CI here is stuck, maybe it just needs a kick, thanks

sbc100:
Hmm.. strange.   I'm not sure what to do other than try re-basing/merging and push'ing again?

bkotsopoulossc:
Alright, I rebased and pushed. Let's see if CI wakes up. auto-merge also got disabled, it didn't like that i pushed :(

sbc100:
There seems to be some issue with circleci when new user contribute to emscripten to emsdk the jobs sometimes never start.  It should not be required the you need a login at circleci to run these jobs but this bug seems to mean that sometime we get blocked when you don't.  Would you might trying to log into circleci to see if that makes a differnce?  (no paid account should be needed).

bkotsopoulossc:
Looks like it's working now!

PR #941, opened 2021-11-30
Add closure example for wasm_cc_binary

kjlubick:
This does not work yet, but sketches out the pieces we need. I got stuck getting the file from the label and getting the cmd flag to be used in the env_set. Any advice?

If we aren't thrilled with adding @bazel_skylib as a dependency just for string_flag, that should be easy enough to write the Starlark ourselves.

@walkingeyerobot and @trybka for assistance.


trybka:
Can you give us more context on what you are trying to achieve? I feel I've missed something.

kjlubick:
I'm trying to figure out how to pass my externs file to closure using the Bazel emsdk rules. Without Bazel, I would set an environment variable, but I'm not sure how to do that. I see there are a few env vars set by the toolchain (e.g. EMCC_DEBUG), but those seem to be static values, and not something that could get passed in via an attr on wasm_cc_binary.
Am I missing something?

sbc100:
I would avoid environment variables completely and instead use the `--clusure-args` link flag

kjlubick:
That works! WDYT about me changing this PR to just be an example of using closure externs (and sending a separate PR to document --closure-args on https://emscripten.org/docs/tools_reference/emcc.html)?

kjlubick:
So, it does work on https://skia-review.googlesource.com/c/skia/+/478307, but when I try to replicate it, I'm seeing a very strange error:
```
building:ERROR: Closure compiler run failed:

building:ERROR: ERROR - [JSC_READ_ERROR] Cannot read file hello-embind-externs.js: hello-embind-externs.js
```
This is odd because I see the file in the sandbox when I add `--sandbox_debug`. The only difference I can think of is that we use Bazel 6 in the linked CL, and this is using Bazel 4.2

kjlubick:
Docs PR: https://github.com/emscripten-core/emscripten/pull/15683

kjlubick:
Interesting note: It was working in the linked PR because we were using version 2.0.13. When I try to move forward to 2.0.32, I see the same error I see here - my file appears to be missing even though I see it in the sandbox.

So it's not a Bazel version issue at all (but still a regression of some kind)

kjlubick:
I think the issue is https://github.com/emscripten-core/emscripten/commit/424ac8660b2b07ea882ca572fd7ff8f7780d52ed

After that commit, the closure command is run with a cwd of https://github.com/emscripten-core/emscripten/blob/e6ac12aaf5cdf218e7680114a97ad47c61030037/tools/building.py#L825 which means the passed in file (a relative file location) is not reachable from there.

kjlubick:
So, we need a way to pass in the absolute file name instead of the relative one. I tried looking up how to do that, but with no success. The execpath I tried did not work, it only gave me a relative path:
https://docs.bazel.build/versions/main/be/make-variables.html#predefined_label_variables

sbc100:
Oh.. we should support compiler-relative path names being passed in emscripten.. does that not work?   Can you open an emscripten bug for that?

sbc100:
This test looks like it uses a relative pathname: https://github.com/emscripten-core/emscripten/blob/424ac8660b2b07ea882ca572fd7ff8f7780d52ed/tests/test_other.py#L7471-L7476

trybka:
relative paths can be a bit finicky in Bazel.

Can you use $(location relative/path/externs.js) instead? 

https://docs.bazel.build/versions/main/be/general.html#general-advice

kjlubick:
I tried location. Instead of passing in `hello-embind-externs.js` to the closure compiler, it passed in `./hello-embind-externs.js` which did not really help.

kjlubick:
Should I wait on this PR until the non-absolute closure path fix is landed and a new release cut? Or should I try to get Bazel to give me an absolute path?

kjlubick:
I'm still running into issues here with absolute vs relative paths (even using emsdk 3.1.3). Using `$(location hello-embind-externs.js)` gives me a seemingly-correct relative path, but the file is not found when getting passed into node.

Of note, ` --externs=./hello-embind-externs.js` is the only relative externs path seen.

kjlubick:
I think I see the issue. [building.py](https://github.com/emscripten-core/emscripten/blob/main/tools/building.py#L887) does not appear to convert relative paths to absolute paths if the argument is in the form "--externs=foo.js" only if it is two arguments like "--externs", "foo.js".

kjlubick:
Welcome back! Waiting on v3.1.4 to be available so this PR works.

kjlubick:
PTAL - I added on to test_bazel.sh to prove the minification does not touch the function mentioned by the extern file.

walkingeyerobot (approved):

PR #951, opened 2021-12-20
Add test and remove unused args

sbc100:
Followup to #930: Add a tests, and remove newly-unused arguments
to unzip and untargz.

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -40,11 +40,11 @@ def check_call(cmd, **args):
   subprocess.check_call(cmd, **args)
 
 
-def checked_call_with_output(cmd, expected=None, unexpected=None, stderr=None):
+def checked_call_with_output(cmd, expected=None, unexpected=None, stderr=None, env=None):
   cmd = cmd.split(' ')
   print('running: %s' % cmd)
   try:
-    stdout = subprocess.check_output(cmd, stderr=stderr, universal_newlines=True)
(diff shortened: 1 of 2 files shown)


sbc100:
CC @jmglogow

test/test.py
```diff
@@ -272,6 +272,15 @@ def test_activate_missing(self):
     run_emsdk('install latest')
     failing_call_with_output(emsdk + ' activate 2.0.1', expected="error: tool is not installed and therefore cannot be activated: 'releases-upstream-13e29bd55185e3c12802bc090b4507901856b2ba-64bit'")
 
+  def test_keep_downloads(self):
+    env = os.environ.copy()
+    env['EMSDK_KEEP_DOWNLOADS'] = '1'
+    # With EMSDK_KEEP_DOWNLOADS the downloading should happen on the first
+    # install, but not the second time we install 5c776e6a
```
aheejin, line 279: What is 5c776e6a? (I can't see that commit both in emscripten and emsdk..) But anyway I guess it's better to change it to something like 2.0.28?

aheejin (approved):

test/test.py
sbc100: Oh sorry, the comment is out-of-date.  It should say `2.0.28` as you suggest.  I will update before landing.

PR #953, opened 2021-12-21
Minor node version update v14.15.5 -> v14.18.2

sbc100:
The newer versions of eslint require 14.17.0 or above.  This
updates our node version to the latest in the 14.XX series.

I don't expect any user-visible changes.

### DIFF
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -258,6 +258,53 @@
     "activated_env": "EMSDK_NODE=%installation_dir%/bin/node%.exe%",
     "is_old": true
   },
+
+
+  {
+    "id": "node",
+    "version": "14.18.2",
+    "bitness": 32,
+    "arch": "x86",
+    "windows_url": "node-v14.18.2-win-x86.zip",
+    "activated_path": "%installation_dir%/bin",
+    "activated_cfg": "NODE_JS='%installation_dir%/bin/node%.exe%'",
+    "activated_env": "EMSDK_NODE=%installation_dir%/bin/node%.exe%"
+  },
+  {
+    "id": "node",
+    "version": "14.18.2",
+    "arch": "arm",
+    "bitness": 32,
+    "linux_url": "node-v14.18.2-linux-armv7l.tar.xz",
+    "activated_path": "%installation_dir%/bin",
+    "activated_cfg": "NODE_JS='%installation_dir%/bin/node%.exe%'",
+    "activated_env": "EMSDK_NODE=%installation_dir%/bin/node%.exe%"
+  },
+  {
+    "id": "node",
+    "version": "14.18.2",
+    "bitness": 64,
+    "arch": "x86_64",
+    "macos_url": "node-v14.18.2-darwin-x64.tar.gz",
+    "windows_url": "node-v14.18.2-win-x64.zip",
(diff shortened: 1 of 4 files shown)


emsdk_manifest.json
```diff
@@ -260,43 +260,43 @@
...
+    "version": "14.18.2",
     "arch": "aarch64",
     "bitness": 64,
-    "macos_url": "node-v14.15.5-darwin-x64.tar.gz",
-    "linux_url": "node-v14.15.5-linux-arm64.tar.xz",
+    "macos_url": "node-v14.18.2-darwin-x64.tar.gz",
+    "linux_url": "node-v14.18.2-linux-arm64.tar.xz",
     "activated_path": "%installation_dir%/bin",
     "activated_cfg": "NODE_JS='%installation_dir%/bin/node%.exe%'",
     "activated_env": "EMSDK_NODE=%installation_dir%/bin/node%.exe%"
```
juj, line 302: We should not change the existing declarations from 14.15.5 to 14.18.2, but instead add new declarations. Otherwise if people have installer batch scripts that do `emsdk install node-14.15.6-64bit`, they will break.

emsdk_manifest.json
```diff
@@ -655,60 +655,60 @@
...
+    "uses": ["node-14.18.2-64bit", "python-3.9.2-1-64bit", "releases-fastcomp-%releases-tag%-64bit"],
     "os": "macos",
     "arch": "aarch64",
     "custom_install_script": "emscripten_npm_install"
   },
   {
     "version": "releases-fastcomp-%releases-tag%",
     "bitness": 64,
-    "uses": ["node-14.15.5-64bit", "python-3.7.4-pywin32-64bit", "java-8.152-64bit", "releases-fastcomp-%releases-tag%-64bit"],
+    "uses": ["node-14.18.2-64bit", "python-3.7.4-pywin32-64bit", "java-8.152-64bit", "releases-fastcomp-%releases-tag%-64bit"],
```
juj, line 711: These changes retroactively change old releases to begin to reinstall with new node version, making it no longer possible to install old release the way that hey were. emsdk has (or used to have) a version filter that allowed one to make declarations like `releases >= version number` and `releases < version number` to accommodate for this, so that old releases would not retroactively mutate on a change like this.

Though I don't personally care about this issue happening (we don't use the emsdk releases tags) - but some years ago we had users who raised this as an issue, which led to creating the version >= and < filters. Not sure if that is relevant anymore tbh.

juj (changes requested):
Let's keep both the 14.15.5 and 14.18.2 declarations in the manifest, so people can keep explicitly installing the earlier node version?

kripken (commented):
lgtm with @juj's comments

emsdk_manifest.json
sbc100: Done

emsdk_manifest.json
sbc100: I think this is such a minor update that the risks of it breaking an older release are very low so I suggest we keep it simple and just update across the board here.

sbc100:
Kept the old version.   Can you unblock?

juj (approved):
Great!

PR #964, opened 2022-01-08
Remove `download_even_if_exists` argument to download_and_unzip

sbc100:
These days this argument really means `install_even_if_directory_exists`
(at least since #9300).

However, by the time we call `download_and_unzip` we have already checked
that `is_installed()` is false so we know we want to install for sure.
If the installation directory already existed and contained the correct
contents we would never get as far as `download_and_unzip`.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1520,18 +1520,11 @@ def build_binaryen_tool(tool):
   return success
 
 
-def download_and_unzip(zipfile, dest_dir, download_even_if_exists=False,
-                       filename_prefix='', clobber=True):
+def download_and_unzip(zipfile, dest_dir, filename_prefix='', clobber=True):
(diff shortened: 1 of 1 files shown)


kripken (commented):
I believe you are right. The relevant line is this though:
```
if self.is_installed() and not hasattr(self, 'git_branch'):
```
So components using a git branch might be broken by this? Do we still use that?

sbc100:
]> I believe you are right. The relevant line is this though:
> 
> ```python
> if self.is_installed() and not hasattr(self, 'git_branch'):
> ```
> 
> So components using a git branch might be broken by this? Do we still use that?

Tools using `git_branch` are by definition now using `download_and_unzip` (they download via git).   See: 


https://github.com/emscripten-core/emsdk/blob/2c09626c34d4778ca648992146c01454ea508ce7/emsdk.py#L2002-L2011

kripken (approved):
Ah, thanks, that's what I was missing.

PR #968, opened 2022-01-15
Fix building rules_foreign_cc targets in Bazel.

PiotrSikora:

### DIFF
--- a/bazel/emscripten_toolchain/env.sh
+++ b/bazel/emscripten_toolchain/env.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-export ROOT_DIR=`(pwd -P)`
+export ROOT_DIR=${EXT_BUILD_ROOT:-$(pwd -P)}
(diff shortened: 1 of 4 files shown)


bazel/emscripten_toolchain/env.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-export ROOT_DIR=`(pwd -P)`
+export ROOT_DIR=${EXT_BUILD_ROOT:-`pwd -P`}
```
sbc100, line 3: Maybe switch to `$(pwd -P)`  synctax to match the changes above?

PiotrSikora: Done.

walkingeyerobot (approved):

PiotrSikora:
I've just noticed that there are also `.bat` versions of those scripts. Could someone with access to Windows machine make the corresponding changes to them? Thanks!

PR #969, opened 2022-01-18
Don't advertise "per_object_debug_info" feature in Bazel.

PiotrSikora:
It breaks @emsdk in workspaces that have this feature enabled.

The following error appears when using the latest commit:

    cc_toolchain '@emsdk//emscripten_toolchain:everything' with
    identifier 'emscripten-wasm' doesn't define a tool path for 'dwp'

The following error appears after adding `emdwp` tool and scripts:

    output '*.dwo' was not created

### DIFF
--- a/bazel/emscripten_toolchain/crosstool.bzl
+++ b/bazel/emscripten_toolchain/crosstool.bzl
@@ -298,10 +298,6 @@ def _impl(ctx):
         # Blaze also tests if this feature is supported, before setting the "pic" build-variable.
         feature(name = "pic"),
 
-        # Blaze requests this feature if fission is requested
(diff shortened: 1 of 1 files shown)


PiotrSikora:
cc @walkingeyerobot

sbc100:
Interesting..  should we also try to fix this on the emscripten side?   Could you open a bug in emscripten about emdwp not doing what its seems to be expected to do? 

.. also we should probably change all those `blaze` comments to `bazel` .. but thats a different issue.

PiotrSikora:
@sbc100 Done (https://github.com/emscripten-core/emscripten/issues/16057).

walkingeyerobot (approved):

PR #974, opened 2022-01-20
Add `git submodule update` when doing a git checkout

sbc100:
This is required to build binaryen since
https://github.com/WebAssembly/binaryen/pull/4466

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -812,7 +812,7 @@ def git_clone(url, dstpath):
   if GIT_CLONE_SHALLOW:
     git_clone_args += ['--depth', '1']
   print('Cloning from ' + url + '...')
-  return run([GIT(), 'clone'] + git_clone_args + [url, dstpath]) == 0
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -838,6 +838,7 @@ def git_checkout_and_pull(repo_path, branch_or_tag):
       ret = run([GIT(), 'merge', '--ff-only', 'origin/' + branch_or_tag], repo_path)
     if ret != 0:
       return False
+    run([GIT(), 'submodule', 'update', '--init'], repo_path, quiet=True)
```
tlively, line 841: Alternatively, we could add `--recurse-submodules` (and conditionally `--shallow-submodules`) to the clone step. If not, we should probably add `--recursive` and conditionally `--depth=1` if `GIT_CLONE_SHALLOW` here.

sbc100: Good idea.  I added `--recurse-submodules` to both clone and checkout step.  Sadly the GIT_CLONE_SHALLOW option is not tested anywhere.. and even building from source at all is barely tested.

tlively (approved):

PR #975, opened 2022-01-27
Rebuild MacOS python with lzma module installed

sbc100:
Fixes: https://github.com/emscripten-core/emscripten/issues/16073


juj (approved):
The change to drop the revision number will break users' build scripts, though I am fine with that, since it will be an easy fix.

PR #978, opened 2022-01-29
Support Bazel builds on Apple silicon

mattsoulanille:
Add `macos_arm64` as a supported platform.

### DIFF
--- a/bazel/WORKSPACE
+++ b/bazel/WORKSPACE
@@ -4,6 +4,11 @@ load(":deps.bzl", "deps")
 
 deps()
 
+load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories")
+node_repositories(
+    node_version = "16.6.2",
+)
+
 load(":emscripten_deps.bzl", "emscripten_deps")
 
 emscripten_deps()
--- a/bazel/WORKSPACE
+++ b/bazel/WORKSPACE
@@ -4,11 +4,6 @@ load(":deps.bzl", "deps")
 
 deps()
 
-load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories")
-node_repositories(
-    node_version = "16.6.2",
-)
-
 load(":emscripten_deps.bzl", "emscripten_deps")
 
 emscripten_deps()
--- a/bazel/revisions.bzl
+++ b/bazel/revisions.bzl
@@ -12,6 +12,7 @@ EMSCRIPTEN_TAGS = {
         hash = "6626e25d6d866cf283147ca68d54ac9326fe399f",
         sha_linux = "4fb53364a2ba1de8978445aa26b2204bfd215b41da5d7df04f231040b197010a",
         sha_mac = "a8e347accb1ff402d96a128912ac8cda1731611c9f89095fee0ad39a6a18bbc3",
+        sha_mac_arm64 = "4374f5c852d0403b0a3b0e9dc8a3856a340e9d82ecf0f20aa8b36c6179d31fc8",
         sha_win = "e96f6ab8252fefa42f461676311d4c4e2d96fdc2e876ece07d9d7a49ef31aef0",
     ),
     "3.1.1": struct(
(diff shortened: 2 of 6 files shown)


mattsoulanille:
I would have added an Apple silicon presubmit test, but it doesn't look like CircleCI supports it yet.

bazel/WORKSPACE
```diff
@@ -4,6 +4,11 @@ load(":deps.bzl", "deps")
 
 deps()
 
+load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories")
+node_repositories(
+    node_version = "16.6.2",
```
mattsoulanille, line 9: Bumped node to 16 since it's the [first version that supports Apple silicon](https://github.com/nodejs/TSC/issues/886).

mattsoulanille: I moved this to `emscripten_deps.bzl`.

BertsLifee (approved):

walkingeyerobot (changes requested):
Thanks for this! Overall looks great. The only thing missing is that scripts/update_bazel_workspace.sh needs to be updated so when emsdk updates it's easy for maintainers to get the new hash into revisions.bzl.

Also sorry it took me so long to review; I took a 2-week vacation :sweat_smile:

mattsoulanille:
Thanks @walkingeyerobot! I just updated the Bazel script to include mac arm shas. No worries on the review time; I hope you had a good vacation!

walkingeyerobot (approved):

bazel/revisions.bzl
```diff
@@ -6,12 +6,14 @@ EMSCRIPTEN_TAGS = {
...
         sha_linux = "8b840819eb88f9178c11bad25859ce448a0559e485823a863a6add21380636ca",
         sha_mac = "0cb3f9bfbcc744233eae9d20036155738409405eacf8a3d4f9beefc5919d809a",
+        sha_mac_arm64 = "ee2772f380419df17d154e00388a16bcddc78c7af035c16a2ee534d6ecf099aa",
         sha_win = "c0549e1dbaa581ae66934c38beebd4250cd450cc2778e9a602cd9431bc81bc37",
     ),
     "3.1.2": struct(
         hash = "6626e25d6d866cf283147ca68d54ac9326fe399f",
         sha_linux = "4fb53364a2ba1de8978445aa26b2204bfd215b41da5d7df04f231040b197010a",
         sha_mac = "a8e347accb1ff402d96a128912ac8cda1731611c9f89095fee0ad39a6a18bbc3",
+        sha_mac_arm64 = "4374f5c852d0403b0a3b0e9dc8a3856a340e9d82ecf0f20aa8b36c6179d31fc8",
```
bkotsopoulossc, line 16: Is it possible to backfill some of these older versions between 3.1.2 and 2.0.33?

sbc100: Most likely there were never actually built by the builders.. and building them would be a fair amount of work for someone.

We generally really on the waterfall to produce all our builds for us: https://ci.chromium.org/p/emscripten-releases

Sine x64 binaryen happen to run fine on M1 silicon (albient slightly slower), one possible solution would be to simply copy the `sha_mac` for all the older releases.     Can you confirm that this works?   And if so maybe you could open a PR to duplicate the `sha_mac` lines for all older releases?

bkotsopoulossc: I don't know if it will be that simple, I think we'd have to tell bazel to use the x86 arch. I also wonder if we could refactor the code a bit so that it doesn't fail on the missing arm links. The issue I am seeing is if I use a new emsdk version but try to force an old emscripten version, it fails saying the arm url is not defined, even though I'm not building for arm

sbc100: I'm not sure what you mean can you just point `sha_mac_arm64` at the tar archive for `sha_mac`?  Why wouldn't that work.  The x86 executables should run completely transparently on M1, right?

sbc100: Wait.. are you not running on M1 hardware?   The `sha_mac_arm64` archive is for use on M1 hardward.. it has nothing to do with targeting arm (emscripten doesn't target arm or x86 .. only wasm)

bkotsopoulossc: I'm aware it's for M1 only. I am not running on M1. I am simply trying to use older versions from newer emsdk releases and they fail because the arm url is not defined. Copying the URLs might work, I just don't have a machine to try it. If it doesn't work, we should be able to shuffle the bazel scripts to avoid failing because arm URLs aren't defined, if I am just building on x86

sbc100: Sure... the missing arm build should not block installation on non-arm.  Can you open a new bug about that?

Issue #980, opened 2022-01-30
macOS Python doesn't run -  dyld issue, /usr/local/opt/gettext/lib/libintl.8.dylib not found


tangobravo:
It looks like the current python-3.9.2-64bit for macOS was linked on a system with gettext installed via brew or macports. The binary doesn't run for me, on a pretty vanilla M1 Pro MacBook Pro on the latest 12.1 macOS.

```
% python/3.9.2_64bit/bin/python3
dyld[64115]: Library not loaded: /usr/local/opt/gettext/lib/libintl.8.dylib
  Referenced from: /Users/xxxxx/Development/Tools/emsdk/python/3.9.2_64bit/bin/python3.9
  Reason: tried: '/usr/local/opt/gettext/lib/libintl.8.dylib' (no such file), '/usr/local/lib/libintl.8.dylib' (no such file), '/usr/lib/libintl.8.dylib' (no such file)
zsh: abort      python/3.9.2_64bit/bin/python3
```

tangobravo:
This seems to have been introduced with the recent rebuild from #975 - I just downloaded the previous 3.9.2-1 archive that runs without issue.

Here's the difference in the linked libraries from tool - the previous 3.9.2-1:
```
% otool -L ~/Downloads/python-3.9.2-1/bin/python3
/Users/xxxxx/Downloads/python-3.9.2-1/bin/python3:
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1770.106.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.0.0)
```

And the current revision 2 build:
```
% otool -L python/3.9.2_64bit/bin/python3
python/3.9.2_64bit/bin/python3:
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1677.104.0)
	/usr/local/opt/gettext/lib/libintl.8.dylib (compatibility version 11.0.0, current version 11.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
```

tangobravo:
Just trying again with some checkouts into clean folders.

An `emsdk install latest` is picking up python-3.9.2-1-macos-arm64.tar.gz which is still working OK.

[...]

PR #981, opened 2022-01-30
Fix mac python by avoiding linking libintl

sbc100:
Fixes: #980


kripken (approved):
Not familiar with the MacOS side here, but generally lgtm.

Can we test these things?

sbc100:
> Not familiar with the MacOS side here, but generally lgtm.
> 
> Can we test these things?

I tested locally by running `otool -L` to confirm that the dependencies were not present... we could invest in more testing here but (such as running otool post-build) but I'm not sure it worth it.

PR #803, opened 2021-04-29
Remove explicit closure install step

juj:
Remove explicit closure install step, the first install command does already seem to install google-closure-compiler-windows at least.

See #802 , #793, #636.

This would help from Linux and macOS testing to see if that works for people. CC also @curiousdannii

Issue #802, opened 2021-04-28
google-closure-compiler-linux version not properly pinned when installing Emscripten


akien-mga:
Since yesterday we're getting CI build issues for Godot ([example](https://github.com/akien-mga/godot/runs/2455400498?check_suite_focus=true)) due to an incompatibility between the old Emscripten version we use on CI (`1.39.20`, context below) and the apparently unbounded version of [google-closure-compiler-linux](https://www.npmjs.com/package/google-closure-compiler-linux) (which was updated yesterday on npm, probably introducing this compatibility breakage).

Here's the error we get:
```
/home/akien/Projects/godot/emscripten/emsdk/node/14.15.5_64bit/bin/node /home/akien/Projects/godot/emscripten/emsdk/upstream/emscripten/node_modules/.bin/google-closure-compiler --compilation_level ADVANCED_OPTIMIZATIONS --externs /home/akien/Projects/godot/godot-3.x/platform/javascript/js/engine/engine.externs.js --js /home/akien/Projects/godot/godot-3.x/platform/javascript/js/engine/preloader.js --js /home/akien/Projects/godot/godot-3.x/platform/javascript/js/engine/config.js --js /home/akien/Projects/godot/godot-3.x/platform/javascript/js/engine/engine.js --js_output_file /home/akien/Projects/godot/godot-3.x/bin/godot.javascript.opt.engine.js
Building compilation database compile_commands.json
progress_finish(["progress_finish"], [])
Linking Program        ==> bin/godot.javascript.opt.js
building:ERROR: Closure compiler run failed:

building:ERROR: /home/akien/Projects/godot/emscripten/emsdk/upstream/emscripten/src/closure-externs/closure-externs.js:80:4: ERROR - [JSC_VAR_MULTIPLY_DECLARED_ERROR] Variable MemoryDescriptor declared more than once. First occurrence: externs.zip//webassembly.js
  80| var MemoryDescriptor;
          ^^^^^^^^^^^^^^^^

/home/akien/Projects/godot/emscripten/emsdk/upstream/emscripten/src/closure-externs/closure-externs.js:87:4: ERROR - [JSC_VAR_MULTIPLY_DECLARED_ERROR] Variable TableDescriptor declared more than once. First occurrence: externs.zip//webassembly.js
  87| var TableDescriptor;
          ^^^^^^^^^^^^^^^

[...]

PR #995, opened 2022-03-04
Fix native Closure Compiler to work.

juj:
Fix native Closure Compiler to work.

Reverts my earlier PR #803 which was a great mistake, as it caused Emscripten to silently fall back to Java version of the Closure Compiler.

After this PR, Closure Compiler will work on user platforms that do not have Java installed.

Also forcibly remove Java version of Closure Compiler on systems where installing the native version succeeds, in order to save on disk size.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1440,6 +1440,7 @@ def emscripten_npm_install(tool, directory):
     except KeyError as e:
       # The target version of Emscripten does not (did not) have a package.json that would contain google-closure-compiler. (fastcomp)
       # Skip manual native google-closure-compiler installation there.
+      print(str(e))
       print('Emscripten version does not have a npm package.json with google-closure-compiler dependency, skipping native google-closure-compiler install step')
       return True
 
--- a/emsdk.py
+++ b/emsdk.py
@@ -1402,6 +1402,8 @@ def emscripten_npm_install(tool, directory):
   env = os.environ.copy()
   env["PATH"] = node_path + os.pathsep + env["PATH"]
   print('Running post-install step: npm ci ...')
+  # Do a --no-optional install to avoid bloating disk size:
+  # https://github.com/emscripten-core/emscripten/issues/12406
   try:
     subprocess.check_output(
         [npm, 'ci', '--production', '--no-optional'],
@@ -1412,7 +1414,7 @@ def emscripten_npm_install(tool, directory):
     return False
 
   # Manually install the appropriate native Closure Compiler package
-  # This is currently needed because npm ci will install the packages
+  # This is currently needed because npm ci would install the packages
   # for Closure for all platforms, adding 180MB to the download size
   # There are two problems here:
   #   1. npm ci does not consider the platform of optional dependencies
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -1468,16 +1468,22 @@ def emscripten_npm_install(tool, directory):
...
+          raise Exception('Failed to patch google-closure-compiler-java dependency away!')
+        open(compiler_filename, 'w').write(new_js)
+
+        # Now that we succeeded to install the native version and patch away the Java dependency, delete the Java version
+        # since that takes up ~12.5MB of installation space that is no longer needed.
+        # This process is currently a little bit hacky, see https://github.com/google/closure-compiler/issues/3926
+        remove_tree(os.path.join(directory, 'node_modules', 'google-closure-compiler-java'))
+        print('Removed google-closure-compiler-java dependency.')
+      else:
+        errlog('Failed to patch away google-closure-compiler Java dependency. ' + compiler_filename + ' does not exist.')
```
sbc100, line 1486: This whole block starting from 1471 can move outside the try/catch here I think.

emsdk.py
```diff
@@ -1440,6 +1440,54 @@ def emscripten_npm_install(tool, directory):
...
+  closure_compiler_native = ''
+  if LINUX and ARCH in ('x86', 'x86_64'):
+    closure_compiler_native = 'google-closure-compiler-linux'
+  if MACOS and ARCH in ('x86', 'x86_64'):
+    closure_compiler_native = 'google-closure-compiler-osx'
+  if WINDOWS and ARCH == 'x86_64':
+    closure_compiler_native = 'google-closure-compiler-windows'
+
+  if closure_compiler_native:
+    closure_compiler_native += '@20220104.0.0'
```
sbc100, line 1463: Won't this need updating every time we do an npm update on the emscripten-side?   (iirc we have updated several times over the years so any one value here might be work).

sbc100 (approved):

emsdk.py
juj: Yes it does, and that is quite unfortunate. I wrote a little bit about that earlier in https://github.com/emscripten-core/emsdk/issues/802#issuecomment-829095886 . I don't see another way but to keep these in sync.

emsdk.py
juj: check

emsdk.py
```diff
@@ -1440,6 +1440,54 @@ def emscripten_npm_install(tool, directory):
     errlog('Error running %s:\n%s' % (e.cmd, e.output))
     return False
 
+  # Manually install the appropriate native Closure Compiler package
+  # This is currently needed because npm ci will install the packages
+  # for Closure for all platforms, adding 180MB to the download size
+  # There are two problems here:
```
sbc100, line 1446: Since `npm ci` is actually run above, perhaps mention either up there, or here in this comment that the above command deliberately excludes the native versions of closure compiler by the use of the `--no-optional` flag.

emsdk.py
sbc100: Can we maybe do something like this:

```
closure_version = json.load('package.json')['dependencies']['google-closure-compiler']
```

juj:
> Can we maybe do something like this:

That is a great idea! I'll look into that a bit later.

Issue #972, opened 2022-01-18
Unable to force copts and/or linkopts for wasm_cc_binary and its dependencies.


PiotrSikora:
For context, I'm working on updating Proxy-Wasm C++ SDK from our custom toolchain to @emsdk (see: https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/pull/132) and I'm seeing some regressions in terms of features.

1. We hardcoded and always enabled [`c++17`](https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/c17bf8257e5ebad2399f95cc15c8ff64e515519f/toolchain/cc_toolchain_config.bzl#L92-L101) and [`various optimizations`](https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/c17bf8257e5ebad2399f95cc15c8ff64e515519f/toolchain/cc_toolchain_config.bzl#L126-L147) in our toolchain, so all invocations of Emscripten included those, even when building dependencies. However, with @emsdk, even if I provide those `copts`/`linkopts` in `cc_binary` that's the target of `wasm_cc_binary`, those flags apply only to the given target and not to any of its dependencies.

2. On the other hand, `copts` and `linkopts` provided in `.bazelrc` or on the command line are picked up by `wasm_cc_binary`, which means that features unsupported for `wasm32-unknown-emscripten` target (such as `-fsanitize=function`) are getting picked up and break the build.

Both issues result in some weird issues when using this toolchain in larger workspaces (e.g. Envoy), where we want an optimized `.wasm` modules to be built and loaded in debug or instrumented runtime.

PiotrSikora:
Thinking more about this, I think the only option to make this work for us (if this is doable, of course) is to add `copts` and `linkopts` parameters to `wasm_cc_binary` that would be used when building given target and all its dependencies (and only those, ignoring anything set in `.bazelrc` or on the command line). This would of course be optional, and in absence of those parameters, `wasm_cc_binary` would behave as it does currently, picking `copts` and `linkopts` from the Bazel environment.

walkingeyerobot:
I believe what you're running into here is a limitation of bazel. It is not unreasonable for someone to want to build, say, a wasm module in `-c dbg` but the webserver that serves it in `-c fastbuild` in a single `bazel` invocation. However, such a thing isn't possible; compilation mode applies to the entire build graph regardless of transitions (except I _think_ for the host transition which is hard coded to be fastbuild).

[...]

PR #998, opened 2022-03-09
Provide "executable" in wasm_cc_binary's DefaultInfo.

PiotrSikora:

### DIFF
--- a/bazel/emscripten_toolchain/wasm_cc_binary.bzl
+++ b/bazel/emscripten_toolchain/wasm_cc_binary.bzl
@@ -80,6 +80,7 @@ def _wasm_binary_impl(ctx):
     )
 
     return DefaultInfo(
+        executable = ctx.outputs.wasm,
         files = depset(outputs),
         # This is needed since rules like web_test usually have a data
         # dependency on this target.


PiotrSikora:
This is needed to allow custom rules and/or transitions on top of `wasm_cc_binary` (e.g. to workaround #972).

It's debatable whether `.wasm` is the correct `executable` target for non-standalone Wasm builds, but we don't have a way to differentiate those in `wasm_cc_binary`, and considering that this attribute isn't exported now, it shouldn't break anything.

cc @trybka

trybka:
LGTM.

@walkingeyerobot or @sbc100 can review / import

sbc100 (approved):
rubberstamp lgtm

bkotsopoulossc:
Can you explain what this helps accomplish? I’m not quite following

PiotrSikora:
This allows accessing `.wasm` module via [`DefaultInfo.files_to_run.executable`](https://bazel.build/rules/lib/DefaultInfo#files_to_run), and using `wasm_cc_binary` as an input in another rule.

I'm using this to force extra transitions in [`proxy_wasm_cc_binary`](https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/05862acc58782f108d2193a7f2db461bb6601fdd/bazel/defs.bzl).

PR #1001, opened 2022-03-16
Switch to nuget python package for windows

sbc100:
Previously we were using the "embedded" version python which has caused
us a few problems over time.  Currently the issue is that it does not include
pip.  This is a more normal distribution of python that does include pip.

Fixes https://github.com/emscripten-core/emscripten/issues/16466

### DIFF
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -422,6 +422,15 @@
   },
   {
     "id": "python",
+    "version": "3.9.2-nuget",
+    "bitness": 64,
+    "arch": "x86_64",
+    "windows_url": "python-3.9.2-4-amd64+pywin32.zip",
+    "activated_cfg": "PYTHON='%installation_dir%/python.exe'",
+    "activated_env": "EMSDK_PYTHON=%installation_dir%/python.exe"
+  },
+  {
+    "id": "python",
     "version": "3.9.2",
     "bitness": 64,
     "arch": "x86_64",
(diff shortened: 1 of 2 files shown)


sbc100:
PR testing that this working on emscripten CI: https://github.com/emscripten-core/emscripten/pull/16501

emsdk_manifest.json
```diff
@@ -425,7 +425,7 @@
     "version": "3.9.2",
     "bitness": 64,
     "arch": "x86_64",
-    "windows_url": "python-3.9.2-1-embed-amd64+pywin32.zip",
+    "windows_url": "python-3.9.2-4-amd64+pywin32.zip",
```
juj, line 428: This could be a risky change.. so I think it is not a good idea to pull the existing package to a new one. This prevents people from being able to update to it (when they have the old one, emsdk will see it as installed, so it won't attempt to reinstall), but it also prevents people from downgrading if they are met with issues, since the old package will vanish from the registry.

Maybe instead we can register a new python package and leave the old one intact, and then could also take that as an opportunity to update to the latest python 3.10.2?

sbc100: Done

juj (commented):
Great!

Would you have uncompressed size delta numbers at hand by any chance?

juj (approved):

juj:
Ops, yeah, forgot to approve, lgtm.

PR #1011, opened 2022-03-22
Add prebuilt binaries of 3.1.7 for linux/arm64

sbc100:
Thanks for @otterley for doing that hard work of actually building these binaries.

I basically uploaded them and then did and confirmed they are installable have have the right architecture:

```
$ EMSDK_ARCH=aarch64 ./emsdk install 3.1.7
$ file upstream/bin/clang-15 
upstream/bin/clang-15: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, BuildID[xxHash]=15798223cc2ffa5f, not stripped
```

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -42,7 +42,7 @@ emsdk_packages_url = 'https://storage.googleapis.com/webassembly/emscripten-rele
 
 emscripten_releases_repo = 'https://chromium.googlesource.com/emscripten-releases'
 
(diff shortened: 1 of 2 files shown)


dschuff (approved):
This looks good. If we ultimately move to a different host for the Aarch64 binaries, we can do what we did when we moved from mozilla to google hosting, and also move any old release binaries we want to keep available.

PR #996, opened 2022-03-04
Fix macOS build

juj:
Emsdk is unable to build Binaryen from source on macOS since Binaryen adopted new libstdc++ that required a newer macOS, issue https://github.com/WebAssembly/binaryen/issues/4299

Bump the minimum build requirement up to that version.

PR #1014, opened 2022-03-23
Remove non-ascii char from emsdk.py. NFC

sbc100:
This was introduced by mistake in #996 and cause `Emsdk.test_no_32bit` to fail.


juj:
Doh.. what a classic! :)

PR #1017, opened 2022-03-27
Set environment for powershell on all platforms

KiruyaMomochi:
## Before this PR

Running ` . ./emsdk_env.ps1` on Linux PowerShell (`pwsh`) only output `Setting environment variables`, but **none
of these variables are set actually**.

```pwsh
PS /home/kyaru/projects/emsdk> . ./emsdk_env.ps1
Adding directories to PATH: ...
Setting environment variables: ...

PS /home/kyaru/projects/emsdk> emsdk
emsdk: The term 'emsdk' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```

## Solution

Decide whether to call `write_set_env_script` or not by checking current shell, not platform.

## After this PR

The script works on all platforms.

```pwsh
PS /home/kyaru/projects/emsdk> . ./emsdk_env.ps1
Adding directories to PATH: ...
Setting environment variables: ...

PS /home/kyaru/projects/emsdk> emsdk
Missing command; Type 'emsdk help' to get a list of commands.
```

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2461,7 +2461,7 @@ def process_tool_list(tools_to_activate):
 
 
 def write_set_env_script(env_string):
-  assert(WINDOWS)
+  assert(CMD or POWERSHELL)
   open(EMSDK_SET_ENV, 'w').write(env_string)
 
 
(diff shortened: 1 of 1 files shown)


sbc100:
TIL, you can run powershell outside of windows.  Out of interested, on what platform are you running it?

sbc100 (approved):

KiruyaMomochi:
I'm using it on all platforms. 

See this GitHub Action file: https://github.com/KiruyaMomochi/AssetStudio/blob/cross/.github/workflows/texture-decoder.yml

PR #1022, opened 2022-04-07
Pass --no-checkout to git clone commands

juj:
Pass --no-checkout to git clone commands because installer will always issue an explicit clone command right afterwards. This is a small optimization on installation disk performance.


juj:
Actually, this won't work well with shallow clones, but needs a wider refactor.

PR #1028, opened 2022-04-13
Pass -DENABLE_WERROR=0 when building Binaryen

juj:
We are experiencing an issue at our Unity CI with building Binaryen: https://github.com/WebAssembly/binaryen/issues/4588

It seems that for end users, disabling -Werror is a good general measure to enable wider chance of success to build. Emsdk installations are unlikely to be used by Binaryen developers to iterate on Binaryen development, so it is not necessary there?

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1538,7 +1538,7 @@ def build_binaryen_tool(tool):
   build_type = decide_cmake_build_type(tool)
 
   # Configure
-  args = ['-DENABLE_WERROR=0'] # -Werror is not useful for end users
+  args = ['-DENABLE_WERROR=0']  # -Werror is not useful for end users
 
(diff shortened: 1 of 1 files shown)


sbc100 (approved):
Do you think we should turn is off by default in binaryen instead then?

juj:
I don't know how many end users directly git clone Binaryen to use it, though generally I do recommend -Werror to be a setting that developers and CI would have enabled, and end users would by default get disabled. It can be tricky to simultaneously set up a "developers by default get -Werror on and end users by default get -Werror off" if both users follow the same setup steps. But it is probably good to assume that most people installing via emsdk will be end users, and most people git cloning binaryen directyl would be developers. So maybe Binaryen repo could default to having -Werror enabled, and then emsdk can disable it by default.

PR #1029, opened 2022-04-13
Enable building LLVM with about-to-be deprecated toolchains

juj:
Enable building LLVM with about-to-be deprecated toolchains for best chance of succeeding with the build.


emsdk.py
```diff
@@ -1043,6 +1043,9 @@ def cmake_configure(generator, build_root, src_root, build_type, extra_cmake_arg
     # Target macOS 10.14 at minimum, to support widest range of Mac devices from "Early 2008" and newer:
     # https://en.wikipedia.org/wiki/MacBook_(2006-2012)#Supported_operating_systems
     cmdline += ['-DCMAKE_OSX_DEPLOYMENT_TARGET=10.14']
+    # To enable widest possible chance of success for building, let the code compile through with older toolchains
```
sbc100, line 1046: Would you mind setting your textwidth to 80 for emscripten and related projects?  (to avoid these long lines that wrap awkwardly in github and in local editors).

sbc100 (approved):

emsdk.py
juj: Sure, updated a ruler to my text editor for 80 columns.

PR #1030, opened 2022-04-13
Git clone desired target branch directly

juj:
When doing a git clone of a branch, instead of doing a general git clone that first would check out the default (main/master) branch and then later doing a checkout to switch to the target branch, instead specify the `--branch` option to the clone command line to immediately clone and checkout the desired final branch.

This helps first checkout runtime performance on CIs, especially when using shallow clones (GIT_CLONE_SHALLOW option).

Also this sidesteps a really odd git clone issue we are currently seeing in our CI, where it is unable to `checkout` the googletest submodule in binaryen, but would fail saying "path not found".

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -788,21 +788,21 @@ def git_recent_commits(repo_path, n=20):
     return []
 
 
-def git_clone(url, dstpath):
+def git_clone(url, dstpath, branch):
   debug_print('git_clone(url=' + url + ', dstpath=' + dstpath + ')')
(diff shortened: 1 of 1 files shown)


juj:
There are different custom steps that we need, and we use a "canonicalized" directory layout to enable multiple external Unity tools to be able to access the different executables and utilities in a version agnostic manner.

And then we do a bunch of custom stripping to remove things that we don't need to reduce code size, and do a legal pass on the third party items that we use and don't use.

Issue #1020, opened 2022-04-01
Rename Bazel @nodejs repo to @nodejs_host


attilaolah:
Upstream bazel_nodejs rules have removed the `@nodejs` repo, now the repo is named `@nodejs_host` to access the files for the host architecture, or e.g. `@nodejs_linux_amd64` to target specific architectures (this was the case earlier too).

Since the toolchain is typically run on the host, we should simply replace `@nodejs//` with `@nodejs_host//` in `BUILD.bazel` files. Note however that this is a breaking change, and will also require users to update rules_nodejs.

However, without this change, users cannot upgrade rules_nodejs without patching the Emscripten SDK.

kripken:
cc @walkingeyerobot

walkingeyerobot:
I attempted to make this change but I can't quite get it to a state where it's working. I wasn't able to actually get bazel to find the repo named `@nodejs_host`. Based on bazelbuild/rules_nodejs#3375, it looks like this has just been proposed and not yet implemented. @attilaolah if you have more information, or even better if you want to put together a PR, I'd appreciate it!

jfirebaugh:
It seems like this has been implemented in rules_nodejs already. Or at least, current versions of rules_nodejs and emsdk do not work. Here's a minimal repro: https://github.com/jfirebaugh/bazel_emscripten_test. If you run `bazel build ...` in this workspace, you'll get:

> `ERROR: [...]/external/emsdk/emscripten_toolchain/BUILD.bazel:26:10: no such package '@nodejs//': The repository '@nodejs' could not be resolved: Repository '@nodejs' is not defined and referenced by '@emsdk//emscripten_toolchain:link-emscripten'`

trungdinhth:
Hi, I'm also seeing the same issue, trying to used emscripten and nodejs rule in the same workspace.
If specifying node js rule before emscripten, I got error: 
> no such package '@nodejs//': The repository '@nodejs' could not be resolved: Repository '@nodejs' is not defined and referenced by '@emsdk//emscripten_toolchain:common_files'

If specifying emscripten rule before node js, I got error:
> ERROR: error loading package '': cannot load '@build_bazel_rules_nodejs//:repositories.bzl': no such file

In waiting for a real fix, do you know any possible workaround or where to patch?

attilaolah:
[...]

PR #1035, opened 2022-04-14
Conditionally use `@nodejs_host` or `@nodejs`

jfirebaugh:
Which of these exists depends on the version of rules_nodejs being used.

Fixes https://github.com/emscripten-core/emsdk/issues/1020.


walkingeyerobot (approved):

jfirebaugh:
Well, now that I'm getting around to trying this in my own project, which uses rules_nodejs 5.2.0, it looks like it does not work. I get the same "no such package '@nodejs//'" error. I guess `existing_rule` does not work like I thought it did in this context, and `existing_rule("@nodejs_host//:node_files")` always returns false.

PR #1036, opened 2022-04-14
[bazel] Switch to platforms-based toolchain resolution

### LINKED ISSUE #984: Bazel: `@emsdk` is broken with `--incompatible_enable_cc_toolchain_resolution`
This was already discussed in https://github.com/emscripten-core/emsdk/pull/766#issuecomment-806192512, but that PR was merged without a fix, and this isn't tracked anywhere else.

Platforms-based toolchain resoution (`--incompatible_enable_cc_toolchain_resolution`) is the future (https://github.com/bazelbuild/bazel/issues/7260), and majority of Bazel rules/toolchains already support it, but `@emsdk` doesn't  (build falls back to using the default CC toolchain), which breaks cross-compilation in any Bazel workspace that uses `@emsdk`.

cc @walkingeyerobot @celentes @bkotsopoulossc

jfirebaugh:
Fixes #984. However, the way I have done it requires `--incompatible_enable_cc_toolchain_resolution` and no longer supports `--crosstool_top=...`. I'm not sure how to support both at the same time. If that's necessary, can someone give me pointers on how to do it?

### DIFF
--- a/bazel/emscripten_toolchain/wasm_cc_binary.bzl
+++ b/bazel/emscripten_toolchain/wasm_cc_binary.bzl
@@ -26,13 +26,10 @@ def _wasm_transition_impl(settings, attr):
         features.append("wasm_simd")
 
     return {
-        "//command_line_option:compiler": "emscripten",
-        "//command_line_option:crosstool_top": "@emsdk//emscripten_toolchain:everything",
-        "//command_line_option:cpu": "wasm",
         "//command_line_option:features": features,
         "//command_line_option:dynamic_mode": "off",
         "//command_line_option:linkopt": linkopts,
-        "//command_line_option:platforms": [],
+        "//command_line_option:platforms": ["@emsdk//:platform_wasm"],
         "//command_line_option:custom_malloc": "@emsdk//emscripten_toolchain:malloc",
     }
 
@@ -43,9 +40,6 @@ _wasm_transition = transition(
         "//command_line_option:linkopt",
     ],
     outputs = [
-        "//command_line_option:compiler",
-        "//command_line_option:cpu",
-        "//command_line_option:crosstool_top",
         "//command_line_option:features",
         "//command_line_option:dynamic_mode",
         "//command_line_option:linkopt",
--- a/bazel/emscripten_toolchain/wasm_cc_binary.bzl
+++ b/bazel/emscripten_toolchain/wasm_cc_binary.bzl
@@ -26,6 +26,9 @@ def _wasm_transition_impl(settings, attr):
         features.append("wasm_simd")
 
     return {
+        "//command_line_option:compiler": "emscripten",
+        "//command_line_option:crosstool_top": "@emsdk//emscripten_toolchain:everything",
+        "//command_line_option:cpu": "wasm",
         "//command_line_option:features": features,
         "//command_line_option:dynamic_mode": "off",
         "//command_line_option:linkopt": linkopts,
@@ -40,6 +43,9 @@ _wasm_transition = transition(
         "//command_line_option:linkopt",
     ],
     outputs = [
+        "//command_line_option:compiler",
+        "//command_line_option:cpu",
+        "//command_line_option:crosstool_top",
         "//command_line_option:features",
         "//command_line_option:dynamic_mode",
         "//command_line_option:linkopt",
--- a/bazel/test_external/WORKSPACE
+++ b/bazel/test_external/WORKSPACE
@@ -10,3 +10,7 @@ deps()
 load("@emsdk//:emscripten_deps.bzl", "emscripten_deps")
 
 emscripten_deps()
+
+load("@emsdk//:toolchains.bzl", "register_emscripten_toolchains")
+
+register_emscripten_toolchains()
--- a/bazel/.bazelrc
+++ b/bazel/.bazelrc
@@ -1,0 +1,1 @@
+build --incompatible_enable_cc_toolchain_resolution
--- a/bazel/BUILD
+++ b/bazel/BUILD
@@ -66,3 +66,10 @@ alias(
         "//conditions:default": ":empty",
     }),
 )
+
+platform(
+    name = "platform_wasm",
+    constraint_values = [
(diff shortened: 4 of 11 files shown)


jfirebaugh:
It looks like test-bazel-mac runs a version of bazel that does not include `@platforms//cpu:wasm32`. I think this is what https://github.com/emscripten-core/emsdk/pull/766#issuecomment-806195568 was referring to. What's the preferred way to fix this?

zaucy:
@jfirebaugh if you rebase now with #1049 merged in the bazel mac CI should pass!

jfirebaugh:
Done.

Does emsdk need to continue to support both toolchain and non-toolchain (`--crosstool_top=...`) based resolution?

walkingeyerobot (approved):

bazel/test_external/WORKSPACE
```diff
@@ -10,3 +10,7 @@ deps()
 load("@emsdk//:emscripten_deps.bzl", "emscripten_deps")
 
 emscripten_deps()
+
+load("@emsdk//:toolchains.bzl", "register_emscripten_toolchains")
+
+register_emscripten_toolchains()
```
sbc100, line 16: Can/should we fold these two into a single step?

jfirebaugh: Seems like other rulesets tend to keep them separate. E.g. rules_go has `go_rules_dependencies ` and `go_register_toolchains `, rules_nodejs has `build_bazel_rules_nodejs_dependencies` and `node_repositories`, rules_rust has `rules_rust_dependencies` and `rust_register_toolchains`.

jun-sheaf:
Just found this PR. We've been using #1060 for a while internally, so we wanted to upstream those updates. I've noticed this PR misses out on some parts. Specifically, we should label the toolchain as `wasm32` (since that is what it is) and also fix the static linker flags. All of this is already done in #1060 (I've also downstreamed some of the changes here to there).
.

bazel/emscripten_toolchain/wasm_cc_binary.bzl
```diff
@@ -32,20 +32,20 @@ def _wasm_transition_impl(settings, attr):
...
+        "//command_line_option:platforms": ["@emsdk//:platform_wasm"],
         "//command_line_option:custom_malloc": "@emsdk//emscripten_toolchain:malloc",
     }
 
 _wasm_transition = transition(
     implementation = _wasm_transition_impl,
     inputs = [
+        "//command_line_option:compiler",
+        "//command_line_option:cpu",
+        "//command_line_option:crosstool_top",
```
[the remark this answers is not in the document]
jun-sheaf: This will not fix the CI. You need to set the incompatible toolchain flag. See #1060.

jun-sheaf:
> > Just found this PR. We've been using #1060 for a while internally, so we wanted to upstream those updates. I've noticed this PR misses out on some parts. Specifically, we should label the toolchain as `wasm32` (since that is what it is) and also fix the static linker flags. All of this is already done in #1060 (I've also downstreamed some of the changes here to there). .
> 
> 
> 
> Thanks for the PR! I'm happy with the state this PR is in (sans the small fix I commented on), so I'd like to get this in before reviewing yours.

I'd recommend taking a look at #1060 before moving forward with this one. There are some non-trivial fixes there and also improvements to the implementation. We also want it upstreamed sooner so we can check back out to @emsdk/main.

jfirebaugh:
I’m away from a computer for a few weeks and won’t be able to make the changes until July. Feel free to make the changes yourself or pursue the other PR, I won’t be offended. 😄

bazel/test_external/WORKSPACE
jfirebaugh: I saw that #1068 spells this `emsdk_register_toolchains`. Based on other rulesets, it does seem like the prevailing convention is to prefix with the language or toolset, rather than the verb `register`. However, I would propose that we retain consistency with `emscripten_deps`, and therefore spell this `emscripten_register_toolchains`. @walkingeyerobot @jun-sheaf WDYT?

jun-sheaf: Hey @jfirebaugh! Your observation is correct. But just some information regarding this, I have upstreamed this to internal already. Hopefully by August I can get the internal toolchain open-sourced.

jfirebaugh: What does "upstreamed to internal" mean?

jun-sheaf: I meant the patch has already been published within Google (the internal repo).

jfirebaugh:
Ping @walkingeyerobot.

HappyCerberus:
Hi, I tried using this patch, but run into a problem. I had to comment out this from `emscripten_deps.bzl`:

```
    if "emscripten_bin_mac_arm64" not in excludes:
        http_archive(
            name = "emscripten_bin_mac_arm64",
            strip_prefix = "install",
            url = emscripten_url.format("mac", revision.hash, "-arm64", "tbz2"),
            sha256 = revision.sha_mac_arm64,
            build_file_content = BUILD_FILE_CONTENT_TEMPLATE.format(bin_extension = ""),
            type = "tar.bz2",
        )
```

Without removing it I was getting the following error:

```
                sha256 = revision.sha_mac_arm64,
Error: 'struct' value has no field or method 'sha_mac_arm64'
Available attributes: hash, sha_linux, sha_mac, sha_win
```

jfirebaugh:
@HappyCerberus Are you sure it's related to this patch? It looks like you're using an emscripten version that doesn't provide an arm64 macOS build. Judging by revisions.bzl, 3.1.2 was the first version that supported it:

https://github.com/emscripten-core/emsdk/blob/c220895fd1163c01f0a8b44229fb9e4fe0ae0958/bazel/revisions.bzl#L124-L136

HappyCerberus:
@jfirebaugh Thanks. I did not realize this was it. Retested and it works now.

Perhaps would make sense to remove the older revisions from the revisions file, since the code doesn't work with them anyway?

Issue #883, opened 2021-09-07
Does using "bazel test" work?


spyle2:
I cant seem to use bazel to run a test that includes "emscripten.h" or any other emscripten libs.
I wrote a cc_test and use --config=wasm and have gone through many, many errors that I am slowly working through. cc_binary/cc_libary work fine using "bazel build --config=wasm".

I have gone through many errors; so manual editing of the toolchain flags fixed most. Currently I am stuck on it not being able to include node_modules.

```
internal/modules/cjs/loader.js:797
    throw err;
    ^

Error: Cannot find module './lib/utils'
Require stack:
```

That specific error isn't really my main question. I really want to know if anyone has gotten cc_test to work at all with including emscripten libs?
Should we create a wasm_cc_test?

walkingeyerobot:
I can confirm that `bazel test` does not currently work.

sbc100:
@DoDoENT @walkingeyerobot do you if this works now?

DoDoENT:
I'm running WASM tests regularly in my playground repo. See here: https://github.com/DoDoENT/bazel-playground

PR #1037, opened 2022-04-18
[bazel] Set CLOSURE_COMPILER to workaround RBE+symlinks issue

kjlubick:
When trying to compile using emscripten/emsdk on [Bazel's remote executors](https://bazel.build/docs/remote-execution), I ran into an error like:

```
Linking ... failed: (Exit 1): emcc_link.sh failed: error executing command external/emsdk/emscripten_toolchain/emcc_link.sh @bazel-out/wasm-opt-ST-d9bb9c985c3e/bin/...-2.params
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module './lib/utils'
Require stack:
- /b/f/w/external/emscripten_bin_linux/emscripten/node_modules/.bin/google-closure-compiler
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:94:18)
    at Object.<anonymous> (/b/f/w/external/emscripten_bin_linux/emscripten/node_modules/.bin/google-closure-compiler:20:57)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:79:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/b/f/w/external/emscripten_bin_linux/emscripten/node_modules/.bin/google-closure-compiler'
  ]
}
node:internal/modules/cjs/loader:936
  throw err;
  ^
```

By looking at the remote execution logs (e.g. appending `--execution_log_json_file=/tmp/execlog.json` to my Bazel command), I was able to figure out that Bazel was [not preserving symlinks](https://bazel.build/docs/remote-execution-rules#manage-workspace-rules) that were created by npm install, but rather uploading the linked file in place of the symlink.

Instead of uploading
```
.../node_modules/.bin/google-closure-compiler # symlink to ../google-closure-compiler/cli.js
.../node_modules/google-closure-compiler/cli.js
.../node_modules/google-closure-compiler/lib/utils.js
```
it was uploading
```
.../node_modules/.bin/google-closure-compiler # copy of ../google-closure-compiler/cli.js
.../node_modules/google-closure-compiler/cli.js
.../node_modules/google-closure-compiler/lib/utils.js
```

Then, when executing `node_modules/.bin/google-closure-compiler`, it would try to load `node_modules/.bin/lib/utils.js` and not be able to find it.

To work around this issue, we can bypass the symlink and run the google-closure-compiler/cli.js ourselves, which is successfully able to load the lib/utils.js and all other dependent files.

https://github.com/emscripten-core/emsdk/issues/883 is possibly a related issue


bazel/emscripten_toolchain/emscripten_config
```diff
@@ -3,12 +3,21 @@ import platform
...
 NODE_JS = ROOT_DIR + "/external/nodejs_{}_{}/{}".format(system.lower(), machine, nodejs_binary)
+
+
+# This works around an issue with Bazel RBE where the symlinks in node_modules/.bin
+# are uploaded as the linked files, which means the cli.js cannot load its
+# dependencies from the expected locations.
+# See https://github.com/emscripten-core/emscripten/pull/16640 for more
+if system != "Windows":
+    CLOSURE_COMPILER = os.path.join(EMSCRIPTEN_ROOT, "node_modules",
+                                    "google-closure-compiler", "cli.js")
```
sbc100, line 23: Missing newline.

sbc100 (commented):
Seems like a reasonable fix to me.   Do we need to prefix this command with the node executable, or does the #! work well enough?

bazel/emscripten_toolchain/emscripten_config
kjlubick: Done

kjlubick:
I tested this locally on my Linux box and in a Linux RBE instance and it seems to work as is.

sbc100:
> I tested this locally on my Linux box and in a Linux RBE instance and it seems to work as is.

If it works, I think it would be safer to do `CLOSURE_COMPILER = [NODE_JS, ...]` since then we will be sure we are using the configured version of node.

kjlubick:
> > I tested this locally on my Linux box and in a Linux RBE instance and it seems to work as is.
> 
> If it works, I think it would be safer to do `CLOSURE_COMPILER = [NODE_JS, ...]` since then we will be sure we are using the configured version of node.

I tried that out and it appears to have worked.

walkingeyerobot (approved):

PR #1043, opened 2022-04-29
Make emsdk a proper Python package

martinRenou:
# Description

This PR turns this repo into a proper Python package which installs the emsdk script as a Python entry point, meaning that `pip install emsdk` will make the `emsdk` script discoverable in the path.

It also allows other Python packages to depend on emsdk and use it from Python.

# Motivation

The motivation behind this PR is to make a `conda-forge` package for emsdk, allowing people to install emsdk with conda:

```
conda install -c conda-forge emsdk
emsdk install 3.1.2
emsdk activate 3.1.2
```

You can also publish it on PyPi if you want, but that's not mandatory.

# Other comments

- I am opening this PR as draft and hoping to get some feedback from the emscripten devs on this before pushing more on the PR (updating the CI so that it tests the Python installed emsdk).

- I did my best to not change your usage of emsdk

- I had to move some scripts, JSON and txt files under the emsdkpy Python package, otherwise setuptools would not install them as part of the Python package.


.circleci/config.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       - run:
           name: test.py
           command: |
-            source emsdk_env.sh
+            source emsdkpy/emsdk_env.sh
```
kripken, line 48: In general I think this sounds good - having a proper python package could be useful in various ways!

I am concerned by breaking changes, though, like moving scripts out of the toplevel where users have expected them for long time. Is there a way to avoid such breaking changes? Perhaps we could create a new repo just for the package, which would include the main emsdk repo underneath it? Or maybe symlinks could help?

martinRenou: Thanks for your response. Indeed I've thought about not moving the scripts around at first, but I've not been able to do that. IIRC the issue was for installing the scripts as package_data for the Python package, which was not possible if not under a Python module. But I can give another try.

sbc100:
Closing for now.  Feel free to re-open if needed.

Issue #1026, opened 2022-04-11
Sandboxed bazel builds with emsdk are slow


jfirebaugh:
I noticed that a `wasm_cc_binary` target subjectively felt a lot slower to compile than its `cc_binary` equivalent. During the build, the Bazel java process was consuming 300-400% CPU. I used [the --profile flag](https://bazel.build/rules/performance#performance-profiling) and it showed that almost all the time was being spent in  `sandbox.createFileSystem` and `sandbox.delete`. So I used `--sandbox_debug` and poked into a few of the resulting directories. Each contains 17-18k symlinks, including the following:
* An `emscripten_npm_mac` tree containing ~600 directories and ~9k symlinks.
* An `emscripten_bin_mac` tree containing ~500 directories and ~8k symlinks.

emsdk is configured in my workspace as follows:

```
http_archive(
    name = "emsdk",
    sha256 = "7dc13d967705582e11ff62ae143425dbc63c38372f1a1b14f0cb681fda413714",
    strip_prefix = "emsdk-3.1.4/bazel",
    url = "https://github.com/emscripten-core/emsdk/archive/refs/tags/3.1.4.tar.gz",
    patch_args = ["-p2"],
    patch_tool = "patch",
    patches = [
        "//bazel/patches:0001-Fix-https-github.com-emscripten-core-emsdk-issues-10.patch",
    ],
)

load("@emsdk//:deps.bzl", emsdk_deps = "deps")
emsdk_deps()

load("@emsdk//:emscripten_deps.bzl", emsdk_emscripten_deps = "emscripten_deps")
emsdk_emscripten_deps(emscripten_version = "3.1.4")
```

Is there any way to improve / optimize this?

PR #1045, opened 2022-04-30
Optimize sandbox performance

jfirebaugh:
Link just the files needed to compile, link, or archive, rather than the entire directory tree. This drastically improves build times on macOS, where the sandbox is known to be slow (https://github.com/bazelbuild/bazel/issues/8230).

Fixes https://github.com/emscripten-core/emsdk/issues/1026

### DIFF
--- a/bazel/emscripten_toolchain/BUILD.bazel
+++ b/bazel/emscripten_toolchain/BUILD.bazel
@@ -8,44 +8,52 @@ package(default_visibility = ["//visibility:public"])
 node_files = "@nodejs_host//:node_files" if existing_rule("@nodejs_host//:node_files") else "@nodejs//:node_files"
 
 filegroup(
-    name = "common-script-includes",
+    name = "common_files",
     srcs = [
-        "emar.sh",
-        "emar.bat",
-        "emcc.sh",
-        "emcc.bat",
         "emscripten_config",
         "env.sh",
         "env.bat",
-        "@emsdk//:binaries",
-        "@emsdk//:node_modules",
         node_files,
     ],
 )
 
 filegroup(
-    name = "compile-emscripten",
-    srcs = [":common-script-includes"],
+    name = "compiler_files",
+    srcs = [
+        "emcc.sh",
+        "emcc.bat",
+        "@emsdk//:compiler_files",
+        ":common_files",
+    ],
 )
 
 filegroup(
-    name = "link-emscripten",
+    name = "linker_files",
     srcs = [
         "emcc_link.sh",
         "emcc_link.bat",
         "link_wrapper.py",
-        ":common-script-includes",
-        "@emsdk//:binaries",
-        node_files,
+        "@emsdk//:linker_files",
+        ":common_files",
     ],
 )
 
 filegroup(
-    name = "every-file",
+    name = "ar_files",
     srcs = [
-        ":compile-emscripten",
-        ":link-emscripten",
-        "@emsdk//:binaries",
+        "emar.sh",
+        "emar.bat",
+        "@emsdk//:ar_files",
+        ":common_files",
+    ],
+)
+
+filegroup(
+    name = "all_files",
+    srcs = [
+        ":compiler_files",
+        ":linker_files",
+        ":ar_files",
         node_files,
     ],
 )
@@ -59,7 +67,7 @@ emscripten_cc_toolchain_config_rule(
     name = "wasm",
     cpu = "wasm",
     em_config = "emscripten_config",
-    emscripten_binaries = "@emsdk//:binaries",
+    emscripten_binaries = "@emsdk//:compiler_files",
     script_extension = select({
         "@bazel_tools//src/conditions:host_windows": "bat",
         "//conditions:default": "sh",
@@ -68,12 +76,12 @@ emscripten_cc_toolchain_config_rule(
 
 cc_toolchain(
     name = "cc-compiler-wasm",
-    all_files = ":every-file",
-    ar_files = ":common-script-includes",
+    all_files = ":all_files",
+    ar_files = ":ar_files",
     as_files = ":empty",
-    compiler_files = ":compile-emscripten",
+    compiler_files = ":compiler_files",
     dwp_files = ":empty",
-    linker_files = ":link-emscripten",
+    linker_files = ":linker_files",
     objcopy_files = ":empty",
     strip_files = ":empty",
     toolchain_config = "wasm",
--- a/bazel/emscripten_toolchain/emscripten.BUILD
+++ b/bazel/emscripten_toolchain/emscripten.BUILD
@@ -1,6 +1,41 @@
 package(default_visibility = ['//visibility:public'])
 
 filegroup(
-    name = "all",
-    srcs = glob(["**"]),
+    name = "includes",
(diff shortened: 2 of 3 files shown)


bazel/emscripten_toolchain/emscripten.BUILD
```diff
@@ -1,6 +1,41 @@
...
+
+filegroup(
+    name = "linker_files",
+    srcs = [
+        "emscripten/emcc.py",
+        "bin/llvm-nm",
+        "bin/llvm-objcopy",
+        "bin/wasm-emscripten-finalize",
+        "bin/wasm-ld",
+        "bin/wasm-opt",
```
sbc100, line 31: What about the libraries (from the sysroot) that are used during linking?  Are they handled separately?

bazel/emscripten_toolchain/BUILD.bazel
```diff
@@ -68,12 +76,12 @@ emscripten_cc_toolchain_config_rule(
 
 cc_toolchain(
     name = "cc-compiler-wasm",
-    all_files = ":every-file",
-    ar_files = ":common-script-includes",
+    all_files = ":all_files",
```
sbc100, line 79: Is all_files even needed at all anymore?

bazel/emscripten_toolchain/emscripten.BUILD
```diff
@@ -1,6 +1,41 @@
...
+        "emscripten/cache/sysroot/include/compat/**",
+        "emscripten/cache/sysroot/include/**",
+        "lib/clang/15.0.0/include/**",
+    ]),
+)
+
+filegroup(
+    name = "compiler_files",
+    srcs = [
+        "emscripten/emcc.py",
```
sbc100, line 16: Where do `tools/*.py` and stuff like that get included?

sbc100 (commented):
I this works I think it would be great improvement!

jfirebaugh:
I should say -- I don't have any deep knowledge of emscripten. My methodology here was just to start each of `compiler_files`, `linker_files`, `ar_files` with an empty list and add things until both local testing (on macOS) and my own project built successfully. So it is likely that dependencies are missing if they aren't covered by those cases.

> What about the libraries (from the sysroot) that are used during linking? Are they handled separately?

I don't know. What libraries would those be and when would they be needed?

> Where do `tools/*.py` and stuff like that get included?

What are those used for?

sbc100:
> I should say -- I don't have any deep knowledge of emscripten. My methodology here was just to start each of `compiler_files`, `linker_files`, `ar_files` with an empty list and add things until both local testing (on macOS) and my own project built successfully. So it is likely that dependencies are missing if they aren't covered by those cases.
> 
> > What about the libraries (from the sysroot) that are used during linking? Are they handled separately?
> 
> I don't know. What libraries would those be and when would they be needed?

At link time you would access to all the system library in sysroot/libc/wasm32-emscripten.  For example `libc.a`.   The fact that your link is working without this makes me thing maybe the whole sysroot is still available somehow.

> 
> > Where do `tools/*.py` and stuff like that get included?
> 
> What are those used for?

They are part of the emcc compiler.  emcc.py depends a whole lot of other .py file such as the ones in `tools/` and won't load without those.  If `emcc.py` is running then somehow those files must be available.. but I can't see how you are making them available.

jfirebaugh:
> At link time you would access to all the system library in sysroot/libc/wasm32-emscripten. For example `libc.a`. The fact that your link is working without this makes me thing maybe the whole sysroot is still available somehow.

It looks like the answer is simply that nothing I've tested actually requires linking any of these libraries. For example, I can in fact remove the `--sysroot` argument entirely (by removing `ACTION_NAMES.cpp_link*` from the appropriate `flag_set`) and test_external still builds successfully.

> They are part of the emcc compiler. emcc.py depends a whole lot of other .py file such as the ones in tools/ and won't load without those. If emcc.py is running then somehow those files must be available.. but I can't see how you are making them available.

Ah, this appears to be because python [adds the containing directory of the script as the first entry in `sys.path`](https://docs.python.org/3/library/sys.html#sys.path) and if the script is a symlink, does so _after_ resolving the symlink (I can't find this detail documented however). This is a sandbox bypass so I imagine bazel has found a workaround for it somewhere, but I'm not sure where to look.

jfirebaugh:
It's an unresolved issue in Bazel: https://github.com/bazelbuild/bazel/issues/7091

jfirebaugh:
I don't know. As far as I can tell, `bin/clang` and `bin/clang++` are correctly registered as inputs for compile actions, and I don't have access to a Windows environment to debug further. In case it was something related to unsandboxed builds, I tested `bazel build //hello-world:hello-world-wasm --spawn_strategy=local` locally, but that works.

bazel/emscripten_toolchain/emscripten.BUILD
```diff
@@ -1,6 +1,44 @@
...
+        "bin/clang",
+        "bin/clang++",
```
zaucy, line 18: On windows all these executables need to have the `.exe` extension. A separate BUILD file specifically for windows may be necessary or maybe using `build_file_content` and template it.

jfirebaugh: Thank you for this hint! I've taken the templated `build_file_content` approach.

walkingeyerobot (approved):

kjlubick:
@trybka, you had asked about a before and after benchmark. I noticed a **substantial** improvement when building CanvasKit on my 24 core Linux machine.

```
# emsdk-3.1.10
$ bazel clean && bazel build //modules/canvaskit:canvaskit_webgl --compilation_mode=dbg --sandbox_base=/dev/shm
...
INFO: Elapsed time: 6031.283s, Critical Path: 701.38s
INFO: 1662 processes: 101 internal, 1561 linux-sandbox.
INFO: Build completed successfully, 1662 total actions

# emsdk-3.1.10 with RAM disk
$ bazel clean && bazel build //modules/canvaskit:canvaskit_webgl --compilation_mode=dbg --sandbox_base=/dev/shm
...
INFO: Elapsed time: 179.974s, Critical Path: 75.22s
INFO: 1662 processes: 101 internal, 1561 linux-sandbox.
INFO: Build completed successfully, 1662 total actions

# emsdk-3.1.13
$ bazel clean && bazel build //modules/canvaskit:canvaskit_webgl --compilation_mode=dbg
...
INFO: Elapsed time: 215.952s, Critical Path: 71.56s
INFO: 1662 processes: 101 internal, 1561 linux-sandbox.
INFO: Build completed successfully, 1662 total actions

# emsdk-3.1.13 with RAM disk
$ bazel clean && bazel build //modules/canvaskit:canvaskit_webgl --compilation_mode=dbg --sandbox_base=/dev/shm
...
INFO: Elapsed time: 126.265s, Critical Path: 78.31s
INFO: 1662 processes: 101 internal, 1561 linux-sandbox.
INFO: Build completed successfully, 1662 total actions
```

With my sandbox on a RAM disk - a 30% improvement. Using the default sandboxing, a **~30x** improvement.

However, I do notice some issues when trying to compile the release mode and with a remote executor. Not all the files are specified, e.g. the closure compiler. I can get a PR up that addresses those issues for me.

Issue #986, opened 2022-02-11
emscripten:hello-world-wasm fails with tarfile.ReadError


luxe:
I followed the setup instructions [here](https://github.com/emscripten-core/emsdk/tree/main/bazel#setup-instructions).
The build command I'm using is:  `bazel build --verbose_failures //code/programs/examples/emscripten:hello-world-wasm`
The same code and BUILD definition in the instructions.

The error I get is:
```
INFO: Analyzed target //code/programs/examples/emscripten:hello-world-wasm (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: /home/luxe/Desktop/unilang/source/code/programs/examples/emscripten/BUILD:9:15: Action code/programs/examples/emscripten/hello-world-wasm/hello-world.js failed: (Exit 1): wasm_binary failed: error executing command 
  (cd /home/luxe/.cache/bazel/_bazel_luxe/9b1d289b10540c9752c68b55618a2156/sandbox/linux-sandbox/838/execroot/__main__ && \
  exec env - \
  bazel-out/k8-opt-exec-2B5CBBC6/bin/external/emsdk/emscripten_toolchain/wasm_binary --output_path bazel-out/k8-fastbuild/bin/code/programs/examples/emscripten/hello-world-wasm --archive bazel-out/wasm-fastbuild-ST-95dcf5f511a2/bin/code/programs/examples/emscripten/hello-world)
Execution platform: @local_config_platform//:host

Use --sandbox_debug to see verbose messages from the sandbox
Traceback (most recent call last):
  File "/home/luxe/.cache/bazel/_bazel_luxe/9b1d289b10540c9752c68b55618a2156/sandbox/linux-sandbox/838/execroot/__main__/bazel-out/k8-opt-exec-2B5CBBC6/bin/external/emsdk/emscripten_toolchain/wasm_binary.runfiles/emsdk/emscripten_toolchain/wasm_binary.py", line 68, in <module>
    main()
  File "/home/luxe/.cache/bazel/_bazel_luxe/9b1d289b10540c9752c68b55618a2156/sandbox/linux-sandbox/838/execroot/__main__/bazel-out/k8-opt-exec-2B5CBBC6/bin/external/emsdk/emscripten_toolchain/wasm_binary.runfiles/emsdk/emscripten_toolchain/wasm_binary.py", line 49, in main
    tar = tarfile.open(args.archive)
  File "/usr/lib/python3.9/tarfile.py", line 1616, in open
[...]

PR #1047, opened 2022-05-06
Explicit outputs for wasm_cc_binary

zaucy:
I originally started this PR because I was running into something similar to #986 and found out that my archive was `libfoo.so` and not the expected `foo.so` `wasm_binary.py` required. This PR doesn't just solve that problem for me, but it also allows me to specify the name of the outputs as well as _which outputs_ I would like.

```python
wasm_cc_binary(
    name = "foo",
    cc_target = ":foo_cc",
    # NEW
    outputs = [
        "foo.wasm",
        "foo.js",
    ],
)
```
I think this is a good change for 2 reasons:

1. [rule `outputs` atttribute is deprecated](https://docs.bazel.build/versions/main/skylark/rules.html#deprecated-predeclared-outputs)
2. Outputs are more strict. If I want a sepcific output and my configuration doesn't allow that output, I get an error.

Let me know what you think!

### DIFF
--- a/bazel/emscripten_toolchain/wasm_cc_binary.bzl
+++ b/bazel/emscripten_toolchain/wasm_cc_binary.bzl
@@ -54,56 +54,45 @@ _wasm_transition = transition(
     ],
 )
 
+_allow_output_extnames = [
+    '.js',
+    '.wasm',
+    '.wasm.map',
+    '.worker.js',
+    '.js.mem',
+    '.data',
+    '.fetch.js',
+    '.js.symbols',
+    '.wasm.debug.wasm',
+    '.html',
+]
+
 def _wasm_binary_impl(ctx):
     args = ctx.actions.args()
-    args.add("--output_path", ctx.outputs.loader.dirname)
+    args.add_joined("--outputs", ctx.outputs.outputs, join_with=",")
     args.add_all("--archive", ctx.files.cc_target)
 
-    outputs = [
-        ctx.outputs.loader,
-        ctx.outputs.wasm,
-        ctx.outputs.map,
-        ctx.outputs.mem,
-        ctx.outputs.fetch,
-        ctx.outputs.worker,
-        ctx.outputs.data,
-        ctx.outputs.symbols,
-        ctx.outputs.dwarf,
-        ctx.outputs.html,
-    ]
+    for output in ctx.outputs.outputs:
+        valid_extname = False
+        for allowed_extname in _allow_output_extnames:
+            if output.path.endswith(allowed_extname):
+                valid_extname = True
+                break
+        if not valid_extname:
+            fail("Invalid output '{}'. Allowed extnames: {}".format(output.basename, ", ".join(_allow_output_extnames)))
 
     ctx.actions.run(
         inputs = ctx.files.cc_target,
-        outputs = outputs,
+        outputs = ctx.outputs.outputs,
         arguments = [args],
         executable = ctx.executable._wasm_binary_extractor,
     )
 
     return DefaultInfo(
-        executable = ctx.outputs.wasm,
-        files = depset(outputs),
         # This is needed since rules like web_test usually have a data
         # dependency on this target.
-        data_runfiles = ctx.runfiles(transitive_files = depset(outputs)),
+        data_runfiles = ctx.runfiles(transitive_files = depset(ctx.outputs.outputs)),
     )
-
-def _wasm_binary_outputs(name, cc_target):
-    basename = cc_target.name
-    basename = basename.split(".")[0]
-    outputs = {
-        "loader": "{}/{}.js".format(name, basename),
-        "wasm": "{}/{}.wasm".format(name, basename),
-        "map": "{}/{}.wasm.map".format(name, basename),
-        "mem": "{}/{}.js.mem".format(name, basename),
-        "fetch": "{}/{}.fetch.js".format(name, basename),
-        "worker": "{}/{}.worker.js".format(name, basename),
-        "data": "{}/{}.data".format(name, basename),
-        "symbols": "{}/{}.js.symbols".format(name, basename),
-        "dwarf": "{}/{}.wasm.debug.wasm".format(name, basename),
-        "html": "{}/{}.html".format(name, basename),
-    }
-
-    return outputs
 
 # Wraps a C++ Blaze target, extracting the appropriate files.
 #
@@ -134,6 +123,10 @@ wasm_cc_binary = rule(
         "simd": attr.bool(
             default = False,
         ),
+        "outputs": attr.output_list(
+            allow_empty = False,
+            mandatory = True,
+        ),
         "_allowlist_function_transition": attr.label(
             default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
         ),
@@ -144,5 +137,4 @@ wasm_cc_binary = rule(
             default = Label("@emsdk//emscripten_toolchain:wasm_binary"),
         ),
     },
-    outputs = _wasm_binary_outputs,
 )
--- a/bazel/emscripten_toolchain/wasm_cc_binary.bzl
+++ b/bazel/emscripten_toolchain/wasm_cc_binary.bzl
@@ -77,7 +77,7 @@ def _wasm_binary_impl(ctx):
         basename = cc_target.label.name
         basename = basename.split(".")[0]
         for extname in _allow_output_extnames:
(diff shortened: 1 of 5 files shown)


bazel/emscripten_toolchain/wasm_cc_binary.bzl
```diff
@@ -134,6 +123,10 @@ wasm_cc_binary = rule(
         "simd": attr.bool(
             default = False,
         ),
+        "outputs": attr.output_list(
+            allow_empty = False,
```
walkingeyerobot, line 127: Is there a way to set this to `True` instead? We'd rather not break existing users.

walkingeyerobot (changes requested):
Thanks for the PR! I'm all for moving away from deprecated bazel APIs :)

I'd like to learn more about what exactly is allowed by the new outputs parameter. What happens if I specify `['foo.wasm', 'bar.js']`? I imagine this would fail during link in wasm_binary.py because bar.js does not exist? If this is the case, I'd prefer to fail earlier, i.e. in the rule, so we don't waste time doing a full build that's not going to work.

It's also noteworthy that I believe we can't just specify extensions in an outputs attribute because those would not be known to bazel ahead of time and therefore wouldn't be able to be used by rules that consume `wasm_cc_binary` files (although it would work if users were using `wasm_cc_binary` as their final rule).

bazel/emscripten_toolchain/wasm_cc_binary.bzl
walkingeyerobot: Ok we might be able to do something with a macro that sits in front of `wasm_cc_binary` and sets `outputs` to a default list if it's not set, effectively keeping the behavior. I am happy to help with this part as it might be a bit tricky.

The big thing here is I don't want to break an unknown number of projects depending on this rule with a significant breaking API change.

zaucy:
> I'd like to learn more about what exactly is allowed by the new outputs parameter. What happens if I specify `['foo.wasm', 'bar.js']`? I imagine this would fail during link in wasm_binary.py because bar.js does not exist?

Ah yes that would fail. My main usecase has been standalone wasm so I'm a little out of touch with the other output files. Would checking for certain output filenames matching be scalable? Or is that a painful maintenance burden? I don't know the details for which files are required with which settings.

> It's also noteworthy that I believe we can't just specify extensions in an outputs attribute because those would not be known to bazel ahead of time and therefore wouldn't be able to be used by rules that consume wasm_cc_binary files

The output paths (and extname) are known in the rule with `ctx.outputs.outputs`. Would creating a [Provider](https://bazel.build/rules/lib/Provider) for each type of output (or maybe groups of outputs?) be flexible enough for rules consuming the `wasm_cc_binary` files?

Alternatively since the outputs are explicit they could also explicitly be passed into another rule, correct?

bazel/emscripten_toolchain/wasm_cc_binary.bzl
zaucy: I added a default outputs when the `outputs` attribute is empty. That does mean that #986 would still be broken (maybe?) when the outputs are empty.

walkingeyerobot: Sadly this is still broken. To reproduce, create a new folder next to hello-world called "foo", add a BUILD file with these contents:
```bazel
filegroup(
    name = "foo",
    srcs = ["//hello-world:hello-world.wasm"],
)
```
Because `hello-world.wasm` is not an explicit output, the build fails. If you add it under the `outputs` attribute, this works (although you might need to mess with visibility or pass `--nocheck_visibility`).

zaucy: Ah I see. Unfortunately simply wrapping the rule in a macro for default outputs wouldn't work either since currently on the `main` branch the outputs depend on the `cc_target`'s name, not just the `wasm_cc_binary`'s name.

If the rule was wrapped in a macro and it's okay to break the default output paths not considering `cc_target`s name then that's great! Otherwise there could be _two_ internal rules that the macro chooses from.

I'm capable of doing either option. If you have a preference let me know and I'll make the necessary changes.

walkingeyerobot: You're right, this is more tricky than I initially thought.

I think we should parse the `cc_target` label in the macro to keep existing behavior as much as possible. Inside the macro, you should just be able to treat the `cc_target` as a string. I acknowledge that if someone is `select()`ing on their `cc_target` this will break as the `select()`s are not resolved when the macro is resolving. I don't see a way to avoid breaking that, but I don't believe the number of users affected would be very high.

I still believe this change is quite valuable overall. Thank you for your patience and effort here.

trybka:
> The output paths (and extname) are known in the rule with ctx.outputs.outputs. Would creating a [Provider](https://bazel.build/rules/lib/Provider) for each type of output (or maybe groups of outputs?) be flexible enough for rules consuming the wasm_cc_binary files?

They are only known if specifying the attribute (i.e. non-empty)--this doesn't help with supporting pre-existing legacy usecase(s) where folks are just depending on these implicit outputs existing. 

e.g. you could do the following (and people do this):

```starlark
genrule(
    name = "compile_js",
    srcs = ["//mywasm:foo.js", "mycustomfooter.js"],
    outs = ["compiled.js"],
    cmd = ["cat $(SRCS) > $(OUTS)"]
)
```

Providers for specific use-cases are a fine extension to add for, e.g. `js_library`, but I don't think they help here. 

`OutputGroupInfo` is a provider specifically for doing different file outputs, but still doesn't solve our pre-declared output file problem, I'm afraid. 

> Alternatively since the outputs are explicit they could also explicitly be passed into another rule, correct?

Yes. See above.

bazel/emscripten_toolchain/wasm_cc_binary.bzl
zaucy: I went ahead and created a macro that decies to use either a "legacy" rule or the new rule. I thought that might be better than parsing the `cc_target` label in the macro. https://github.com/emscripten-core/emsdk/pull/1047/commits/ce1cea7d8cbe2722503c945948afe59aa8a74059

bazel/emscripten_toolchain/wasm_cc_binary.bzl
```diff
@@ -105,6 +191,12 @@ def _wasm_binary_outputs(name, cc_target):
 
     return outputs
 
+_wasm_cc_binary_legacy = rule(
```
trybka, line 194: can you set `name = "wasm_cc_binary"` for both of these? 

that way the target type matches for things that care about the rule type.

zaucy: Done! I didn't know that existed. Thanks!

Two questions if you don't mind answering:
1) Is it not a problem that both `_wasm_cc_binary_legacy` and `_wasm_cc_binary` have the same `name`?
2) Is it recommended that the rule `name` attribute is set for any rule that is wrapped in this style? I'd like to update my internal rules if that's the case. I'm unsure what the consequences of not doing it is.

trybka: 1) This is basically the only use for the `name` parameter AFAIK--when you need to do something spicy like this.
See also: [cc_test.bzl's usage](https://github.com/bazelbuild/bazel/blob/fc5bf5459a52d1ea68b3dcd9fa582de118e452d1/src/main/starlark/builtins_bzl/common/cc/cc_test.bzl#L116-L141)

2) I dunno if it's "recommended" or not, but when you care about the rule type being the same (as, in this case, we do -- for things like `bazel query "kind(wasm_cc_binary...)"` or `native.existing_rules`), you have two options: 
  a) either use `name` or 
  b) do some file import dance like [cc_binary does](https://github.com/bazelbuild/bazel/blob/fc5bf5459a52d1ea68b3dcd9fa582de118e452d1/src/main/starlark/builtins_bzl/common/cc/cc_binary_wrapper.bzl#L21-L37). `cc_binary` does this thing where it defines the rules in separate files so they can both be named `cc_binary`, but then imports them with a different name to do this macro wrap. 

I personally prefer using `name`, but I suppose both are _correct_ so it might just be a matter of style.

zaucy:
I've cherry picked #1049 to get past the macOS issue (I can remove if this PR gets approved) and I also rebased onto 891b4491419c42f9d2b9f97c47d1043b26dfd3e5 because the latest on main (93f21c9ef30bab52de24f9d4ea3f2f377cf6326a) linux bazel is failing.

walkingeyerobot (approved):

PR #1049, opened 2022-05-17
Using bazelisk on macOS CI

zaucy:
Using `bazelisk` instead of an outdated version of bazel on macOS should unblock CI failures for #1036 and #1047

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -201,9 +201,9 @@ jobs:
       - checkout
       - run: brew install grep
       - run:
-          name: install bazel
+          name: install bazelisk
           command: |
-            brew install bazel
+            brew install bazelisk
       - run: test/test_bazel_mac.sh
 
   test-bazel-windows:


sbc100:
What is bazelisk and why is the bazel package outdated?

zaucy:
[https://github.com/bazelbuild/bazelisk](bazelisk) automatically downloads and uses the latest version of bazel or a version of bazel specified by a `.bazelversion` file or environment variables. There's more as well, but I'd recommend reading the GitHub repository for more. Bazelisk is also mentioned in the official bazel documentation: https://docs.bazel.build/versions/main/install-bazelisk.html

I'm not sure why the bazel package is outdated on brew on the machine circle ci is configured with.

walkingeyerobot (approved):
@sbc100 do you have any other concerns over merging this?

sbc100:
Just the test failure.. do you know what up with that?

PR #1052, opened 2022-05-19
include node modules glob with linker files. 

walkingeyerobot:
Also some minor formatting fixes.

This fixes the bazel build.


bazel/emscripten_deps.bzl
```diff
@@ -40,7 +40,7 @@ filegroup(
         "bin/wasm-emscripten-finalize{bin_extension}",
         "bin/wasm-ld{bin_extension}",
         "bin/wasm-opt{bin_extension}",
-   ],
+   ] + glob(["emscripten/node_modules/**"]),
```
sbc100, line 43: BTW, I still don't understand how the linker can work without the libraries in `emscripten/cache/sysroot/lib`

sbc100 (approved):

PR #1057, opened 2022-06-06
Add additional files necessary for building with closure and on RBE

kjlubick:
With emsdk 3.1.13, I was unable to use Bazel compile [CanvasKit](https://github.com/google/skia/blob/main/modules/canvaskit/BUILD.bazel) locally in release mode or remotely in either debug or release mode.

Some files were missing from the specification, as [mentioned here](https://github.com/emscripten-core/emsdk/pull/1045#issuecomment-1121444545).

By specifying a few more files, it now works locally and remotely, while still being significantly faster than 3.1.0 (before #1045 landed).

### DIFF
--- a/bazel/emscripten_deps.bzl
+++ b/bazel/emscripten_deps.bzl
@@ -70,6 +70,7 @@ filegroup(
             "emscripten/src/**",
         ],
     ),
+)
 
 filegroup(
     name = "ar_files",
--- a/bazel/emscripten_deps.bzl
+++ b/bazel/emscripten_deps.bzl
@@ -19,11 +19,32 @@ filegroup(
 )
 
 filegroup(
+    name = "emcc",
+    srcs = [
+        "emscripten/emcc.py",
+        "emscripten/emscripten.py",
+        "emscripten/emscripten-version.txt",
+        "emscripten/cache/sysroot_install.stamp",
+        "emscripten/src/settings.js",
+        "emscripten/src/settings_internal.js",
+    ] + glob(
+        include = [
+            "emscripten/third_party/**",
+            "emscripten/tools/**",
+        ],
+        exclude = [
+            "**/__pycache__/**",
+        ],
+    ),
+)
+
+filegroup(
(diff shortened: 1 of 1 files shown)


bazel/emscripten_deps.bzl
```diff
@@ -40,15 +61,33 @@ filegroup(
...
         "bin/wasm-ld{bin_extension}",
         "bin/wasm-opt{bin_extension}",
-   ] + glob(["emscripten/node_modules/**"]),
+        "bin/wasm-metadce{bin_extension}",
+        ":emcc",
+    ] + glob(
+        include = [
+            "emscripten/cache/sysroot/lib/**",
+            "emscripten/node_modules/**",
+            "emscripten/src/**",
```
sbc100, line 70: emcc.py depends on stuff in `tools` so I'm surprised not to see it here.  

I think `tools` is needed be all steps (since it contains a lot of shared python code that all entry points depend on) so I would imagine it would need to go in some kind of "common" fileset?

sbc100: Oh I see `emcc` is that name for this shared set... In that case you can remove `emcc.py` from this set?

kjlubick: Done.

bazel/emscripten_deps.bzl
```diff
@@ -40,15 +59,33 @@ filegroup(
...
-        "emscripten/emar.py",
         "bin/llvm-ar{bin_extension}",
-   ],
+        "emscripten/emar.py",
+        "emscripten/emscripten-version.txt",
+        "emscripten/src/settings.js",
+        "emscripten/src/settings_internal.js",
+    ] + glob(
+        include = [
+            "emscripten/tools/**",
```
sbc100, line 83: If you want to you could probably exclude `tools/mains/`, `tools/experimental`, `tools/scons`, and `tools/websocket_to_posix_proxy/`, although they are not very big.

sbc100 (approved):

PR #1078, opened 2022-07-23
Exclude msys from path fix function.

5cript:
The `//?/` prefix does not work in the MSYS environment.
Excluding MSYS from the windows fix worked for me to install emscripten.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -529,6 +529,8 @@ def untargz(source_filename, dest_dir):
 def fix_potentially_long_windows_pathname(pathname):
   if not WINDOWS:
     return pathname
+  if MSYS:
+    return pathname
   # Test if emsdk calls fix_potentially_long_windows_pathname() with long
   # relative paths (which is problematic)
   if not os.path.isabs(pathname) and len(pathname) > 200:
--- a/emsdk.py
+++ b/emsdk.py
@@ -527,9 +527,7 @@ def untargz(source_filename, dest_dir):
 # See https://msdn.microsoft.com/en-us/library/aa365247.aspx#maxpath and http://stackoverflow.com/questions/3555527/python-win32-filename-length-workaround
 # In that mode, forward slashes cannot be used as delimiters.
 def fix_potentially_long_windows_pathname(pathname):
-  if not WINDOWS:
-    return pathname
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -529,6 +529,8 @@ def untargz(source_filename, dest_dir):
 def fix_potentially_long_windows_pathname(pathname):
   if not WINDOWS:
     return pathname
+  if MSYS:
```
sbc100, line 532: Combine this with the above block?  `if not WINDOWS or MSYS:`?

sbc100: Do you know what happens if the path length is above 260 chars?  This prefix seems to specifically be about issues with  such long filenames.

sbc100 (commented):
Thanks!

I wonder if we can find a way to test under MSYS as part of CI?

emsdk.py
5cript: It fails either way, even if I change line 539 from `if MINGW:` to `if MINGW or MSYS:` so that it no longer fails with "cannot find the file specified". It then still fails with "The filename or extension is too long: '//?/D:/Development/emsdkdeep/abcdefghijklmnopqrstuvxyz/abcdefghijklmnopqrstuvxyz/abcdefghijklmnopqrstuvxyz/abcdefghijklmnopqrstuvxyz/abcdefghijklmnopqrstuvxyz/abcdefghijklmnopqrstuvxyz/abcdefghijklmnopqrstuvxyz/emsdk/node/unzip_temp/node-v14.18.2-win-x64'"

So the fix for windows does not seem to work for the msys environment

5cript: There is some hidden layer that messes with paths but does not understand the prefix that is intended to circumvent MAX_PATH. There is also documentation on msys2 on path related conversions: https://www.msys2.org/docs/filesystem-paths/#windows-unix-path-conversion

But I failed to devices a fix either by configuration or by code that is able to handle the prefix / long paths.

sbc100: Perhaps on MSYS we should just error out here bigger than MAX_PATH?  Since `cannot find the file specified` is less useful of an error message.

Happy to land this change either way.

sbc100 (approved):

Issue #1079, opened 2022-07-26
emsdk_env.sh doesn't work on cygwin due to wrong PATH parsing


Julianiolo:
the `emsdk_env.sh` script produces a jumbled up `PATH` when used under cygwin.

I was able to track the problem down to the wrong `ENVPATH_SEPERATOR` value beeing picked in `emsdk.py` (line 126-129).
The standard cygwin shell is bash so it uses ':' instead of ';' to seperate it's `PATH`, but here it is parsed as if it were using ';'.

I'm not sure how to fix it, I added an extra `and not MSYS` to the `if WINDOWS:` line(line 126), and that seemed to work, but I'm not sure if that's the optimal solution.

I'm also not sure what other things might me affected by this problem, everything else besides the mentioned `emsdk_env.sh` worked for me.

sbc100:
Feel free to send patches to improve the cygwin support.  Sadly we don't have any testing of the cygwin environment, so one thing that would be very useful is adding cynwin to the github CI setup.

Julianiolo:
#1080 :)

sbc100:
Are you using the cygwin version or python.exe or a windows version of python.exe?    Because in CI we are running under bash.exe, but we are using are native windows version of python.exe (which reports os.environ['PATH'] as with windows path separators)

Julianiolo:
I'm using the python from cygwin (as you might see in the PR comment), specifically python 3.9.10 on cygwin

PR #1080, opened 2022-07-26
fix to #1079

Julianiolo:
This isn't the exact fix described in the issue, but I figured it would make more sense to test if we are using bash?

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -123,7 +123,7 @@ if not CSH and not POWERSHELL and not BASH and not CMD:
   else:
     BASH = True
 
-if WINDOWS:
+if WINDOWS and not BASH:
   ENVPATH_SEPARATOR = ';'
 else:
   ENVPATH_SEPARATOR = ':'


sbc100:
Interesting.. it looks like we already have special case in place for when we actually construct the path:  https://github.com/emscripten-core/emsdk/blob/5ad3ff05a79781256e88cde3829c5b0348be861f/emsdk.py#L2652-L2653

We are running the CI here with windows python running under bash.exe.. and it seems that python.exe (the windows version) is sees an incoming path that has windows separators.

Julianiolo:
Oh Interesting, yeah then maybe a special check like that should be used instead here:
https://github.com/emscripten-core/emsdk/blob/5ad3ff05a79781256e88cde3829c5b0348be861f/emsdk.py#L2623

That means that this line shouln't work for msys though, right?

Julianiolo:
Another idea I had was to just try both or check to see if we find a ':' or ';' in the PATH string, but I guess that may be not that robust.

sbc100:
> Another idea I had was to just try both or check to see if we find a ':' or ';' in the PATH string, but I guess that may be not that robust.

I wonder if just `os.name == 'nt'` here?  This seems to suggest that would work because cynwin pyhton reports `os.name` as `posix`: https://stackoverflow.com/a/7637706

Can you confirm if that works here?

sbc100:
> > Another idea I had was to just try both or check to see if we find a ':' or ';' in the PATH string, but I guess that may be not that robust.
> 
> I wonder if just `os.name == 'nt'` here? This seems to suggest that would work because cynwin pyhton reports `os.name` as `posix`: https://stackoverflow.com/a/7637706
> 
> Can you confirm if that works here?

i.e. what does os.name and os.pathsep return for your cygwin version of python?

sbc100:
> > > Another idea I had was to just try both or check to see if we find a ':' or ';' in the PATH string, but I guess that may be not that robust.
> > 
> > 
> > I wonder if just `os.name == 'nt'` here? This seems to suggest that would work because cynwin pyhton reports `os.name` as `posix`: https://stackoverflow.com/a/7637706
> > Can you confirm if that works here?
> 
> i.e. what does os.name and os.pathsep return for your cygwin version of python?

Maybe we don't need ENVPATH_SEPARATOR at all and we can just use `os.pathsep`?

sbc100:
> > > Another idea I had was to just try both or check to see if we find a ':' or ';' in the PATH string, but I guess that may be not that robust.
> > 
> > 
> > I wonder if just `os.name == 'nt'` here? This seems to suggest that would work because cynwin pyhton reports `os.name` as `posix`: https://stackoverflow.com/a/7637706
> > Can you confirm if that works here?
> 
> i.e. what does os.name and os.pathsep return for your cygwin version of python?

Also, what does `os.environ['PATH']` report in your version of python?

Julianiolo:
Sorry that I took so long to answer ':)

as to your ideas, I think that might be the way to go.

the python related stuff you asked:
```
$ python
Python 3.9.10 (main, Jan 20 2022, 21:37:52) 
[GCC 11.2.0] on cygwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.name
'posix'
>>> os.pathsep
':'
>>> os.environ['PATH']
'/usr/local/bin:/usr/bin:/cygdrive/d/Programme/General/CUDA/dev/bin:/cygdrive/d/Programme/General/CUDA/dev/libnvvp:/cygdrive/c/Program Files/Common Files/Oracle/Java/javapath:/cygdrive/c/Program Files 
(x86)/Common Files/Oracle/Java/javapath:/cygdrive/c/Windows/system32:/cygdrive/c/Windows:/cygdrive/c/Windows/System32/Wbem:/cygdrive/c/Windows/System32/WindowsPowerShell/v1.0:/cygdrive/c/Windows/System32/OpenSSH:/cygdrive/c/Program Files (x86)/GtkSharp/2.12/bin:/cygdrive/c/WINDOWS/system32:/cygdrive/c/WINDOWS:/cygdrive/c/WINDOWS/System32/Wbem:/cygdrive/c/WINDOWS/System32/WindowsPowerShell/v1.0:/cygdrive/c/WINDOWS/System32/OpenSSH:/cygdrive/c/Program Files (x86)/NVIDIA Corporation/PhysX/Common:/cygdrive/c/Program Files/NVIDIA Corporation/NVIDIA NvDLISR:/cygdrive/c/Program Files/PuTTY:/cygdrive/c/Program Files/nodejs:/cygdrive/c/cmake-3.19.0-rc2-win64-x64/bin:/cygdrive/c/Git/cmd:/usr/bin:/cygdrive/c/Program Files/dotnet:/cygdrive/c/Program Files/Liquid Technologies/Liquid Studio 2021/XmlDataBinder19/Redist19/cpp/win32/bin:/cygdrive/c/Program Files/Liquid Technologies/Liquid Studio 2021/XmlDataBinder19/Redist19/cpp/win64/bin:/cygdrive/d/Programme/General/UltraEdit:/cygdrive/c/Users/Julian/Desktop/Dateien/programms/ffmpeg-2021-10-07-git-b6aeee2d8b-full_build/bin:/cygdrive/c/WINDOWS/system32:/cygdrive/c/WINDOWS:/cygdrive/c/WINDOWS/System32/Wbem:/cygdrive/c/WINDOWS/System32/WindowsPowerShell/v1.0:/cygdrive/c/WINDOWS/System32/OpenSSH:/cygdrive/c/Program Files/NVIDIA Corporation/Nsight Compute 2022.1.1:/cygdrive/d/Programme/General/Graphviz/bin:/cygdrive/c/Users/Julian/Desktop/Dateien/programms/ffmpeg-2022-06-01-git-c6364b711b-essentials_build/bin:/cygdrive/c/Users/Julian/Desktop/Dateien/programms/emsdk:/cygdrive/c/Users/Julian/Desktop/Dateien/programms/emsdk/upstream/emscripten:/cygdrive/c/Users/Julian/Desktop/Dateien/programms/emsdk/node/14.18.2_64bit/bin:/cygdrive/c/Users/Julian/AppData/Local/Programs/Python/Python38/Scripts:/cygdrive/c/Users/Julian/AppData/Local/Programs/Python/Python38:/cygdrive/c/Users/Julian/AppData/Local/Microsoft/WindowsApps:/cygdrive/c/Users/Julian/AppData/Local/atom/bin:/cygdrive/c/Users/Julian/AppData/Local/Programs/Microsoft VS Code/bin:/cygdrive/c/Users/Julian/AppData/Local/GitHubDesktop/bin:/cygdrive/c/Users/Julian/AppData/Roaming/npm:%USERPROFILE%/AppData/Local/Microsoft/WindowsApps'
>>>
```

so yeah I guess that should work :)

Julianiolo:
I didn't know `os.pathsep` was a thing, but if it exists, maybe we should just use it?

sbc100:
> I didn't know `os.pathsep` was a thing, but if it exists, maybe we should just use it?

One problem is that sometimes folks use windows python inside a mingw shell.   This means `os.pathsep` will be `;` for windows but create_env needs to build a path that is suitable for the mingw shell.

So, we have two different path setups

1. The paths uses in the incoming `os.environ['PATH'`.  This depends I think on whether we are using windows python of cygwin python.  Here we always want to use `os.pathsep`.
2. The path convention used by the calling shell.  i.e. are we using cmd.exe or bash.exe to run `create_env`.   This could be different to the `os.pathsep` used by python.exe.

Julianiolo:
What do you mean by create_env exactly?

sbc100:
Sorry I meant `construct_env`.   When you run `emsdk_env` it called `emsdk.py construct_env` to build a set of environment variables for your shell.

Julianiolo:
what bash do you mean in the context of mingw, the one from git? Or does mingw has its own bash? I'm not too well versed with mingw ':)

Julianiolo:
if thats the bash you mean, then yeah there is defenitly a problem as python says 'nt' for name and pathsep is ';' so that wont work for constructing a path, since this bash uses ':'.

Currently I would think that maybe we should do the parsing of the current path with os.pathsep and then for constructing we still need to somehow find the right seperator.

Julianiolo:
one idea I have is to maybe just set an environment variable in the activate.bat file and then check for that in python? The bat file should be the only case where a ';' seperated path is needed, right?

PR #1097, opened 2022-08-25
Add double quotes to allow spaces in path

joelvaneenwyk:
Minimal fixes to allow spaces in path when using shell scripts. #1096

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -8,14 +8,14 @@
 
 # First look for python bundled in Emsdk
 if [ -z "$EMSDK_PYTHON" ]; then
-  PYTHON3=$(dirname $0)/python/3.9.2-1_64bit/bin/python3
-  PYTHON3_CERT_FILE=$(dirname $0)/python/3.9.2-1_64bit/lib/python3.9/site-packages/certifi/cacert.pem
-  if [ ! -f $PYTHON3 ]; then
-    PYTHON3=$(dirname $0)/python/3.7.4-2_64bit/bin/python3
-    PYTHON3_CERT_FILE=$(dirname $0)/python/3.7.4-2_64bit/lib/python3.7/site-packages/certifi/cacert.pem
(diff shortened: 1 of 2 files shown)


joelvaneenwyk:
@sbc100, it looks like you have done some testing of this in the past based on your https://github.com/emscripten-core/emsdk/tree/test_spaces branch. Any thoughts on this one? This change is pretty safe as-is, but I could also update CI tests to include something on this if folks are open to it.

emsdk
```diff
@@ -26,15 +26,14 @@ if [ -z "$EMSDK_PYTHON" ]; then
...
+    export SSL_CERT_FILE="$PYTHON3_CERT_FILE"
   fi
 fi
 
 # If bundled python is not found, look for `python3` in PATH.  This is especially important on macOS (See:
 # https://github.com/emscripten-core/emsdk/pull/273)
 if [ -z "$EMSDK_PYTHON" ]; then
-  PYTHON3=$(which python3 2> /dev/null)
-  if [ $? = 0 ]; then
+  if PYTHON3="$(which python3 2>/dev/null)"; then
```
sbc100, line 36: Are you sure its safe to combine these two lines?

sbc100 (approved):
lgtm!  (with one comment)

sbc100:
The mac CI faliures are unrelated and should be fixed by #1084

emsdk
joelvaneenwyk: Yes, it is expected to be safe across all POSIX compliant shells, but I did do some digging after you asked just to see if I could find any exceptions to this and was not able to. It seems consistent that combining the command with the check is the recommended approach. More juicy/useful info on the this at [ShellCheck: SC2181 – Check exit code directly with e.g. `if mycmd;`, not indirectly with `$?`.](https://www.shellcheck.net/wiki/SC2181). Thanks for reviewing this!

PR #1105, opened 2022-09-15
Add ability to install prebuilt ninja

sbc100:
We may start using it internally in emscripten. See https://github.com/emscripten-core/emscripten/pull/17809

Its also useful to be able to install it when building other things from source (See #1736)

### DIFF
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -625,6 +625,27 @@
   },
   {
     "id": "ninja",
+    "version": "1.13.2",
+    "bitness": 64,
+    "windows_url": "ninja-1.13.2-win-x64.zip",
+    "linux_url": "ninja-1.13.2-linux-x64.zip",
+    "macos_url": "ninja-1.13.2-mac.zip",
+    "arch": "x86_64",
+    "activated_cfg": "NINJA_ROOT='%installation_dir%'",
+    "activated_path": "%installation_dir%"
+  },
+  {
+    "id": "ninja",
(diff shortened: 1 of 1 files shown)


dschuff:
this looks fine, although we probably don't need to land it until/unless we decide to take the dependence for external developers

emsdk_manifest.json
```diff
@@ -623,6 +623,16 @@
     "activated_cfg": "MINGW_ROOT='%installation_dir%'",
     "activated_path": "%installation_dir%/bin"
   },
+  {
+    "id": "ninja",
+    "version": "1.11.1",
+    "bitness": 64,
+    "windows_url": "ninja-1.11.1-win.zip",
```
juj: The artifact would be good to be named x64 here. (I presume the Windows artifact is x64 and not arm64). Does this work with arm64? Can we bundle a separate windows arm64 version?

emsdk_manifest.json
```diff
@@ -623,6 +623,16 @@
...
+    "linux_url": "ninja-1.11.1-linux.zip",
+    "macos_url": "ninja-1.11.1-mac.zip",
```
juj: Likewise here for Linux x64 vs arm64, and mac x64 vs arm64?

dschuff:
So is the plan to just switch to depending on ninja to build the libraries, and delete all the python that does the same thing?

sbc100:
Yes, see #1742

PR #1115, opened 2022-10-03
Use x64 version for Windows on Arm

pbo-linaro:
At Linaro, [Windows on Arm](https://www.linaro.org/windows-on-arm/) team is helping open source projects to enable this platform. We spent some time to work on dart-sdk.

A few days ago, a regression was found. [Dart](https://github.com/dart-lang/sdk) started to use emscripten for their dart2wasm module.

This fail to build under Windows on Arm, because emsdk can't find a toolchain to download (when having an arm64 python on the PATH). This PR adds a fallback to download an x64 version instead (supported via emulation from windows 11).

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -142,6 +142,9 @@ elif machine.endswith('86'):
   ARCH = 'x86'
 elif machine.startswith('aarch64') or machine.lower().startswith('arm64'):
   ARCH = 'aarch64'
+  if WINDOWS:
+      errlog('No support for Windows on Arm, fallback to x64')
+      ARCH = 'x86_64'
 elif machine.startswith('arm'):
   ARCH = 'arm'
 else:


emsdk.py
```diff
@@ -142,6 +142,9 @@ def exit_with_error(msg):
   ARCH = 'x86'
 elif machine.startswith('aarch64') or machine.lower().startswith('arm64'):
   ARCH = 'aarch64'
+  # no support for Windows on Arm yet, fallback to x64
+  if WINDOWS:
```
sbc100, line 146: Can you print a warning here (using the `errlog` function)?

sbc100: Also, perhaps we should open a bug about missing windows arm64 version of emsdk and reference the bug here?

pbo-linaro: > Can you print a warning here (using the `errlog` function)?

That's better yes! (added in latest version)

pbo-linaro: > 

Yes sure, I just wanted to know first if that workaround would be acceptable.

pbo-linaro:
Created issue here: https://github.com/emscripten-core/emsdk/issues/1116

emsdk.py
sbc100: I think you folks are only current users we have on arm64 windows so its mostly up to you if you want to consider this is long or short term solution.  What is the typical performance overhead of x86 emulation?

sbc100 (approved):

pbo-linaro:
@sbc100 Is something missing for that PR? Thanks, Pierrick

PR #1131, opened 2022-11-09
remove "name" attribute from bazel rules

luxe:
The `name` attribute for [rule](https://bazel.build/extending/rules#dependency_attributes) will be deprecated in future bazel releases.
See:  [incompatible_remove_rule_name_parameter](https://github.com/bazelbuild/bazel/issues/16301)
To prepare for bazel 6, let's remove the attribute.


walkingeyerobot (approved):

sbc100:
I just landed it with prejudice since the failure looked completely unrelated.

PR #1135, opened 2022-11-20
Change base to Ubuntu 22.04 LTS (jammy)

forderud:
Done to upgrade CMake from [3.16.3](https://packages.ubuntu.com/focal/cmake) to [3.22.1](https://packages.ubuntu.com/jammy/cmake).

Motiation: [CMake 3.21 or newer](https://doc.qt.io/qt-6/cmake-get-started.html) is needed to build the Qt 6 sources with emscripten.

### CI problem
CI is failing with a code signing error. I suspect that this is due to https://github.com/docker/docker-install/issues/244, which is fixed by updating the  Docker version on the CI node.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:focal AS stage_build
+FROM ubuntu:jammy AS stage_build
 
 ARG EMSCRIPTEN_VERSION=tot
 ENV EMSDK /emsdk
@@ -56,7 +56,7 @@ RUN echo "## Aggressive optimization: Remove debug symbols" \
 # -------------------------------- STAGE DEPLOY --------------------------------
 # ------------------------------------------------------------------------------
 
-FROM ubuntu:focal AS stage_deploy
+FROM ubuntu:jammy AS stage_deploy
 
 COPY --from=stage_build /emsdk /emsdk
 
@@ -105,7 +105,7 @@ RUN echo "## Update and install packages" \
         build-essential \
         make \
         ant \
-        libidn11 \
+        libidn12 \
         cmake \
         openjdk-11-jre-headless \
     # Standard Cleanup on Debian images


.circleci/config.yml
```diff
@@ -141,7 +141,7 @@ jobs:
           name: install docker
           command: apt-get update -q && apt-get install -q -y docker.io
       - setup_remote_docker:
-          version: 19.03.13
+          version: 20.10.21
```
kleisauke, line 144: Could you try this?:
```suggestion
          version: 20.10.18
```

According to https://circleci.com/docs/building-docker-images/#docker-version, that appears to be the latest version available.

Note that I just cancelled the [` build-docker-image` ](https://circleci.com/gh/emscripten-core/emsdk/9824) job on this PR, as it's stuck almost 17 hours now (probably because of this).

forderud: Thanks for the tip @kleisauke. I've now tried to update the PR accordingly. Please let me know if you also want me to decouple the Docker engine update from the Ubuntu upgrade by submitting it in a separate PR that will have to be merged _before_ this one.

kleisauke: Great! It seems now to fail due to `Unable to locate package libidn11`, I think replacing `libidn11` in the Dockerfile with `libidn12` would fix that.

> Please let me know if you also want me to decouple the Docker engine update from the Ubuntu upgrade by submitting it in a separate PR that will have to be merged _before_ this one.

I think it's probably fine to do this in the same PR.

kleisauke: ... although, perhaps we can drop that dependency, I'm not sure why it was added in the first place (maybe it was needed by OpenJDK 11, which is needed by Closure Compiler?). Let me investigate that as a possible follow-up.

kleisauke (approved):
LGTM! I tested this locally on the [wasm-vips](https://github.com/kleisauke/wasm-vips) project without any issues using:
```bash
$ docker build --network host --build-arg=EMSCRIPTEN_VERSION=3.1.26 -t emscripten/emsdk:3.1.26 -f docker/Dockerfile .
$ cd ~/wasm-vips
# Ensure we use the locally build Docker image
$ sed -i 's/docker build -t wasm-vips ./& --pull=false/' package.json
$ npm run build
$ npm run test
```

@sbc100 Feel free to merge if it looks good to you. Perhaps we should mention this in Emscripten's `ChangeLog.md`, something along?:
```
- The `emscripten/emsdk` Docker base image was updated from Ubuntu 20.04 LTS
  (Focal) to Ubuntu 22.04 LTS (Jammy). (emscripten-core/emsdk#1135)
```

forderud:
I've now rebased the PR on top of main after https://github.com/emscripten-core/emsdk/pull/1117 was merged.

sbc100 (approved):

forderud:
@kleisauke @sbc100 Is it possible to request release of a new Docker image that includes the Ubuntu 22.04 update?

sbc100:
We normally do a release ever week or two.. it would be simplest if you could just wait for the next emsdk release?

In the interim,  isn't it possible to simply point of the Dockerfile in this repo (i.e. just use it directly form here without needing a publication of dockerhub?)  I'm not docker expert but I seem to remember it fairly easy to just directly use a dockerfile from git?

forderud:
> We normally do a release ever week or two.. it would be simplest if you could just wait for the next emsdk release?

I can live with that. Anyhow, thanks for your willingness to merge the Ubuntu upgrade so quickly. :-)

> In the interim, isn't it possible to simply point of the Dockerfile in this repo (i.e. just use it directly form here without needing a publication of dockerhub?) I'm not docker expert but I seem to remember it fairly easy to just directly use a dockerfile from git?

Thanks for the tip. Will investigate it.

PR #1141, opened 2022-12-01
--skip-npm flag to Skip putting emsdk's npm on PATH

### LINKED ISSUE #1142: Installing emsdk overwrites node on PATH
This is a small but annoying issue. When installing the emscripten sdk, or whenever you install a new version of the toolchain, you're told to execute `source ./emsdk_env.sh`.

This is rather annoying, as the toolchain comes packaged with its own specific version of node, and this conflicts with all other node tools we're used to using. For example, it put node 4.1.1 or my path while I previously had 7.4.0. I've noticed the same goes for `clang`, but it doesn't bother my that much. Even when relying on a tool like nvm, it's annoying because switching to another version would essentially break emscripten/emsdk?

If emscripten is reliant on a specific version of node, should it not be its own responsibility to make sure its scripts run with its own internally packaged version? This just seems a tad messy.

(**Note:** One possible temporary workaround is to simply avoid `emsdk_env.sh` in this case and just manually add `emsdk/upstream/emscripten` to your path.  That is all that should be needed to get emscripten working on many systems).

shlomif:
Fixes: #1142


sbc100 (commented):
I don't know if this patch aligns with the approch that we discuessed in the bug, but if we do take this approach I think you would just want to modify `get_required_path`.   We still want to install and activate our own version of node, we just don't want to add it to the users path, right?

shlomif:
@sbc100 : thx. i fixed some low-hanging fruit

sbc100:
I think this can be closed now that #1189 has landed emsdk will no longer add its node to the PATH if you already have one .

PR #1145, opened 2022-12-02
Add standalone_wasm feature to bazel emscripten_toolchain

robbertvanginkel:
This allows building with `--features=standalone_wasm` into a WASI binary.

### DIFF
--- a/bazel/emscripten_toolchain/toolchain.bzl
+++ b/bazel/emscripten_toolchain/toolchain.bzl
@@ -441,6 +441,9 @@ def _impl(ctx):
             name = "output_format_js",
             enabled = True,
         ),
+        feature(
+            name = "standalone_wasm",
+        ),
     ]
 
     crosstool_default_flag_sets = [
@@ -1014,6 +1017,11 @@ def _impl(ctx):
             actions = all_compile_actions,
             flags = ["-Werror"],
             features = ["wasm_warnings_as_errors"],
+        ),
+        flag_set(
+            actions = all_link_actions,
+            flags = ["-s", "STANDALONE_WASM"],
+            features = ["standalone_wasm"],
         ),
     ]
 
--- a/bazel/emscripten_toolchain/toolchain.bzl
+++ b/bazel/emscripten_toolchain/toolchain.bzl
@@ -442,7 +442,7 @@ def _impl(ctx):
             enabled = True,
         ),
         feature(
-            name = "standalone_wasm",
+            name = "wasm_standalone",
         ),
     ]
 
@@ -1021,7 +1021,7 @@ def _impl(ctx):
         flag_set(
             actions = all_link_actions,
             flags = ["-sSTANDALONE_WASM"],
-            features = ["standalone_wasm"],
+            features = ["wasm_standalone"],
         ),
     ]
 
--- a/bazel/emscripten_toolchain/toolchain.bzl
+++ b/bazel/emscripten_toolchain/toolchain.bzl
@@ -1020,7 +1020,7 @@ def _impl(ctx):
         ),
         flag_set(
             actions = all_link_actions,
-            flags = ["-s", "STANDALONE_WASM"],
+            flags = ["-sSTANDALONE_WASM"],
             features = ["standalone_wasm"],
         ),
     ]
--- a/bazel/emscripten_toolchain/wasm_cc_binary.bzl
+++ b/bazel/emscripten_toolchain/wasm_cc_binary.bzl
@@ -25,6 +25,9 @@ def _wasm_transition_impl(settings, attr):
     if attr.simd:
         features.append("wasm_simd")
 
+    if attr.standalone:
+        features.append("wasm_standalone")
+
     return {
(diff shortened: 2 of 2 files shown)


bazel/emscripten_toolchain/toolchain.bzl
```diff
@@ -1015,6 +1018,11 @@ def _impl(ctx):
             flags = ["-Werror"],
             features = ["wasm_warnings_as_errors"],
         ),
+        flag_set(
+            actions = all_link_actions,
+            flags = ["-s", "STANDALONE_WASM"],
```
sbc100, line 1023: No need to two arg here.  You can just do `"-sSTANDALONE_WASM"`

sbc100 (commented):
lgtm, but will leave final approval to @walkingeyerobot

robbertvanginkel:
👋 first time contributor here, I've looked for any contributing instructions but didn't find anything beyond https://emscripten.org/docs/contributing/developers_guide.html, please let me know if I missed anything (particularly tests).

This would make building binaries targetting WASI easier from emsdk. 

I'm wondering if we could somehow enable this by default when targetting `@platforms//os:wasi` from https://github.com/bazelbuild/platforms/blob/3fbc687756043fb58a407c2ea8c944bc2fe1d922/os/BUILD#L98-L103, so far I've found that the [`@emsdk//:platform_wasm`](https://github.com/emscripten-core/emsdk/blob/e8606a7e3bfb86e86831f43db19ee639f3996532/bazel/emscripten_toolchain/wasm_cc_binary.bzl#L35) platform in `wasm_cc_binary` defaults to the cpu only
 https://github.com/emscripten-core/emsdk/blob/005063304bb4e575f97b7c5f50368f937c06cb54/bazel/BUILD#L70-L75.

sbc100:
emsdk only supports the emcc compiler, which int turn only supports building with `wasm32-emscripten` as the triple.   We can still build wasi-compatible modules, which is perhaps enough?

robbertvanginkel:
Thanks for that context, still learning the emscripten toolchain so this is helpful!

I'm interested in wasi as an OS since I found a constraint like:
```
select({
    "@platforms//os:wasi": [],
    "//conditions:default": ["-pthread"],
})
```
while compiling re2: https://github.com/google/re2/blob/5723bb8950318135ed9cf4fc76bed988a087f536/BUILD#L70-L72

From what I gathered so far Emscripten does support pthreads, but not in standalone mode. If `--features=standalone_wasm` could also add wasi as the os constraint that would help selecting the right case there. The PR as is already helps, but looking for the way to make the experience the best out of the box for `wasm_cc_binary`.

I also noticed that in abseil, there's a use of a `STANDALONE_WASM` used as a compile directive:
```
#elif defined(__EMSCRIPTEN__) && !defined(STANDALONE_WASM)
#define ABSL_STACKTRACE_INL_HEADER \
  "absl/debugging/internal/stacktrace_emscripten-inl.inc"
```
https://github.com/abseil/abseil-cpp/blob/4e5ff1559ca3bd7bb777a1c48106464cb656e041/absl/debugging/internal/stacktrace_config.h#L41

I could not find anything about this being common in the Emscripten docs so was hesitant to include that in the PR. If it is, I could add it as a default compile opt when this flag is on?

bazel/emscripten_toolchain/toolchain.bzl
```diff
@@ -1015,6 +1018,11 @@ def _impl(ctx):
             flags = ["-Werror"],
             features = ["wasm_warnings_as_errors"],
         ),
+        flag_set(
+            actions = all_link_actions,
```
walkingeyerobot, line 1022: do we need to define anything during compile? I want to say we need to add something like this:
```bazel
        flag_set(
            actions = ALL_COMPILE_ACTIONS,
            flags = ["-DSTANDALONE_WASM=1"],
            features = ["wasm_standalone"],
        ),
```

maybe @sbc100 knows?

sbc100: I don't think any C macro should be needed no.

walkingeyerobot (approved):

bazel/emscripten_toolchain/toolchain.bzl
robbertvanginkel: Note that I did find that in the wild (https://github.com/emscripten-core/emsdk/pull/1145#issuecomment-1336212208), but could find no reference to this being standard.

sbc100: Its not a standard macro though.  We don't use that anywhere in emscripten.  

We have one place where we do compile something differently in standalone mode in in that case we just use `-DEMSCRIPTEN_STANDALONE_WASM` explicitly: https://github.com/emscripten-core/emscripten/blob/7c9b97ade051d1b0b871198d9067167dbec23152/tools/system_libs.py#L1862

sbc100:
I think the problem is that somehow all the test phases show an different names now.  All the test did actually pass, you can see them if you scroll down.  You can just land this as (with force).

PR #1146, opened 2022-12-03
Draft: add support for weak dependencies

kleisauke:
Allowing users to easily install, activate and use the Emscripten toolchain in combination with a system-installed version of Node.js.

This can be useful since emsdk currently provides Node.js 14.x, whose active support ended a year ago.

Current (and unchanged) behavior:
```console
$ ./emsdk install latest
$ ./emsdk activate latest
$ source ./emsdk_env.sh

$ which node
/home/kleisauke/emsdk/node/14.18.2_64bit/bin/node
$ node -v
v14.18.2
```

Behavior with the optional `--no-install-recommends` flag:
```console
$ ./emsdk install latest --no-install-recommends
$ ./emsdk activate latest
$ source ./emsdk_env.sh

$ which node
/usr/bin/node
$ node -v
v18.12.1
```

This PR also allows uninstalling the bundled Node.js version before the `activate` phase (this would previously fail with the `tool is not installed and therefore cannot be activated` error).
```console
$ ./emsdk install latest
$ ./emsdk uninstall node-14.18.2-64bit
$ ./emsdk activate latest
$ source ./emsdk_env.sh

$ which node
/usr/bin/node
$ node -v
v18.12.1
```

Marked this PR as draft due to these TODO-items:
- [ ] Rename this flag? It's basically the same flag as used by the APT package manager (I can't think of a better name for it).
- [ ] Split the commits in multiple PRs?
- [ ] Add documentation and tests for this flag.

(since this is considered a draft, I deliberately did not link any issues to this PR)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -23,7 +23,7 @@ COPY . ${EMSDK}
 
 RUN echo "## Install Emscripten" \
     && cd ${EMSDK} \
-    && ./emsdk install ${EMSCRIPTEN_VERSION} \
+    && ./emsdk install ${EMSCRIPTEN_VERSION} --no-install-recommends \
     && echo "## Done"
 
 # This generates configuration that contains all valid paths according to installed SDK
@@ -42,8 +42,6 @@ RUN cd ${EMSDK} \
 # Cleanup Emscripten installation and strip some symbols
 RUN echo "## Aggressive optimization: Remove debug symbols" \
     && cd ${EMSDK} && . ./emsdk_env.sh \
-    # Remove debugging symbols from embedded node (extra 7MB)
-    && strip -s `which node` \
     # Tests consume ~80MB disc space
     && rm -fr ${EMSDK}/upstream/emscripten/tests \
     # Fastcomp is not supported
@@ -64,8 +62,7 @@ COPY --from=stage_build /emsdk /emsdk
 # This will let use tools offered by this image inside other Docker images
 # (sub-stages) or with custom / no entrypoint
 ENV EMSDK=/emsdk \
-    EMSDK_NODE=/emsdk/node/14.18.2_64bit/bin/node \
-    PATH="/emsdk:/emsdk/upstream/emscripten:/emsdk/upstream/bin:/emsdk/node/14.18.2_64bit/bin:${PATH}"
+    PATH="/emsdk:/emsdk/upstream/emscripten:/emsdk/upstream/bin:${PATH}"
 
 # ------------------------------------------------------------------------------
 # Create a 'standard` 1000:1000 user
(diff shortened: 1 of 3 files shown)


sbc100:
Thanks for working on this.

On first glance, this seems like quite a lot of complexity to add simply to solve the issue of "how to avoid adding emsdk node".

In theory, if you don't want any of the sdk dependencies I think it should be possible to pick and choose which you want to install.   We already have the release binaries which are separate from their dependencies.  The `sdk-releases-upstream-%releases-tag%"` sdk pulls in `"releases-upstream-%releases-tag%-64bit"` altough with its dependencies.  I feel like if you just want to install  `"releases-upstream-%releases-tag%-64bit"` then we should just find an easy way to do it.

For example today you can do this:

```
$ ./emsdk install releases-upstream-48ce0b44015d0182fc8c27aa9fbc0a4474b55982-64bit
$ ./emsdk activate releases-upstream-48ce0b44015d0182fc8c27aa9fbc0a4474b55982-64bit
```

That will install the release binaries without any of the dependencies.  You can then pick and choose any other packages you want to install. 

One problem is that we don't have nice aliases like `latest` for these raw packages.  Perhaps adding the aliases would be enough..

kleisauke:
I think the complexity is reasonable. Commit 6a8d213273e3d21eae30d3aeea9a83054fe7c5f3 is the one that adds support for this, which are 65 additions and 60 deletions, not counting the whitespace changes. I could reduce the diff even more, since I added some comments and removed `recursive_dependencies()` in favor of `resolve_tools(recursive=True)` and `can_simultaneously_activate()` in favor of `is_remnant_or_conflicting_with()`. But I'm not sure if this is worth it, given that I'm modifying a 2733(!) SLOC file (my feeling would be to refactor this first if there are concerns about complexity).

You're right that this is only a issue for the SDK tags/aliases and that you can use `releases-upstream-%releases-tag%-64bit` for that. However, adding another `-base` alias for that (what is done in the subsequent PR) feels more like a workaround to me. Ideally, this should be a flag that is optional and that we can (hopefully) make the default behavior in the future.

Note that I considered two things before opening this draft PR:
1. Upgrade the bundled Node.js version to the latest LTS version;
2. Add a similar flag to `emsdk activate` and/or `emsdk_env` instead.

(1) could be easy in terms of implementation, but the discussion about dropping supporting Windows 7 has not been resolved (AFAIK). This could also lead to backward compatibility issues because we are not testing with the minimum supported Node.js version anymore. But in my opinion, that should not be a valid reason for not testing with newer versions. That said, we should implement the `-sMIN_NODEJS_VERSION` link-time option in Emscripten first to ensure the minimum supported Node.js version is defined and documented (this would also allow conditional exclusion of unnecessary polyfills based on the Node.js version).

By implementing (1), regressions would be noticed earlier, especially those behind the `--experimental-wasm-*` flag in V8. It could also reduce the flakiness in both the auto-roller and CI, or at least makes reproducing easier (sometimes I didn't investigate a flaky test further because it wasn't reproducible on newer Node.js versions).

(2) could be an alternative solution, but I think it would be neater to do this a stage earlier, as done in this draft PR. This way, users can still install and activate the bundled Node.js separately if we decide to enable this flag by default (thus preventing weak dependencies from being installed by default). This implementation is also a bit more future-proof; for example, it's easy to add a WASI-compliant runtime to the weak dependencies or to mark the Python dependency as weak on Windows and macOS (although we should probably remove that dependency once we are certain that Python 2.x is no longer used as a bootstrap).

kleisauke:
I'll abandon this PR, will revisit this when there's more interest.

PR #1151, opened 2022-12-17
Get node binary via @nodejs//:node

jfirebaugh:
I had a colleague using an arm64 mac whose builds were failing because `platform.machine()` was returning amd64 for some reason (my guess is the python toolchain was running under Rosetta). Then emscripten_config looked for the node binary in a nonexistent directory -- nodejs_darwin_amd64 rather than nodejs_darwin_arm64.

This switches to a public rules_nodejs API for obtaining the binary, so should always work.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -211,6 +211,7 @@ jobs:
     environment:
       PYTHONUNBUFFERED: "1"
       EMSDK_NOTTY: "1"
+      USE_BAZEL_VERSION: "5.4.0"
     steps:
       - checkout
       - run:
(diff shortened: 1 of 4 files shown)


walkingeyerobot (approved):

jfirebaugh:
I had to make two changes to get `test-bazel-windows` to pass:

* Use `@nodejs//:node_bin` on Windows rather than `@nodejs//:node`. I don't know why this is necessary but it mirrors the previous logic in emscripten_config. `@nodejs//:node_bin` is the `.../bin/nodejs/node.exe` path.
* Pin Bazel to 5.4.0, which seems to be the version that was tested prior to the 6.0.0 release today. This is a workaround for https://github.com/emscripten-core/emsdk/issues/1153.

jfirebaugh:
An equivalent change was merged as #1518.

PR #1155, opened 2022-12-21
emsdk.py: info(key + '=' + value) - no spaces

RandyMcMillan:
The info output now reflects the actual path:

example:
EMSDK=/Volumes/git/emsdk

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2737,7 +2737,7 @@ def construct_env_with_vars(env_vars_to_add):
       # Don't set env vars which are already set to the correct value.
       if key in os.environ and to_unix_path(os.environ[key]) == to_unix_path(value):
         continue
-      info(key + ' = ' + value)
+      info(key + '=' + value)
       if POWERSHELL:
         env_string += '$env:' + key + '="' + value + '"\n'
       elif CMD:


sbc100 (approved):

RandyMcMillan:
This works with linux and macos - and is very convenient (IMO).

A conditional for Windows and other shell environments needs to be added for similar functionality.

Any windows devs - please post a patch - attribution will be added.

sbc100:
> This works with linux and macos - and is very convenient (IMO).
> 
> A conditional for Windows and other shell environments needs to be added for similar functionality.
> 
> Any windows devs - please post a patch - attribution will be added.

With window syntax is `set FOO=BAR`.

And actually with bash/sh wouldn't one have to prefix each line with `export ` or one wanted it to be copy-pastable?

sbc100:
> > This works with linux and macos - and is very convenient (IMO).
> > A conditional for Windows and other shell environments needs to be added for similar functionality.
> > Any windows devs - please post a patch - attribution will be added.
> 
> With window syntax is `set FOO=BAR`.
> 
> And actually with bash/sh wouldn't one have to prefix each line with `export ` or one wanted it to be copy-pastable?

Given that the last part of that `set FOO=bar` and `export FOO=bar` lines is the same, maybe we can land this change as is, and we can decide if want to do something more complex as a followup?

PR #1165, opened 2023-01-05
Remove fastcomp SDK and fastcomp build rules

sbc100:
Folks that want to work with fastcomp will now need to use an older checkout of emsdk.


sbc100:
tested with `./emsdk install sdk-upstream-main-64bit`

kripken (approved):

PR #1166, opened 2023-01-10
Remove `-upstream` component of SDK names

sbc100:
This name existed to distinguish the SDK from fastcomp, but as of #1165, we no longer support fastcomp.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2111,15 +2111,15 @@ def resolve_sdk_aliases(name, verbose=False):
   return name
 
 
-def find_latest_sdk(which):
-  return 'sdk-releases-%s-%s-64bit' % (which, find_latest_hash())
+def find_latest_sdk():
+  return 'sdk-releases-%s-64bit' % (find_latest_hash())
 
 
 def find_tot_sdk():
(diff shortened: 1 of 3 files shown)


kripken:
Is this a breaking change? If so, can we strip `-upstream` from user inputs?

sbc100:
> Is this a breaking change? If so, can we strip `-upstream` from user inputs?

Yes, the `-upstream` is still supported in the user inputs but its basically ignores (we strip it off on 2706).

sbc100:
> > Is this a breaking change? If so, can we strip `-upstream` from user inputs?
> 
> Yes, the `-upstream` is still supported in the user inputs but its basically ignores (we strip it off on 2706).

Because the old names are still supported I think this is a backwards compatible change.

kripken:
Sounds good. Not sure why the tests fail then, you are right that that line should be stripping the suffix.

sbc100:
ping.. ok to land once tests pass?

kripken:
Lgtm, but please add a test (or keep one) that checks we strip `-upstream` properly, if you haven't already.

PR #1171, opened 2023-01-14
Rename `upstream` directory to simply `prebuilt`

sbc100:
The name `upstream` doesn't have any meaning anymore now that there is no more fastcomp.  It was removed from the SDK names in #1166 and this is the last usage of that name/term.


kripken (commented):
This sounds good, but I think it will surprise some users that have hardcoded paths using `upstream/`. We never recommended that but they might have just found where things are, and depended on it.

The best I can think of is to add a changelog to this repo, either in the readme or a separate file, and mention this there. We should also mention it in the emscripten changelog (but just mentioning it there doesn't seem like enough, since this will affect users using our standard install mechanism, which uses newest emsdk of the latest released emscripten; one option might be to wait to land this until right before an emscripten tag, but that has its own risks...)

sbc100:
> This sounds good, but I think it will surprise some users that have hardcoded paths using `upstream/`. We never recommended that but they might have just found where things are, and depended on it.
> 
> The best I can think of is to add a changelog to this repo, either in the readme or a separate file, and mention this there. We should also mention it in the emscripten changelog (but just mentioning it there doesn't seem like enough, since this will affect users using our standard install mechanism, which uses newest emsdk of the latest released emscripten; one option might be to wait to land this until right before an emscripten tag, but that has its own risks...)

Yeah, I was thinking the same thing..  I will hold off on this change for a bit as I know some folks were hardcoding paths to `upstream`.   Ideally nobody would be doing that.. so maybe I'll start by sending out a message to the list to see if we can get folks to stop doing it.

DreamOfIce:
> This sounds good, but I think it will surprise some users that have hardcoded paths using `upstream/`. We never recommended that but they might have just found where things are, and depended on it.
> 
> The best I can think of is to add a changelog to this repo, either in the readme or a separate file, and mention this there. We should also mention it in the emscripten changelog (but just mentioning it there doesn't seem like enough, since this will affect users using our standard install mechanism, which uses newest emsdk of the latest released emscripten; one option might be to wait to land this until right before an emscripten tag, but that has its own risks...)

I think maybe we can create a hard link or a symlink from `upstream/` to the new folder. This will encourage downstream updates to new folder names while retaining as much compatibility as possible

PR #1177, opened 2023-01-27
Add support for Visual Studio 2022

juj:
This PR does the following:

1. Add support for building Emscripten with Visual Studio 2022
2. Remove some code duplication in different places around VS version selection
3. Migrate to using `cmake --build` also when building on Windows. Historically we didn't use `cmake --build` because it was unable to find where MSBuild.exe resided, but it looks like those problems are long gone.
4. To enable `cmake --build` on Windows, leverage the MSBuild 'Multi-ToolTask' feature CL_MPCount to saturate project builds properly to 100% CPU usage so building LLVM builds different cpp files in parallel.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -284,32 +284,34 @@ if WINDOWS:
   # Detect which CMake generator to use when building on Windows
   if '--mingw' in sys.argv:
     CMAKE_GENERATOR = 'MinGW Makefiles'
-  elif '--vs2017' in sys.argv:
-    CMAKE_GENERATOR = 'Visual Studio 15'
+  elif '--vs2022' in sys.argv:
+    CMAKE_GENERATOR = 'Visual Studio 17'
   elif '--vs2019' in sys.argv:
     CMAKE_GENERATOR = 'Visual Studio 16'
+  elif '--vs2017' in sys.argv:
+    CMAKE_GENERATOR = 'Visual Studio 15'
+  elif len(vswhere(17)) > 0:
+    CMAKE_GENERATOR = 'Visual Studio 17'
+  elif len(vswhere(16)) > 0:
+    CMAKE_GENERATOR = 'Visual Studio 16'
+  elif len(vswhere(15)) > 0:
+    # VS2017 has an LLVM build issue, see
+    # https://github.com/kripken/emscripten-fastcomp/issues/185
+    CMAKE_GENERATOR = 'Visual Studio 15'
+  elif which('mingw32-make') is not None and which('g++') is not None:
+    CMAKE_GENERATOR = 'MinGW Makefiles'
   else:
-    vs2019_exists = len(vswhere(16)) > 0
-    vs2017_exists = len(vswhere(15)) > 0
-    mingw_exists = which('mingw32-make') is not None and which('g++') is not None
-    if vs2019_exists:
-      CMAKE_GENERATOR = 'Visual Studio 16'
-    elif vs2017_exists:
-      # VS2017 has an LLVM build issue, see
-      # https://github.com/kripken/emscripten-fastcomp/issues/185
-      CMAKE_GENERATOR = 'Visual Studio 15'
-    elif mingw_exists:
-      CMAKE_GENERATOR = 'MinGW Makefiles'
-    else:
-      # No detected generator
-      CMAKE_GENERATOR = ''
+    # No detected generator
+    CMAKE_GENERATOR = ''
 
 
-sys.argv = [a for a in sys.argv if a not in ('--mingw', '--vs2017', '--vs2019')]
+sys.argv = [a for a in sys.argv if a not in ('--mingw', '--vs2017', '--vs2019', '--vs2022')]
 
 
 # Computes a suitable path prefix to use when building with a given generator.
 def cmake_generator_prefix():
+  if CMAKE_GENERATOR == 'Visual Studio 17':
+    return '_vs2022'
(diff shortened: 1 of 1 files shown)


sbc100 (commented):
Nice!

In the long run, is there any reason not to simply require ninja for performing these builds?

sbc100 (approved):

juj:
> In the long run, is there any reason not to simply require ninja for performing these builds?

If we embed Ninja precompiled into the repository, then maybe not. Though assuming upstream CMake fixes the CL_MPCount flags out of the box one day, doing `cmake` followed by `cmake --build .` should be clean enough.

sbc100:
Looks like the version of cmake we have on the bots doesn't support `--build` + `-j`:

```
Running build: ['cmake', '--build', '.', '--config', 'Release', '-j', '4']
Unknown argument -j
Unknown argument 4
```

sbc100:
I think maybe you need `--` between the cmake flags and the build tool flags (like `-j`)

juj:
Huh, Linux test fails on the `cmake --build -j4` command line. It states

```
Running build: ['cmake', '--build', '.', '--config', 'Release', '-j', '4']
Unknown argument -j
Unknown argument 4
Usage: cmake --build <dir> [options] [-- [native-options]]
Options:
  <dir>          = Project binary directory to be built.
  --target <tgt> = Build <tgt> instead of default targets.
                   May only be specified once.
  --config <cfg> = For multi-configuration tools, choose <cfg>.
  --clean-first  = Build target 'clean' first, then build.
                   (To clean only, use --target 'clean'.)
  --use-stderr   = Ignored.  Behavior is default in CMake >= 3.0.
  --             = Pass remaining options to the native tool.
Build failed due to exception!
Working directory: /root/project/binaryen/main_64bit_binaryen
Command '['cmake', '--build', '.', '--config', 'Release', '-j', '4']' returned non-zero exit status 1.
error: installation failed!
```

When I run `cmake --build` locally, I get
```
C:\emsdk>cmake --build
Usage: cmake --build <dir>             [options] [-- [native-options]]
       cmake --build --preset <preset> [options] [-- [native-options]]
Options:
  <dir>          = Project binary directory to be built.
  --preset <preset>, --preset=<preset>
                 = Specify a build preset.
  --list-presets[=<type>]
                 = List available build presets.
  --parallel [<jobs>], -j [<jobs>]
                 = Build in parallel using the given number of jobs.
                   If <jobs> is omitted the native build tool's
                   default number is used.
                   The CMAKE_BUILD_PARALLEL_LEVEL environment variable
                   specifies a default parallel level when this option
                   is not given.
  -t <tgt>..., --target <tgt>...
                 = Build <tgt> instead of default targets.
  --config <cfg> = For multi-configuration tools, choose <cfg>.
  --clean-first  = Build target 'clean' first, then build.
                   (To clean only, use --target 'clean'.)
  --resolve-package-references={on|only|off}
                 = Restore/resolve package references during build.
  -v, --verbose  = Enable verbose output - if supported - including
                   the build commands to be executed.
  --             = Pass remaining options to the native tool.
```

According to https://github.com/Kitware/CMake/blob/master/Help/manual/cmake.1.rst `-j` flag would have been added in CMake 3.12, which was released on 18 Jul 2018. Building LLVM requires [CMake 3.13.4 at minimum](https://github.com/llvm/llvm-project/blob/bcc10817d5569172ee065015747e226280e9b698/llvm/CMakeLists.txt#L3).. I can't quite understand how it is possible that the Linux cmake would not support the `cmake --build . -j4` option.

I'll work around by passing the build specific `cmake --build . -- -j4` option, but the above doesn't make much sense to me..

emsdk.py
```diff
@@ -916,112 +911,34 @@ def build_env(generator):
...
-  else:
-    make = ['cmake', '--build', '.', '--', '-j' + str(CPU_CORES)]
+  make = ['cmake', '--build', '.', '--config', build_type, '-j', str(CPU_CORES)]
+  if 'Visual Studio' in CMAKE_GENERATOR:
+    # Visual Studio historically has had a two-tier problem in its build system design. A single MSBuild.exe instance only governs
+    # the build of a single project (.exe/.lib/.dll) in a solution. Passing the -j parameter above will only enable multiple MSBuild.exe
+    # instances to be spawned to build multiple projects in parallel, but each MSBuild.exe is still singlethreaded.
+    # To enable each MSBuild.exe instance to also compile several .cpp files in parallel inside a single project, pass the extra
+    # MSBuild.exe specific "Multi-ToolTask" (MTT) setting /p:CL_MPCount. This enables each MSBuild.exe to parallelize builds wide.
+    make += ['--', '/p:CL_MPCount=' + str(CPU_CORES)]
```
sbc100, line 936: Maybe add `['--', '-j', str(CPU_CORES)]` here in the else block?

sbc100:
Looks like the CI bots run ubuntu/bionic which has cmake 3.10.2.

```
Get:25 http://archive.ubuntu.com/ubuntu bionic-updates/main amd64 cmake amd64 3.10.2-1ubuntu2.18.04.2 [3152 kB]
```

We don't test building llvm from source in CI, only building binaryen, which I guess doesn't require such a new version of cmake.   

Leaving the `--` for now seems reasonable.

juj:
Ohhh, that explains..

juj:
All the tests passed, but somehow the checks are all doubled, with a second copy waiting for status to be reported, and the first copy succeeding?

I'll force-land this, looks like it's all green, but something might be sideways with the CI?

PR #1179, opened 2023-01-28
Fix args ordering in LLVM build from previous commit

juj:
Looks like the CI did not fail on an error from the previous PR, but just left the tasks hung. I misinterpreted the two copies of each task (one green, another yellow hung) as a CI fluke, but it might be caused by this typo in refactor.


juj:
Hmm, no that was not it.. the CI is still hanging.

Nevertheless, force-landing this to quickly unblock the breakage.

sbc100 (commented):
I think the CI issues is something to do with circleci job names vs github expectation names.  We probably need to update the settings to match the current names

Issue #1173, opened 2023-01-20
[feature] Upgrade to Node.js 20(LTS)


DreamOfIce:
As written in https://github.com/nodejs/release#release-schedule, support for node.js 16 was end in September 2023.
Therefore, it is recommended to upgrade to node.js 20.
Also, many of the features in node.js 20 are enabled by default, such as `--experimental-wasm-threads` and `--experimental-wasm-bulk-memory`.

sbc100:
Moved to `emsdk` since its the emsdk version of node we are talking about upgrading (emscripten itself already supports newer versions).

DreamOfIce:
> Moved to `emsdk` since its the emsdk version of node we are talking about upgrading (emscripten itself already supports newer versions).

Are there any impediments to updating ci's version of Ubuntu? After all, the five-year free support period for Ubuntu 18 is almost up

sbc100:
Yes, there were impediments.  We don't have direct control over the VM.   But I think we have now overcome those.

BTW, according to https://wiki.ubuntu.com/Releases ubunut/bionic 18.04 LTS is supported until April 2028.

The issue of what version of node we use internally in emsdk should really be an implementation detail though.    I think the real issue is that we are putting our internal version of node in the emsdk user's path, and we should focus on fixing that.  Even if/when we update to node v18 that will also be wrong version for many users who have already installed their own.

DreamOfIce:
I agree with this, forcing the specified version of node into path is a very bad idea. We should at least provide an option to use the system node.

Considering that the lifecycle of node.js14 is coming to an end, updating the bundled node version is unavoidable, even if only for security reasons

BTW, the free LTS period for Ubuntu 18 will end in April and paid commercial support for the next five years

sbc100:
> Considering that the lifecycle of node.js14 is coming to an end, updating the bundled node version is unavoidable, even if only for security reasons

I think unavoidable might be too strong a word there.  Perhaps inadvisable.

sbc100:
> Considering that the lifecycle of node.js14 is coming to an end, updating the bundled node version is unavoidable, even if only for security reasons
> 

Emscripten's use of node is as part of the compiler toolchain and as such it does not process untrusted code.   There is no protection at all against hostile user input in emscripten, regardless of node version.  For example `--library-js=foo.js` executes `foo.js` with full user permissions at build time.

DreamOfIce:
[...]

PR #1183, opened 2023-02-07
Upgrade CI to Ubuntu 20.04 LTS (Focal)

DreamOfIce:
# Why
On one hand Ubuntu18's free LTS cycle is ended, and on the other hand Ubuntu18 does not support nodejs greater than v18.
# What's changed.
1. update the Linux Docker image to buildpack-deps:focal
2. some compatibility changes
3. use the new recommended way to install Docker ([see here](https://docs.docker.com/engine/install/ubuntu/))

See https://github.com/emscripten-core/emsdk/issues/1173


sbc100:
I could of comments:

Firstly, we should gets some consensus that we are ok with dropping support for bionic.  @juj is often concerned about supported older OSes but I think mostly around macOS and windows. 

Secondly, I suggest we update to 20.04 focal which will effect fewer users.   This also happens to be the update that we recently did on the emscripten-releases bots which actually build the emsdk binaries, so I think the makes sense to do that same update there.  If you are ok with that could you update this PR to use focal instead?

sbc100:
Looks like we do document the system requirements here: https://github.com/emscripten-core/emsdk#linux , so those should be updated too.

.circleci/config.yml
```diff
@@ -39,7 +42,7 @@ jobs:
       - checkout
       - run:
           name: Install debian packages
-          command: apt-get update -q && apt-get install -q -y cmake build-essential openjdk-8-jre-headless ksh zsh
+          command: apt-get update -q && apt-get install -q -y cmake build-essential openjdk-8-jre-headless ksh zsh python-is-python3
```
sbc100, line 45: Why is `python-is-python3` needed here?

.circleci/config.yml
```diff
@@ -134,12 +137,19 @@ jobs:
...
           name: install docker
-          command: apt-get update -q && apt-get install -q -y docker.io
+          command: |
+            apt-get update -q 
+            apt-get install -q -y ca-certificates curl gnupg lsb-release
+            mkdir -p /etc/apt/keyrings
+            curl -fsSL https://download.docker.com/linux/ubuntu/gpg | gpg --dearmor -o /etc/apt/keyrings/docker.gpg
+            echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" | tee /etc/apt/sources.list.d/docker.list > /dev/null
+            apt-get update -q
+            apt-get install -q -y docker-ce docker-ce-cli containerd.io docker-compose-plugin
```
sbc100, line 152: Why is all this extra stuff needed here?

DreamOfIce: Docker has updated the recommended installation method, see https://docs.docker.com/engine/install/ubuntu/#uninstall-old-versions
I think since they have updated the system version, these can be updated together.

.circleci/config.yml
DreamOfIce: The purpose of this package is to point python to python3. If not installed, test-linux will complain that python cannot be found:  [log](https://app.circleci.com/pipelines/github/DreamOfIce/emsdk/4/workflows/5e1dbeda-c74e-41ff-9a21-7fea461e7859/jobs/27)

DreamOfIce:
> I could of comments:
> 
> Firstly, we should gets some consensus that we are ok with dropping support for bionic. @juj is often concerned about supported older OSes but I think mostly around macOS and windows.
> 
> Secondly, I suggest we update to 20.04 focal which will effect fewer users. This also happens to be the update that we recently did on the emscripten-releases bots which actually build the emsdk binaries, so I think the makes sense to do that same update there. If you are ok with that could you update this PR to use focal instead?

Sounds good, I'll give it a try 😄

.circleci/config.yml
sbc100: I think it might make more sense to change the #! line in test.py instead.

.circleci/config.yml
sbc100: Is that needed as part of this change to a new ubuntu version?  If not, lets make that a separate change maybe?

DreamOfIce: No, it is not necessary

.circleci/config.yml
DreamOfIce: > I think it might make more sense to change the #! line in test.py instead.

No, this is a problem with https://github.com/DreamOfIce/emsdk/blob/main/test/test.py#L223
Here the command python is called directly

DreamOfIce:
All CI tests under focal passed

.circleci/config.yml
sbc100: I think you can just drop the `python` from that string (or replace it with `sys.executable`)

DreamOfIce: Completed!

.circleci/config.yml
DreamOfIce: > Is that needed as part of this change to a new ubuntu version? If not, lets make that a separate change maybe?

However, as of now, I don't see the new versions 23.0.0 and 21.10.22/23 in the docker.io package, so I think it's best to use the new installation method

juj:
> Firstly, we should gets some consensus that we are ok with dropping support for bionic. @juj is often concerned about supported older OSes but I think mostly around macOS and windows.

We do ship Emsdk on Ubuntu 18 and some other Linuxes, though we do build our own Emsdk + Emscripten bundles manually, and don't rely on the automated CI here.

So I am wholly ok with the automated emsdk CI here being upgraded to newer Linux targets, as we will still test to know that our older target is supported. (it does not look like there is any needed change to actively break support for Emsdk/Emscripten for older Linuxes)

.circleci/config.yml
sbc100: Can you remove this part of the change and submit a separate PR?     Seems like this is switching from the ubuntu packaged docker to directly using the upstream pacakges.  We can discuss that separately there maybe?

sbc100:
> So I am wholly ok with the automated emsdk CI here being upgraded to newer Linux targets, as we will still test to know that our older target is supported. (it does not look like there is any needed change to actively break support for Emsdk/Emscripten for older Linuxes)

The problem is that if we stop testing on older version of linux then dependencies on newer glibc could creep into our binaries without us noticing.    If we decide that the binaries are only tested on focal then the chances of them continuing to work on older linuxs over time will be low. 

Since you are building your own binaries when you target Ubuntu 18 at least we don't need to worry about that use case, but we should reach on the mailing list to see if others are depending on these binaries running on Ubuntu 18.

We actually currently build the binaries against the debian/stretch sysroot:  https://chromium.googlesource.com/emscripten-releases/+/refs/heads/main/src/build.py#65.

So the current situation is a little more confusing that I had thought:

Binaries built against: Debian/Stretch (glibc 2.24)
Binaries built on: Ubuntu/Focal 20.04 (glibc 2.31)
Current document requirement: Ubuntu/Xenial 16.04 (glibc 2.23)
Oldest linux version by our users: ???

Actual glibc dependencies in current binaries:

```
$ readelf -V upstream/bin/clang 
...
Version needs section '.gnu.version_r' contains 8 entries:
 Addr: 0x00000000001111d0  Offset: 0x001111d0  Link: 7 (.dynstr)
...
  0x0180:   Name: GLIBC_2.16  Flags: none  Version: 17
```

GLIBC_2.16 seems to be the current highest value.

.circleci/config.yml
DreamOfIce: Okay, I'll revert this change first, and open a separate one later

DreamOfIce: After switching back to the old way there was a problem with daemon not starting automatically. I tried ssh connection and then manually started docker with the following error:
```shell
root@9cdcd1906338: ~/projectroot@9cdcd1906338:~/project# /usr/bin/dockerd
INFO[2023-02-08T17:16:59.597232157Z] Starting up                                  
INFO[2023-02-08T17:16:59.598653570Z] libcontainerd: started new containerd process  pid=1044
INFO[2023-02-08T17:16:59.598990184Z] parsed scheme: "unix"                         module=grpc
INFO[2023-02-08T17:16:59.599003352Z] scheme "unix" not registered, fallback to default scheme  module=grpc
INFO[2023-02-08T17:16:59.599023135Z] ccResolverWrapper: sending update to cc: {[{unix:///var/run/docker/containerd/containerd.sock  <nil> 0 <nil>}] <nil> <nil>}  module=grpc
INFO[2023-02-08T17:16:59.599036101Z] ClientConn switching balancer to "pick_first"  module=grpc
WARN[0000] deprecated version : `1`, please switch to version `2` 
INFO[2023-02-08T17:16:59.637728889Z] starting containerd                           revision= version="1.5.9-0ubuntu1~20.04.6"
INFO[2023-02-08T17:16:59.657022988Z] loading plugin "io.containerd.content.v1.content"...  type=io.containerd.content.v1
INFO[2023-02-08T17:16:59.657089145Z] loading plugin "io.containerd.snapshotter.v1.aufs"...  type=io.containerd.snapshotter.v1
INFO[2023-02-08T17:16:59.657191573Z] loading plugin "io.containerd.snapshotter.v1.btrfs"...  type=io.containerd.snapshotter.v1
INFO[2023-02-08T17:16:59.657408969Z] skip loading plugin "io.containerd.snapshotter.v1.btrfs"...  error="path /var/lib/docker/containerd/daemon/io.containerd.snapshotter.v1.btrfs (overlay) must be a btrfs filesystem to be used with the btrfs snapshotter: skip plugin" type=io.containerd.snapshotter.v1
INFO[2023-02-08T17:16:59.657430926Z] loading plugin "io.containerd.snapshotter.v1.devmapper"...  type=io.containerd.snapshotter.v1
WARN[2023-02-08T17:16:59.657451965Z] failed to load plugin io.containerd.snapshotter.v1.devmapper  error="devmapper not configured"
INFO[2023-02-08T17:16:59.657463826Z] loading plugin "io.containerd.snapshotter.v1.native"...  type=io.containerd.snapshotter.v1
INFO[2023-02-08T17:16:59.657486806Z] loading plugin "io.containerd.snapshotter.v1.overlayfs"...  type=io.containerd.snapshotter.v1
INFO[2023-02-08T17:16:59.658049391Z] loading plugin "io.containerd.snapshotter.v1.zfs"...  type=io.containerd.snapshotter.v1
INFO[2023-02-08T17:16:59.658252751Z] skip loading plugin "io.containerd.snapshotter.v1.zfs"...  error="path /var/lib/docker/containerd/daemon/io.containerd.snapshotter.v1.zfs must be a zfs filesystem to be used with the zfs snapshotter: skip plugin" type=io.containerd.snapshotter.v1
INFO[2023-02-08T17:16:59.658272166Z] loading plugin "io.containerd.metadata.v1.bolt"...  type=io.containerd.metadata.v1
WARN[2023-02-08T17:16:59.658291790Z] could not use snapshotter devmapper in metadata plugin  error="devmapper not configured"
INFO[2023-02-08T17:16:59.658302759Z] metadata content store policy set             policy=shared
INFO[2023-02-08T17:16:59.658510986Z] loading plugin "io.containerd.differ.v1.walking"...  type=io.containerd.differ.v1
INFO[2023-02-08T17:16:59.658529707Z] loading plugin "io.containerd.gc.v1.scheduler"...  type=io.containerd.gc.v1
INFO[2023-02-08T17:16:59.658562722Z] loading plugin "io.containerd.service.v1.introspection-service"...  type=io.containerd.service.v1
INFO[2023-02-08T17:16:59.658590386Z] loading plugin "io.containerd.service.v1.containers-service"...  type=io.containerd.service.v1
INFO[2023-02-08T17:16:59.658604298Z] loading plugin "io.containerd.service.v1.content-service"...  type=io.containerd.service.v1
INFO[2023-02-08T17:16:59.658617846Z] loading plugin "io.containerd.service.v1.diff-service"...  type=io.containerd.service.v1
INFO[2023-02-08T17:16:59.658632425Z] loading plugin "io.containerd.service.v1.images-service"...  type=io.containerd.service.v1
INFO[2023-02-08T17:16:59.658646337Z] loading plugin "io.containerd.service.v1.leases-service"...  type=io.containerd.service.v1
INFO[2023-02-08T17:16:59.658660716Z] loading plugin "io.containerd.service.v1.namespaces-service"...  type=io.containerd.service.v1
INFO[2023-02-08T17:16:59.658674809Z] loading plugin "io.containerd.service.v1.snapshots-service"...  type=io.containerd.service.v1
INFO[2023-02-08T17:16:59.658688461Z] loading plugin "io.containerd.runtime.v1.linux"...  type=io.containerd.runtime.v1
INFO[2023-02-08T17:16:59.658988183Z] loading plugin "io.containerd.runtime.v2.task"...  type=io.containerd.runtime.v2
INFO[2023-02-08T17:16:59.659048161Z] loading plugin "io.containerd.monitor.v1.cgroups"...  type=io.containerd.monitor.v1
INFO[2023-02-08T17:16:59.659892962Z] loading plugin "io.containerd.service.v1.tasks-service"...  type=io.containerd.service.v1
INFO[2023-02-08T17:16:59.659937347Z] loading plugin "io.containerd.internal.v1.restart"...  type=io.containerd.internal.v1
INFO[2023-02-08T17:16:59.659985473Z] loading plugin "io.containerd.grpc.v1.containers"...  type=io.containerd.grpc.v1
INFO[2023-02-08T17:16:59.660140409Z] loading plugin "io.containerd.grpc.v1.content"...  type=io.containerd.grpc.v1
INFO[2023-02-08T17:16:59.660160719Z] loading plugin "io.containerd.grpc.v1.diff"...  type=io.containerd.grpc.v1
INFO[2023-02-08T17:16:59.660177443Z] loading plugin "io.containerd.grpc.v1.events"...  type=io.containerd.grpc.v1
INFO[2023-02-08T17:16:59.660190131Z] loading plugin "io.containerd.grpc.v1.healthcheck"...  type=io.containerd.grpc.v1
INFO[2023-02-08T17:16:59.660207561Z] loading plugin "io.containerd.grpc.v1.images"...  type=io.containerd.grpc.v1
INFO[2023-02-08T17:16:59.660221505Z] loading plugin "io.containerd.grpc.v1.leases"...  type=io.containerd.grpc.v1
INFO[2023-02-08T17:16:59.660233711Z] loading plugin "io.containerd.grpc.v1.namespaces"...  type=io.containerd.grpc.v1
INFO[2023-02-08T17:16:59.660246460Z] loading plugin "io.containerd.internal.v1.opt"...  type=io.containerd.internal.v1
INFO[2023-02-08T17:16:59.660283010Z] loading plugin "io.containerd.grpc.v1.snapshots"...  type=io.containerd.grpc.v1
INFO[2023-02-08T17:16:59.660298546Z] loading plugin "io.containerd.grpc.v1.tasks"...  type=io.containerd.grpc.v1
INFO[2023-02-08T17:16:59.660311877Z] loading plugin "io.containerd.grpc.v1.version"...  type=io.containerd.grpc.v1
INFO[2023-02-08T17:16:59.660325599Z] loading plugin "io.containerd.grpc.v1.introspection"...  type=io.containerd.grpc.v1
INFO[2023-02-08T17:16:59.660509024Z] serving...                                    address=/var/run/docker/containerd/containerd-debug.sock
INFO[2023-02-08T17:16:59.660580951Z] serving...                                    address=/var/run/docker/containerd/containerd.sock.ttrpc
INFO[2023-02-08T17:16:59.660639047Z] serving...                                    address=/var/run/docker/containerd/containerd.sock
INFO[2023-02-08T17:16:59.660657130Z] containerd successfully booted in 0.023790s  
INFO[2023-02-08T17:16:59.664656375Z] parsed scheme: "unix"                         module=grpc
INFO[2023-02-08T17:16:59.664689485Z] scheme "unix" not registered, fallback to default scheme  module=grpc
INFO[2023-02-08T17:16:59.664705798Z] ccResolverWrapper: sending update to cc: {[{unix:///var/run/docker/containerd/containerd.sock  <nil> 0 <nil>}] <nil> <nil>}  module=grpc
INFO[2023-02-08T17:16:59.664716391Z] ClientConn switching balancer to "pick_first"  module=grpc
INFO[2023-02-08T17:16:59.665349989Z] parsed scheme: "unix"                         module=grpc
INFO[2023-02-08T17:16:59.665371205Z] scheme "unix" not registered, fallback to default scheme  module=grpc
INFO[2023-02-08T17:16:59.665386969Z] ccResolverWrapper: sending update to cc: {[{unix:///var/run/docker/containerd/containerd.sock  <nil> 0 <nil>}] <nil> <nil>}  module=grpc
INFO[2023-02-08T17:16:59.665397481Z] ClientConn switching balancer to "pick_first"  module=grpc
ERRO[2023-02-08T17:16:59.666471002Z] failed to mount overlay: operation not permitted  storage-driver=overlay2
ERRO[2023-02-08T17:16:59.666528603Z] exec: "fuse-overlayfs": executable file not found in $PATH  storage-driver=fuse-overlayfs
ERRO[2023-02-08T17:16:59.666600428Z] AUFS cannot be used in non-init user namespace  storage-driver=aufs
ERRO[2023-02-08T17:16:59.666984325Z] failed to mount overlay: operation not permitted  storage-driver=overlay
WARN[2023-02-08T17:16:59.669131130Z] Your kernel does not support CPU realtime scheduler 
WARN[2023-02-08T17:16:59.669165030Z] Your kernel does not support cgroup blkio weight 
WARN[2023-02-08T17:16:59.669171719Z] Your kernel does not support cgroup blkio weight_device 
INFO[2023-02-08T17:16:59.669321203Z] Loading containers: start.                   
WARN[2023-02-08T17:16:59.670720394Z] Running iptables --wait -t nat -L -n failed with message: `iptables v1.8.4 (legacy): can't initialize iptables table `nat': Permission denied (you must be root)
Perhaps iptables or your kernel needs to be upgraded.`, error: exit status 3 
INFO[2023-02-08T17:16:59.687118997Z] stopping event stream following graceful shutdown  error="<nil>" module=libcontainerd namespace=moby
INFO[2023-02-08T17:16:59.687435414Z] stopping healthcheck following graceful shutdown  module=libcontainerd
INFO[2023-02-08T17:16:59.687450875Z] stopping event stream following graceful shutdown  error="context canceled" module=libcontainerd namespace=plugins.moby
WARN[2023-02-08T17:17:00.688281130Z] grpc: addrConn.createTransport failed to connect to {unix:///var/run/docker/containerd/containerd.sock  <nil> 0 <nil>}. Err :connection error: desc = "transport: Error while dialing dial unix:///var/run/docker/containerd/containerd.sock: timeout". Reconnecting...  module=grpc
failed to start daemon: Error initializing network controller: error obtaining controller instance: failed to create NAT chain DOCKER: iptables failed: iptables -t nat -N DOCKER: iptables v1.8.4 (legacy): can't initialize iptables table `nat': Permission denied (you must be root)
Perhaps iptables or your kernel needs to be upgraded.
```

DreamOfIce: So I think there is still Install it in a new way

DreamOfIce: Reverted to the new installation

DreamOfIce:
> > So I am wholly ok with the automated emsdk CI here being upgraded to newer Linux targets, as we will still test to know that our older target is supported. (it does not look like there is any needed change to actively break support for Emsdk/Emscripten for older Linuxes)
> 
> The problem is that if we stop testing on older version of linux then dependencies on newer glibc could creep into our binaries without us noticing. If we decide that the binaries are only tested on focal then the chances of them continuing to work on older linuxs over time will be low.
> 
> Since you are building your own binaries when you target Ubuntu 18 at least we don't need to worry about that use case, but we should reach on the mailing list to see if others are depending on these binaries running on Ubuntu 18.
> 
> We actually currently build the binaries against the debian/stretch sysroot: https://chromium.googlesource.com/emscripten-releases/+/refs/heads/main/src/build.py#65.
> 
> So the current situation is a little more confusing that I had thought:
> 
> Binaries built against: Debian/Stretch (glibc 2.24) Binaries built on: Ubuntu/Focal 20.04 (glibc 2.31) emsdk testing: Ubuntu/Bionic 18.03 (glibc 2.27) emscripten testing: Ubuntu/Bionic 18.03 (glibc 2.27) emscripten-releases testing: Ubuntu/Focal 20.04 (glibc 2.31) Current document requirement: Ubuntu/Xenial 16.04 (glibc 2.23) Oldest linux version by our users: ???
> 
> Actual glibc dependencies in current binaries:
> 
> ```
> $ readelf -V upstream/bin/clang 
> ...
> Version needs section '.gnu.version_r' contains 8 entries:
>  Addr: 0x00000000001111d0  Offset: 0x001111d0  Link: 7 (.dynstr)
> ...
>   0x0180:   Name: GLIBC_2.16  Flags: none  Version: 17
> ```
> 
> GLIBC_2.16 seems to be the current highest value.
> 
> (node v18 requires glibc 2.28: [nodesource/distributions#1392](https://github.com/nodesource/distributions/issues/1392))

I think it would be better to standardize all the distributions everywhere and change the corresponding minimum requirements in the documentation (e.g. >= Debian10/Ubuntu20.04 or similar)

sbc100:
> I think it would be better to standardize all the distributions everywhere and change the corresponding minimum requirements in the documentation (e.g. >= Debian10/Ubuntu20.04 or similar)

Yes, clearly it would be better to have a more consistent story.    The reason I'm holding of landing this change right away (don't get me wrong I think we can land it eventually) is that I'd like to get the story straight first, and I think we should coordinate with all the other places we do testing an building.

DreamOfIce:
> > I think it would be better to standardize all the distributions everywhere and change the corresponding minimum requirements in the documentation (e.g. >= Debian10/Ubuntu20.04 or similar)
> 
> Yes, clearly it would be better to have a more consistent story. The reason I'm holding of landing this change right away (don't get me wrong I think we can land it eventually) is that I'd like to get the story straight first, and I think we should coordinate with all the other places we do testing an building.

So currently emscripten is built on Google's machine?
https://chromium.googlesource.com/emscripten-releases

sbc100:
> > > I think it would be better to standardize all the distributions everywhere and change the corresponding minimum requirements in the documentation (e.g. >= Debian10/Ubuntu20.04 or similar)
> > 
> > 
> > Yes, clearly it would be better to have a more consistent story. The reason I'm holding of landing this change right away (don't get me wrong I think we can land it eventually) is that I'd like to get the story straight first, and I think we should coordinate with all the other places we do testing an building.
> 
> So currently emscripten is built on Google's machine? https://chromium.googlesource.com/emscripten-releases

Yes, its built on chromium infrastructure bots: https://ci.chromium.org/p/emscripten-releases/g/main/console

We depend on this for things like autorollers (https://autoroll.skia.org/?filter=emscripten) which constantly update the various dependencies and test the results before commited.

DreamOfIce:
> > I think it would be better to standardize all the distributions everywhere and change the corresponding minimum requirements in the documentation (e.g. >= Debian10/Ubuntu20.04 or similar)
> 
> Yes, clearly it would be better to have a more consistent story. The reason I'm holding of landing this change right away (don't get me wrong I think we can land it eventually) is that I'd like to get the story straight first, and I think we should coordinate with all the other places we do testing an building.

Hope this can come soon

sbc100:
> We do ship Emsdk on Ubuntu 18 and some other Linuxes

@juj, do you know how long you are planning to support Ubuntu 18?   I'm asking because there are no binaries for node v18 and above available that run on Ubuntu 18, so we can't look into updated to v18 until we completely drop support for Ubuntu 18.

DreamOfIce:
> > We do ship Emsdk on Ubuntu 18 and some other Linuxes
> 
> @juj, do you know how long you are planning to support Ubuntu 18? I'm asking because there are no binaries for node v18 and above available that run on Ubuntu 18, so we can't look into updated to v18 until we completely drop support for Ubuntu 18.

However it looks like node18 doesn't use the new features of glibc2.28, which requires the same glibc as the previous version.
There are two possible reasons for this situation.
1. node18 upgraded the build system to RHEL8 (since EOL for RHEL7 predates node18), and RHEL8 comes with glibc version 2.28 (refer to https://github.com/nodejs/node/pull/42659#issue-1197560117)
2. A certain commit of V8 raised the requirement for glibc to 2.27, leading the node team to think that v8 might raise the requirement for glibc, but in fact v8 added a patch shortly afterwards to support older versions: https://github.com/v8/v8/commit/4e81f2584ee15cd6300e180585c26f16616210f3

Since we have a good enough testing process, we can consider building node18 without glibc2.28 and distributing it to users

sbc100:
> > > We do ship Emsdk on Ubuntu 18 and some other Linuxes
> > 
> > 
> > @juj, do you know how long you are planning to support Ubuntu 18? I'm asking because there are no binaries for node v18 and above available that run on Ubuntu 18, so we can't look into updated to v18 until we completely drop support for Ubuntu 18.
> 
> However it looks like node18 doesn't use the new features of glibc2.28, which requires the same glibc as the previous version. There are two possible reasons for this situation.
> 
> 1. node18 upgraded the build system to RHEL8 (since EOL for RHEL7 predates node18), and RHEL8 comes with glibc version 2.28 (refer to [doc: update minimum glibc requirements for Linux nodejs/node#42659 (comment)](https://github.com/nodejs/node/pull/42659#issue-1197560117))
> 2. A certain commit of V8 raised the requirement for glibc to 2.27, leading the node team to think that v8 might raise the requirement for glibc, but in fact v8 added a patch shortly afterwards to support older versions: [v8/v8@4e81f25](https://github.com/v8/v8/commit/4e81f2584ee15cd6300e180585c26f16616210f3)
> 
> Since we have a good enough testing process, we can consider building node18 without glibc2.28 and distributing it to users

I'd really like to avoid having to build nodejs from scratch... based on the long discussions on the nodejs issues about this it seems like there are solutions to the problem, including building from scratch, or hacking the pre-built binaries, but that sounds like a lot of maintenance work compared to just downloading the nodejs like we do today.

My hope is that we can there will come a time when we no longer need to support ubuntu 18, and in the mean time we can work on internalizing out version of node so it becomes more of an implementation detail (i.e. its doesn't end up poeple PATH).    Keeping the old version of node for a while for internal use seems harmless enough.

DreamOfIce:
> > > > We do ship Emsdk on Ubuntu 18 and some other Linuxes
> > > 
> > > 
> > > @juj, do you know how long you are planning to support Ubuntu 18? I'm asking because there are no binaries for node v18 and above available that run on Ubuntu 18, so we can't look into updated to v18 until we completely drop support for Ubuntu 18.
> > 
> > 
> > However it looks like node18 doesn't use the new features of glibc2.28, which requires the same glibc as the previous version. There are two possible reasons for this situation.
> > 
> > 1. node18 upgraded the build system to RHEL8 (since EOL for RHEL7 predates node18), and RHEL8 comes with glibc version 2.28 (refer to [doc: update minimum glibc requirements for Linux nodejs/node#42659 (comment)](https://github.com/nodejs/node/pull/42659#issue-1197560117))
> > 2. A certain commit of V8 raised the requirement for glibc to 2.27, leading the node team to think that v8 might raise the requirement for glibc, but in fact v8 added a patch shortly afterwards to support older versions: [v8/v8@4e81f25](https://github.com/v8/v8/commit/4e81f2584ee15cd6300e180585c26f16616210f3)
> > 
> > Since we have a good enough testing process, we can consider building node18 without glibc2.28 and distributing it to users
> 
> I'd really like to avoid having to build nodejs from scratch... based on the long discussions on the nodejs issues about this it seems like there are solutions to the problem, including building from scratch, or hacking the pre-built binaries, but that sounds like a lot of maintenance work compared to just downloading the nodejs like we do today.
> 
> My hope is that we can there will come a time when we no longer need to support ubuntu 18, and in the mean time we can work on internalizing out version of node so it becomes more of an implementation detail (i.e. its doesn't end up poeple PATH). Keeping the old version of node for a while for internal use seems harmless enough.

It would be great if it could be like you said, especially the second one! This is just a solution to upgrade node while retaining compatibility with older versions

DreamOfIce:
Maybe we can detect if the user has a supported node, and for those who don't give a hint if they want to install the bundled version or manually install it and try again?

DreamOfIce:
So any progress yet?

sbc100:
While I would love to see progress here, I'm not sure when I will get a chance to look at it.   Is there some reason the status quo does not currently work for you?   Are you blocked in some way?   Is it the "emsdk version of node gets added to the PATH" issue?  My understanding is that fixing that would be of the most value of the most folks.

DreamOfIce:
It's good to see that node is no longer forced to be bundled 😄

DreamOfIce:
@sbc100 I have updated emscripten's Circle-CI to Debian Buster/Ubuntu Focal in https://github.com/emscripten-core/emscripten/pull/18907

sbc100:
> @sbc100 I have updated emscripten's Circle-CI to Debian Buster/Ubuntu Focal in [emscripten-core/emscripten#18907](https://github.com/emscripten-core/emscripten/pull/18907)

Now that we are not longer forcing our version of node on you, is there any reason to pursue upgrading our minimum linux version?   It seems like it can only hurt portability (i.e. there will be fewer disrtros that are then able to run emsdk)?

Shouldn't we always prefer to run our tests on the oldest possible distro we can?

DreamOfIce:
> > @sbc100 I have updated emscripten's Circle-CI to Debian Buster/Ubuntu Focal in [emscripten-core/emscripten#18907](https://github.com/emscripten-core/emscripten/pull/18907)
> 
> Now that we are not longer forcing our version of node on you, is there any reason to pursue upgrading our minimum linux version? It seems like it can only hurt portability (i.e. there will be fewer disrtros that are then able to run emsdk)?
> 
> Shouldn't we always prefer to run our tests on the oldest possible distro we can?

I don't think it's a good idea to use a system that is no longer maintained
BTW, the next major version (v20) of wasi sdk's pre-release
The version of libc used is very new (Debian 11 is not even enough)
Considering that we seem to use wasi-sdk in our CI, this is also a problem

sbc100:
> > > @sbc100 I have updated emscripten's Circle-CI to Debian Buster/Ubuntu Focal in [emscripten-core/emscripten#18907](https://github.com/emscripten-core/emscripten/pull/18907)
> > 
> > 
> > Now that we are not longer forcing our version of node on you, is there any reason to pursue upgrading our minimum linux version? It seems like it can only hurt portability (i.e. there will be fewer disrtros that are then able to run emsdk)?
> > Shouldn't we always prefer to run our tests on the oldest possible distro we can?
> 
> I don't think it's a good idea to use a system that is no longer maintained

I somewhat agree, but IMHO we would need a stronger reason than this to break compatibility with older systems.

If supporting users on older systems is easy then I think we should do it.   Continuing to build and test against Ubuntu/Bionic is currently not hard for us to do (i.e. we just do nothing)  so why abandon those users?

> BTW, the next major version (v20) of wasi sdk's pre-release The version of libc used is very new (Debian 11 is not even enough) Considering that we seem to use wasi-sdk in our CI, this is also a problem

Sure, if one of our dependencies starts to fail to run on bionic that may be enough of a motivation for us to abandon it too.

Another example would be if we decide we really need a nove v18 feature on our compiler then we may be forced to abandon bionic.

DreamOfIce:
@sbc100  When I update to emsdk 3.1.34 it seems that node 14.18.2 is still installed?
I have installed node18 and added it to the PATH
The following are installation logs:
``` sh
root@doi_wsl:~/emsdk# node -v
v18.14.0
root@doi_wsl:~/emsdk# ./emsdk install latest
Resolving SDK alias 'latest' to '3.1.34'
Resolving SDK version '3.1.34' to 'sdk-releases-2fdd6b9e5b67d5b62f84d0501a876513ff118ef1-64bit'
Installing SDK 'sdk-releases-2fdd6b9e5b67d5b62f84d0501a876513ff118ef1-64bit'..
Installing tool 'node-14.18.2-64bit'..
Downloading: /root/emsdk/zips/node-v14.18.2-linux-x64.tar.xz from https://storage.googleapis.com/webassembly/emscripten-releases-builds/deps/node-v14.18.2-linux-x64.tar.xz, 21848416 Bytes
Unpacking '/root/emsdk/zips/node-v14.18.2-linux-x64.tar.xz' to '/root/emsdk/node/14.18.2_64bit'
Done installing tool 'node-14.18.2-64bit'.
Installing tool 'releases-2fdd6b9e5b67d5b62f84d0501a876513ff118ef1-64bit'..
Downloading: /root/emsdk/zips/2fdd6b9e5b67d5b62f84d0501a876513ff118ef1-wasm-binaries.tbz2 from https://storage.googleapis.com/webassembly/emscripten-releases-builds/linux/2fdd6b9e5b67d5b62f84d0501a876513ff118ef1/wasm-binaries.tbz2, 348659789 Bytes
Unpacking '/root/emsdk/zips/2fdd6b9e5b67d5b62f84d0501a876513ff118ef1-wasm-binaries.tbz2' to '/root/emsdk/upstream'
Done installing tool 'releases-2fdd6b9e5b67d5b62f84d0501a876513ff118ef1-64bit'.
Done installing SDK 'sdk-releases-2fdd6b9e5b67d5b62f84d0501a876513ff118ef1-64bit'.
root@doi_wsl:~/emsdk# source ./emsdk_env.sh 
Adding directories to PATH:
PATH += /root/emsdk/upstream/emscripten
PATH += /root/emsdk/node/14.18.2_64bit/bin

PATH = /root/emsdk/upstream/emscripten:/root/emsdk/node/14.18.2_64bit/bin:/root/emsdk:/root/scripts:/root/.local/bin:/root/.local/share/pnpm:/root/.vscode-server/bin/ee2b180d582a7f601fa6ecfdad8d9fd269ab1884/bin/remote-cli:/root/.nix-profile/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/usr/lib/wsl/lib
root@doi_wsl:~/emsdk# node -v
v14.18.2
```

sbc100:
> @sbc100 When I update to emsdk 3.1.34 it seems that node 14.18.2 is still installed? I have installed node18 and added it to the PATH The following are installation logs:
> 
> ```shell
> root@doi_wsl:~/emsdk# node -v
> v18.14.0
> root@doi_wsl:~/emsdk# ./emsdk install latest
> Resolving SDK alias 'latest' to '3.1.34'
> Resolving SDK version '3.1.34' to 'sdk-releases-2fdd6b9e5b67d5b62f84d0501a876513ff118ef1-64bit'
> Installing SDK 'sdk-releases-2fdd6b9e5b67d5b62f84d0501a876513ff118ef1-64bit'..
> Installing tool 'node-14.18.2-64bit'..
> Downloading: /root/emsdk/zips/node-v14.18.2-linux-x64.tar.xz from https://storage.googleapis.com/webassembly/emscripten-releases-builds/deps/node-v14.18.2-linux-x64.tar.xz, 21848416 Bytes
> Unpacking '/root/emsdk/zips/node-v14.18.2-linux-x64.tar.xz' to '/root/emsdk/node/14.18.2_64bit'
> Done installing tool 'node-14.18.2-64bit'.
> Installing tool 'releases-2fdd6b9e5b67d5b62f84d0501a876513ff118ef1-64bit'..
> Downloading: /root/emsdk/zips/2fdd6b9e5b67d5b62f84d0501a876513ff118ef1-wasm-binaries.tbz2 from https://storage.googleapis.com/webassembly/emscripten-releases-builds/linux/2fdd6b9e5b67d5b62f84d0501a876513ff118ef1/wasm-binaries.tbz2, 348659789 Bytes
> Unpacking '/root/emsdk/zips/2fdd6b9e5b67d5b62f84d0501a876513ff118ef1-wasm-binaries.tbz2' to '/root/emsdk/upstream'
> Done installing tool 'releases-2fdd6b9e5b67d5b62f84d0501a876513ff118ef1-64bit'.
> Done installing SDK 'sdk-releases-2fdd6b9e5b67d5b62f84d0501a876513ff118ef1-64bit'.
> root@doi_wsl:~/emsdk# source ./emsdk_env.sh 
> Setting up EMSDK environment (suppress these messages with EMSDK_QUIET=1)
> Adding directories to PATH:
> PATH += /root/emsdk/upstream/emscripten
> PATH += /root/emsdk/node/14.18.2_64bit/bin
> 
> Setting environment variables:
> PATH = /root/emsdk/upstream/emscripten:/root/emsdk/node/14.18.2_64bit/bin:/root/emsdk:/root/scripts:/root/.local/bin:/root/.local/share/pnpm:/root/.vscode-server/bin/ee2b180d582a7f601fa6ecfdad8d9fd269ab1884/bin/remote-cli:/root/.nix-profile/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/usr/lib/wsl/lib
> root@doi_wsl:~/emsdk# node -v
> v14.18.2
> ```

It should only be added if you don't aready have `node` in our existing path.  If you install a modern version of node on your system then emsdk won't put its version in your PATH.. at least that is the idea.

kleisauke:
It looks like commit https://github.com/emscripten-core/emsdk/commit/c04b1bbb3a62bdfa4d69903c870075339b5c77ae applies the `activated_path_skip` thing only to the Node v14.15.5 binaries and _not_ to the v14.18.2 binaries.

sbc100:
> It looks like commit [c04b1bb](https://github.com/emscripten-core/emsdk/commit/c04b1bbb3a62bdfa4d69903c870075339b5c77ae) applies the `activated_path_skip` thing only to the Node v14.15.5 binaries and _not_ to the v14.18.2 binaries.

Oh, good catch!  We should fix that.

DreamOfIce:
> > It looks like commit [c04b1bb](https://github.com/emscripten-core/emsdk/commit/c04b1bbb3a62bdfa4d69903c870075339b5c77ae) applies the `activated_path_skip` thing only to the Node v14.15.5 binaries and _not_ to the v14.18.2 binaries.
> 
> Oh, good catch! We should fix that.

@sbc100 Any progress?

sbc100:
https://github.com/emscripten-core/emsdk/pull/1209

sbc100:
> We do ship Emsdk on Ubuntu 18 and some other Linuxes, though we do build our own Emsdk + Emscripten bundles manually, and don't rely on the automated CI here.

I'm curious if you have any idea when you guys might drop support for Ubuntu 18? 

I'm curious how many of our non-unity users might also be using Ubuntu 18.. can you elaborate on what kind of folks you know are running it?  e.g. is it corporate users who are on a slower upgrade path?

juj:
> I'm curious if you have any idea when you guys might drop support for Ubuntu 18?

Mirroring comment https://github.com/emscripten-core/emsdk/issues/1173#issuecomment-2124813940 here as well: we do no longer need Ubuntu 18 support, but currently are at Ubuntu 20. So happy to move forward here.

sbc100:
@DreamOfIce , would you have time to rebase this change?  I think we can move foward with landing it now.  If not I can roll it into #1387

DreamOfIce:
I think I can do this. What should be the target platforms for this upgrade, Ubuntu 20.04/22.04/24.04 and Node.js 18/20LTS?

DreamOfIce:
Okay, I'll resolve the conflict in these days.
However, as a reminder, there's less than a year left until the end of Ubuntu 20.04 free LTS.

sbc100:
> Okay, I'll resolve the conflict in these days. However, as a reminder, there's less than a year left until the end of Ubuntu 20.04 free LTS.

Those dates are not really the guide we use to guide the decision on our minimum version.   We are mode guided by the users of emsdk.   Whether or not canonical supports something is less relevenat.   Same with node... we support targeting versions of node that are long long out of LTS support.

sbc100:
For some reason the circleci jobs are not being kicked off here.   Could be a temporary issue with circleci?  Or something else?

sbc100 (approved):

README.md
```diff
@@ -67,8 +67,8 @@ https://emscripten.org/docs/building_from_source/toolchain_what_is_needed.html.
 - `python`: Version 2.7.0 or above.
 - `java`: For running closure compiler (optional)
 
-The emsdk pre-compiled binaries are built against Ubuntu/Xenial 16.04 LTS and
-therefore depend on system libraries compatible with versions of `glibc` and
+The emsdk pre-compiled binaries are built aginst Ubuntu/Focal 20.04 LTS and
```
sbc100, line 70:
```suggestion
The emsdk pre-compiled binaries are built against Ubuntu/Focal 20.04 LTS and
```

README.md
```diff
@@ -67,8 +67,8 @@ https://emscripten.org/docs/building_from_source/toolchain_what_is_needed.html.
 - `python`: Version 2.7.0 or above.
 - `java`: For running closure compiler (optional)
 
-The emsdk pre-compiled binaries are built against Ubuntu/Xenial 16.04 LTS and
-therefore depend on system libraries compatible with versions of `glibc` and
+The emsdk pre-compiled binaries are built against Ubuntu/Focal 20.04 LTS and
+therefore depend on system libraries compatiable with versions of `glibc` and
```
sbc100, line 71:
```suggestion
therefore depend on system libraries compatible with versions of `glibc` and
```

sbc100:
I'm not sure whats up with CI here so testing this is a separate PR: https://github.com/emscripten-core/emsdk/pull/1390.  If tests pass there I'll land this PR here.

.circleci/config.yml
```diff
@@ -50,22 +50,26 @@ commands:
 
 jobs:
   flake8:
-    executor: bionic
+    executor: ubuntu
     steps:
       - checkout
       - run:
-          name: install pip
+          name: install pip3
```
sbc100, line 57:
```suggestion
          name: install pip
```

.circleci/config.yml
```diff
@@ -50,22 +50,26 @@ commands:
...
       - run: python3 -m pip install --upgrade pip
-      - run: python2 -m pip install flake8==3.9.2
       - run: python3 -m pip install flake8==3.9.2
+      - run: python3 -m flake8 --show-source --statistics --extend-exclude=./scripts
+      - run: 
+          name: install pip2
+          command: apt-get install -q -y python-pip
+      - run: python2 -m pip install --upgrade pip
+      - run: python2 -m pip install flake8==3.9.2
       - run: python2 -m flake8 --show-source --statistics --extend-exclude=./scripts
```
sbc100, line 69:
```suggestion
      - run: 
          name: install pip2
          command: apt-get install -q -y python-pip
      - run: python2 -m pip install --upgrade pip
      - run: python2 -m pip install flake8==3.9.2
      - run: python2 -m flake8 --show-source --statistics --extend-exclude=./scripts
```

sbc100, line 69:
```suggestion
      - run: 
          name: install pip2
          command: apt-get install -q -y python-pip
      - run: python2 -m pip install --upgrade pip
      - run: python2 -m pip install flake8==3.9.2
      - run: python2 -m flake8 --show-source --statistics --extend-exclude=./scripts
```
```suggestion
```

PR #1186, opened 2023-02-14
Windows Arm64 support for emsdk.py only

Blackhex:
This PR add support for Visual Studio Build Tools 2022 (Preview) compilation of packages using emsdk.py on and for Windows Arm64.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -142,9 +142,6 @@ elif machine.endswith('86'):
   ARCH = 'x86'
 elif machine.startswith('aarch64') or machine.lower().startswith('arm64'):
   ARCH = 'aarch64'
-  if WINDOWS:
-      errlog('No support for Windows on Arm, fallback to x64')
-      ARCH = 'x86_64'
 elif machine.startswith('arm'):
   ARCH = 'arm'
 else:
@@ -256,7 +253,11 @@ def vswhere(version):
     if not program_files:
       program_files = os.environ['ProgramFiles']
     vswhere_path = os.path.join(program_files, 'Microsoft Visual Studio', 'Installer', 'vswhere.exe')
-    output = json.loads(subprocess.check_output([vswhere_path, '-latest', '-version', '[%s.0,%s.0)' % (version, version + 1), '-requires', 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64', '-property', 'installationPath', '-format', 'json']))
+    # Source: https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022
+    tools_arch = 'ARM64' if ARCH == 'aarch64' else 'x86.x64'
+    # The "-products *" allows detection of Build Tools, the "-prerelease" allows detection of Preview version
+    # of Visual Studio and Build Tools.
+    output = json.loads(subprocess.check_output([vswhere_path, '-latest', '-products', '*', '-prerelease', '-version', '[%s.0,%s.0)' % (version, version + 1), '-requires', 'Microsoft.VisualStudio.Component.VC.Tools.' + tools_arch, '-property', 'installationPath', '-format', 'json']))
     return str(output[0]['installationPath'])
   except Exception:
     return ''
@@ -1014,6 +1015,32 @@ def xcode_sdk_version():
     return subprocess.checkplatform.mac_ver()[0].split('.')
 
 
+def cmake_target_platform(tool):
+  # Source: https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2017%202022.html#platform-selection
+  if hasattr(tool, 'arch'):
+    if tool.arch == 'aarch64':
+      return 'ARM64'
+    elif tool.arch == 'x86_64':
+      return 'x64'
+    elif tool.arch == 'x86':
+      return 'Win32'
+  if ARCH == 'aarch64':
+    return 'ARM64'
+  else:
+    return 'x64' if tool.bitness == 64 else 'Win32'
+
+
+def cmake_host_platform():
+  # Source: https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2017%202022.html#toolset-selection
+  arch_to_cmake_host_platform = {
+    'aarch64': 'ARM64',
+    'arm': 'ARM',
+    'x86_64': 'x64',
+    'x86': 'x86'
+  }
+  return arch_to_cmake_host_platform[ARCH]
+
+
 def get_generator_and_config_args(tool):
   args = []
   cmake_generator = CMAKE_GENERATOR
@@ -1021,8 +1048,8 @@ def get_generator_and_config_args(tool):
     # With Visual Studio 16 2019, CMake changed the way they specify target arch.
     # Instead of appending it into the CMake generator line, it is specified
     # with a -A arch parameter.
-    args += ['-A', 'x64' if tool.bitness == 64 else 'x86']
-    args += ['-Thost=x64']
+    args += ['-A', cmake_target_platform(tool)]
+    args += ['-Thost=' + cmake_host_platform()]
   elif 'Visual Studio' in CMAKE_GENERATOR and tool.bitness == 64:
     cmake_generator += ' Win64'
     args += ['-Thost=x64']
@@ -1831,6 +1858,10 @@ class Tool(object):
     elif hasattr(self, 'git_branch'):
       success = git_clone_checkout_and_pull(url, self.installation_path(), self.git_branch)
     elif url.endswith(ARCHIVE_SUFFIXES):
+      global ARCH
+      if WINDOWS and ARCH == 'aarch64':
+        errlog('No support for Windows on Arm, fallback to x64')
+        ARCH = 'x86_64'
       success = download_and_unzip(url, self.installation_path(), filename_prefix=getattr(self, 'zipfile_prefix', ''))
     else:
       assert False, 'unhandled url type: ' + url


emsdk.py
```diff
@@ -256,7 +253,8 @@ def vswhere(version):
     if not program_files:
       program_files = os.environ['ProgramFiles']
     vswhere_path = os.path.join(program_files, 'Microsoft Visual Studio', 'Installer', 'vswhere.exe')
-    output = json.loads(subprocess.check_output([vswhere_path, '-latest', '-version', '[%s.0,%s.0)' % (version, version + 1), '-requires', 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64', '-property', 'installationPath', '-format', 'json']))
+    tools_arch = ('ARM64' if ARCH == 'aarch64' else 'x86.x64')
+    output = json.loads(subprocess.check_output([vswhere_path, '-latest', '-products', '*', '-prerelease', '-version', '[%s.0,%s.0)' % (version, version + 1), '-requires', 'Microsoft.VisualStudio.Component.VC.Tools.' + tools_arch, '-property', 'installationPath', '-format', 'json']))
```
Blackhex, line 257: The `-products *` allows detection of Build Tools, the `-prerelease` allows detection of Preview version of Visual Studio and Build Tools.

emsdk.py
```diff
@@ -1014,15 +1012,40 @@ def xcode_sdk_version():
...
+
+def decide_cmake_host_platform():
+  if ARCH == 'aarch64':
+    return 'ARM64'
+  elif ARCH == 'arm':
+    return 'ARM'
+  elif ARCH == 'x86_64':
+    return 'x64'
+  else:
+    return 'x86'
```
sbc100, line 1037: How about writing this a dict so it will crash accordingly on unknown arch instead of defaulting to x86. e.g.


```
  cmake_arch = { 'aarch64': 'arm64' ... }
  return cmake_arch[ARCH]
```

Also should ARM64 and ARM be capitalized here.  I'd rather keep them all lowercase if that works?

emsdk.py
```diff
@@ -256,7 +253,8 @@ def vswhere(version):
     if not program_files:
       program_files = os.environ['ProgramFiles']
     vswhere_path = os.path.join(program_files, 'Microsoft Visual Studio', 'Installer', 'vswhere.exe')
-    output = json.loads(subprocess.check_output([vswhere_path, '-latest', '-version', '[%s.0,%s.0)' % (version, version + 1), '-requires', 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64', '-property', 'installationPath', '-format', 'json']))
+    tools_arch = ('ARM64' if ARCH == 'aarch64' else 'x86.x64')
```
sbc100, line 256: Can arm64 be lowercase here? 

Should we switch our internal ARCH=aarch64 to ARCH=arm64 maybe?  

No need for brackets around this expression.

emsdk.py
```diff
@@ -1014,15 +1012,40 @@ def xcode_sdk_version():
     return subprocess.checkplatform.mac_ver()[0].split('.')
 
 
+def decide_cmake_target_platform(tool):
+  if hasattr(tool, 'arch') and tool.arch == 'aarch64':
+    return 'ARM64'
+  elif hasattr(tool, 'arch') and tool.arch == 'x86_64':
+    return 'x64'
+  elif hasattr(tool, 'arch') and tool.arch == 'x86':
+    return 'x86'
```
sbc100, line 1021: How about just one `if hasattr(tool, 'arch')` here.

e.g.

```
if hasattr(tool, 'arch'):
   tools_arch_map = {'aarch64': ARM64' ...}
   return tool_arch_map[tool.arch]
// Fall back to host arch. 
if ARCH == 'aarch64':
    return 'ARM64'
  else:
    return 'x64' if tool.bitness == 64 else 'x86'
```
 
Should we call `decide_cmake_host_platform` for the fallback case?

emsdk.py
```diff
@@ -1360,7 +1383,7 @@ def build_binaryen_tool(tool):
     return False
 
   # Make
-  success = make_build(build_root, build_type, 'x64' if tool.bitness == 64 else 'Win32')
+  success = make_build(build_root, build_type)
```
sbc100, line 1386: I guess maybe the removal of this third unused argument could be its own NFC (non-functional-change)?  But its not a big deal if you want to leave it as part of this one..

sbc100 (approved):
lgtm with comments addressed.

Thanks for working on this!

emsdk.py
```diff
@@ -1014,15 +1012,40 @@ def xcode_sdk_version():
...
   args = []
   cmake_generator = CMAKE_GENERATOR
   if 'Visual Studio 16' in CMAKE_GENERATOR or 'Visual Studio 17' in CMAKE_GENERATOR:  # VS2019 or VS2022
     # With Visual Studio 16 2019, CMake changed the way they specify target arch.
     # Instead of appending it into the CMake generator line, it is specified
     # with a -A arch parameter.
-    args += ['-A', 'x64' if tool.bitness == 64 else 'x86']
-    args += ['-Thost=x64']
+    args += ['-A', decide_cmake_target_platform(tool)]
+    args += ['-Thost=' + decide_cmake_host_platform()]
```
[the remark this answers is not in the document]
Blackhex: I wanted to be consistent with with `decide_cmake_build_type` but OK.

emsdk.py
Blackhex: This is driven by VS component IDs so it should be capital: https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022.

`aarch64` came from Unix-like systems and the value is used both for Linux, MacOS and Windows. I'd keep it as is unless we wan't to distinguish between different OSes.

emsdk.py
Blackhex: The code style has changed and the values fixed according to the source references added. The host platform and target platform values has different syntax as you can see in the sources.

emsdk.py
Blackhex: The code style fixed and the values changed according to the source reference.

emsdk.py
Blackhex: Are you sure this is unused?

Blackhex: You decide, I had those changes written before VS 2022 support was added upstream so when I was updating my old changes including VS 2022 support to the new upstream state, this change was some kind of leftover of the code adding the VS 2022 support.

sbc100: I always prefer smaller and separate changes.  So I would prefer a separate PR in this case if you have time.

emsdk.py
```diff
@@ -256,7 +253,9 @@ def vswhere(version):
     if not program_files:
       program_files = os.environ['ProgramFiles']
     vswhere_path = os.path.join(program_files, 'Microsoft Visual Studio', 'Installer', 'vswhere.exe')
-    output = json.loads(subprocess.check_output([vswhere_path, '-latest', '-version', '[%s.0,%s.0)' % (version, version + 1), '-requires', 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64', '-property', 'installationPath', '-format', 'json']))
+    # Source: https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022
+    tools_arch = 'ARM64' if ARCH == 'aarch64' else 'x86.x64'
```
sbc100, line 257: If this is supposed to be uppercase why isn't the else case "X86.X64"?

emsdk.py
sbc100: Perhaps put this in a comment on the line above?

emsdk.py
```diff
@@ -1014,15 +1013,48 @@ def xcode_sdk_version():
     return subprocess.checkplatform.mac_ver()[0].split('.')
 
 
+def cmake_target_platform_no_tool_arch(tool):
+    if ARCH == 'aarch64':
+      return 'ARM64'
+    else:
+      return 'x64' if tool.bitness == 64 else 'Win32'
```
sbc100, line 1020: Is the return value here supposed to be uppercase?

What happens when this function is called on non-windows platforms?

emsdk.py
```diff
@@ -1014,15 +1013,48 @@ def xcode_sdk_version():
...
+    if tool.arch == 'aarch64':
+      return 'ARM64'
+    elif tool.arch == 'x86_64':
+      return 'x64'
+    elif tool.arch == 'x86':
+      return 'Win32'
+    else:
+      return cmake_target_platform_no_tool_arch(tool)
+  else:
+    return cmake_target_platform_no_tool_arch(tool)
```
sbc100, line 1035: I think you can remove the to final `else` blocks here and just return `cmake_target_platform_no_tool_arch`.  

Also since `cmake_target_platform_no_tool_arch` is very simple and (after the above change) has just one calls site maybe just inline it here?

emsdk.py
```diff
@@ -1014,15 +1013,48 @@ def xcode_sdk_version():
...
+
+
+def cmake_host_platform():
```
sbc100, line 1038: Is this a arch breakdown a cmake thing or visual studio thing?  If its the latter maybe call this `vs_host_platform`?

That would also helper reader know this is a windows-only function.

emsdk.py
Blackhex: I was referring to Arm64 variant only. x86.x64 should be lower case. See [https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022](https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022) for the reference.

emsdk.py
Blackhex: AFAIK, it's CMake thing when the generator (`-G`) is `Visual Studio X YYYY` and other generators do not use it but `-A` option is meant to be generic so there might be other uses of it (at least in the future).

emsdk.py
Blackhex: According to [https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2017%202022.html#platform-selection](https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2017%202022.html#platform-selection) yes. This method should not be called on non-Windows platforms as it is called only when generator is "Visual Studio".

sbc100:
lgtm!  Thanks for working on this.

I suppose the next step would be try to build release binaries for windows/arm64 so folks don't need to build from source all the time.  Do you think we are that point where that it worth the investment?  How many windows/arm64 users are out there right now?

emsdk.py
```diff
@@ -142,9 +142,6 @@ def exit_with_error(msg):
   ARCH = 'x86'
 elif machine.startswith('aarch64') or machine.lower().startswith('arm64'):
   ARCH = 'aarch64'
-  if WINDOWS:
-      errlog('No support for Windows on Arm, fallback to x64')
-      ARCH = 'x86_64'
```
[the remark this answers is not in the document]
Blackhex: You are right, as far as there are no publicly available Arm64 binearies we should instlal the x64 ones running in emulation. But having this code here prevents form possiblility to build the Arm64 binaries from sources using emsdk.py. For that reason I've move this check into the `install_tool()` method. The code is not nice but this should be only temporary solution.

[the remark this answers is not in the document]
sbc100: One option is that we drop the hacks and folks can use `EMSDK_ARCH=x86_64 emsdk install latest` if they want to force the install the x86 binaries?   Or do you think we should continue to make it automatic?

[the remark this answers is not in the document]
Blackhex: I guess the hack is better because it keeps the current behaviour. I can rember when I started with emsdk that it took me a while to figure out that the `EMSDK_ARCH` even exists. On the other, hand note that there is this "strange" behavior: If you have Arm64 Python installed on your system and you do `emsdk.bat install latest` and `emsdk.bat activate latest`, the x64 version of emsdk Python is installed and activated and since then, `emsdk.bat install xxx` installs x64 versions anyway (without the error message). You either need to force Arm64 with `EMSDK_ARCH=aarch64` or install using `python emsdk.py install xxx`. If your system Python is x64, you need to force `EMSDK_ARCH=aarch64` in all cases.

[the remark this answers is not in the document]
pbo-linaro: Yes, the python based system detection is a bit boring, as the architecture of your python dictates which platform you have. As you mentioned, if you have an x64 python (emulated on arm64), it will say you're on an x64 cpu.

The other way to detect this under Windows is to use PROCESSOR_ARCHITECTURE env var. Same joke, from cmd.exe, it says you're on an arm64 system, and from powershell, that you're on an x64 system (because default powershell install is x64).

For this kind of case, IMHO, having an explicit option (available through help) would be the best solution. Or at least, to mention ```EMSDK_ARCH``` in help and official documentation.

Blackhex:
> I suppose the next step would be try to build release binaries for windows/arm64 so folks don't need to build from source all the time. Do you think we are that point where that it worth the investment? How many windows/arm64 users are out there right now?

@sbc100 Thank you for your help with the review and merging.

We already have some experimental binaries and now we are ensuring they can pass all the emscripten repository unit tests and some useful projects can be compiled with them. You can check out our https://github.com/Windows-on-ARM-Experiments/emsdk fork. As our team's mission is an enablement of smooth developer experience on Windows Arm64, we are happy to listen any requests or suggestions in that regards.

sbc100:
Great.  Once you have some binaries you are happen with we could consider adding them emscripten-releases-tags.json like we have to `latest-arm64-linux`:

```
{                                                                                   
  "aliases": {                                                                      
    "latest": "3.1.31",                                                             
    "latest-sdk": "latest",                                                         
    "latest-arm64-linux": "3.1.21",                                                 
    "latest-64bit": "latest",                                                       
    "sdk-latest-64bit": "latest",                                                   
    "latest-upstream": "latest",                                                                              
    "latest-clang-upstream": "latest",                                           
    "latest-releases-upstream": "latest"                                         
  }, 
```

Note that `latest-arm64-linux` is not necessarily up-to-date because we don't build the binaries with the same automatic process.  We could potentially do the same for arm64-windows?

Blackhex:
Why not target to build the Windows Arm64 regularly together with x64 ones? Cross-compilation is always the option.

sbc100:
Sure, we do that for macOS, so if its easy for windows we can do that too

PR #1220, opened 2023-05-28
Docker: avoid sourcing `emsdk_env.sh` in interactive shells

kleisauke:
Instead, rely on the already set `PATH` environment variable.

---

This would ensure that a custom cache directory via `EM_CACHE` is honored rather than being ignored.

Current behavior: 
```console
$ docker run -it --rm -v $(pwd):/src -e EM_CACHE=/src/build/emcache emscripten/emsdk:3.1.39
...
Clearing existing environment variable: EM_CACHE
root@9c00b2231a23:/src# echo $EM_CACHE


```

New behavior: 
```console
$ docker run -it --rm -v $(pwd):/src -e EM_CACHE=/src/build/emcache emscripten/emsdk
root@7ed76e8b10f9:/src# echo $EM_CACHE
/src/build/emcache
```

Context: https://github.com/kleisauke/wasm-vips/issues/49.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -61,6 +61,7 @@ FROM ubuntu:jammy AS stage_deploy
 COPY --from=stage_build /emsdk /emsdk
 
 # Set up emsdk environment
+# This corresponds to the environment variables set during `source ./emsdk_env.sh`
 ENV EMSDK=/emsdk \
     EMSDK_NODE=/emsdk/node/15.14.0_64bit/bin/node \
     PATH="/emsdk:/emsdk/upstream/emscripten:/emsdk/upstream/bin:/emsdk/node/15.14.0_64bit/bin:${PATH}"
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -60,9 +60,7 @@ FROM ubuntu:jammy AS stage_deploy
 
 COPY --from=stage_build /emsdk /emsdk
 
-# Fallback in case Emscripten isn't activated.
-# This will let use tools offered by this image inside other Docker images
-# (sub-stages) or with custom / no entrypoint
+# Set up emsdk environment
 ENV EMSDK=/emsdk \
     EMSDK_NODE=/emsdk/node/15.14.0_64bit/bin/node \
     PATH="/emsdk:/emsdk/upstream/emscripten:/emsdk/upstream/bin:/emsdk/node/15.14.0_64bit/bin:${PATH}"
@@ -79,7 +77,6 @@ RUN echo "## Create emscripten user (1000:1000)" \
     && groupadd --gid 1000 emscripten \
     && useradd --uid 1000 --gid emscripten --shell /bin/bash --create-home emscripten \
     && echo "umask 0000" >> /etc/bash.bashrc \
-    && echo ". /emsdk/emsdk_env.sh" >> /etc/bash.bashrc \
     && echo "## Done"
 
 # ------------------------------------------------------------------------------
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -77,7 +77,6 @@ ENV EMSDK=/emsdk \
 RUN echo "## Create emscripten user (1000:1000)" \
     && groupadd --gid 1000 emscripten \
     && useradd --uid 1000 --gid emscripten --shell /bin/bash --create-home emscripten \
-    && echo "umask 0000" >> /etc/bash.bashrc \
     && echo "## Done"
 
 # ------------------------------------------------------------------------------


sbc100:
Do we want to support this though?  That point of emsdk_env is that it sets up the current env to point to emsdk tools (and emsdk cache).   If we want to change this behaviour maybe we should change emsdk_env instead?

docker/Dockerfile
```diff
@@ -60,9 +60,7 @@ FROM ubuntu:jammy AS stage_deploy
 
 COPY --from=stage_build /emsdk /emsdk
 
-# Fallback in case Emscripten isn't activated.
-# This will let use tools offered by this image inside other Docker images
-# (sub-stages) or with custom / no entrypoint
+# Set up emsdk environment
 ENV EMSDK=/emsdk \
     EMSDK_NODE=/emsdk/node/15.14.0_64bit/bin/node \
```
sbc100, line 65: Seeing this now I think we should remove `EMSDK_NODE` here.. its not used for anything that I know of.

sbc100: Also, EMSDK it only used locally within this file.. is there some way to declare it as local?

kleisauke:
> Do we want to support this though? That point of emsdk_env is that it sets up the current env to point to emsdk tools (and emsdk cache). If we want to change this behaviour maybe we should change emsdk_env instead?

The `emsdk_env.sh`  script (and umask thing) was only run in interactive shells, see for example:
```console
$ docker run -it --rm -v $(pwd):/src emscripten/emsdk:3.1.39 bash -c 'umask'
0022
$ docker run -it --rm -v $(pwd):/src emscripten/emsdk:3.1.39
PATH = /emsdk:/emsdk/upstream/emscripten:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
root@8e8d456cdb1e:/src# umask
0000
```

So, it's therefore better to align the behavior of interactive and non-interactive shells (with the latter being much more common, I think).

sbc100:
> > Do we want to support this though? That point of emsdk_env is that it sets up the current env to point to emsdk tools (and emsdk cache). If we want to change this behaviour maybe we should change emsdk_env instead?
> 
> The `emsdk_env.sh` script (and umask thing) was only run in interactive shells, see for example:
> 
> ```
> $ docker run -it --rm -v $(pwd):/src emscripten/emsdk:3.1.39 bash -c 'umask'
> 0022
> $ docker run -it --rm -v $(pwd):/src emscripten/emsdk:3.1.39
> Setting up EMSDK environment (suppress these messages with EMSDK_QUIET=1)
> Setting environment variables:
> PATH = /emsdk:/emsdk/upstream/emscripten:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
> root@8e8d456cdb1e:/src# umask
> 0000
> ```
> 
> So, it's therefore better to align the behavior of interactive and non-interactive shells (with the latter being much more common, I think).

The ideas is that emsdk_env is in charge to setting everything up.. but I guess we don't need that give that we are manually constructing that PATH.

However, emsdk_env is the official recommended way to setup the PATH so I guess I would prefer to have emsdk_env run in all cases.. and not manually inject `upstream/bin` into the PATH.   Hardcoding those PATHs kind of goes against the idea of how emsdk is supposed to work.

kleisauke:
> The ideas is that emsdk_env is in charge to setting everything up.. but I guess we don't need that give that we are manually constructing that PATH.
> 
> However, emsdk_env is the official recommended way to setup the PATH so I guess I would prefer to have emsdk_env run in all cases.. and not manually inject `upstream/bin` into the PATH. Hardcoding those PATHs kind of goes against the idea of how emsdk is supposed to work.

I see. Unfortunately, you cannot set those `ENV` entries with the result of a command and `RUN` entries cannot modify environment variables (since they are executed in a new shell).

Another option would be to avoid the need for these environment variables altogether by installing the pre-built binaries in systems locations such as `/usr` or `/usr/local`; it will be isolated within the Docker image anyway. Though, I'm not sure how much work this would be.

> Can you split out the removal of fastcomp and EMSDK_NODE into its own separate PR?

Sure, see PR #1225 and #1226.

sbc100:
> > The ideas is that emsdk_env is in charge to setting everything up.. but I guess we don't need that give that we are manually constructing that PATH.
> > However, emsdk_env is the official recommended way to setup the PATH so I guess I would prefer to have emsdk_env run in all cases.. and not manually inject `upstream/bin` into the PATH. Hardcoding those PATHs kind of goes against the idea of how emsdk is supposed to work.
> 
> I see. Unfortunately, you cannot set those `ENV` entries with the result of a command and `RUN` entries cannot modify environment variables (since they are executed in a new shell).
> 
> Another option would be to avoid the need for these environment variables altogether by installing the pre-built binaries in systems locations such as `/usr` or `/usr/local`; it will be isolated within the Docker image anyway. Though, I'm not sure how much work this would be.

Can't we just add `. /emsdk/.emsdk_env.sh` to a place where it will always be run?  Would `~/.bash_profile` be better than `~/.back_profile`?  

Then we can remove the PATH additions above?

kleisauke:
Closing in favor of #1227.

My only concern is that it would break compatibility for users that set the `EM_CACHE` env variable, which is useful for preserving cache across runs. However, one could use a custom entry point to fix that.

PR #1225, opened 2023-06-01
Docker: remove redundant `EMSDK_NODE` env

kleisauke:
Split out from PR #1220.


kleisauke:
FWIW, I just noticed that `emsdk_env.sh` still sets this environment variable.
https://github.com/emscripten-core/emsdk/blob/57539391aaa356df6e5aba409b5819dc4aaccea2/emsdk_manifest.json#L286

sbc100:
> FWIW, I just noticed that `emsdk_env.sh` still sets this environment variable.

Indeed, but none of our tools rely on this.   This is more for folks who might want to run the emsdk version node, even though they have their own version in the PATH.   Perhaps this is useless?  .. but is not needed by emsdk or emscripten.

sbc100 (approved):

PR #1227, opened 2023-06-01
Docker: use a entrypoint script

kleisauke:
In favor of `/etc/bash.bashrc`, which is only read for interactive shells.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,8 @@
 # Ignore everything
 *
 
-# Allow to run the test script inside the Docker container
+# Allow the entrypoint/test script inside the Docker container
+!/docker/entrypoint.sh
 !/docker/test_dockerimage.sh
 
 # Allow license file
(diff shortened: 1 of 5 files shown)


sbc100 (commented):
Looks great.  I didn't know about that.  

@trzecieu WDYT?

kleisauke:
Ah, I forgot that this would break extended Docker images, for example:
```Dockerfile
FROM emscripten/emsdk:3.1.40

RUN embuilder build zlib
```

```console
$ docker build -t test .
STEP 1/2: FROM emscripten/emsdk:3.1.40
STEP 2/2: RUN embuilder build zlib
/bin/sh: 1: embuilder: not found
Error: building at STEP "RUN embuilder build zlib": while running runtime: exit status 127
```

sbc100:
Maybe still worth landing this, but while preserving the PATH with a comment?

sbc100:
Can you rebase this change?  (assuming we still want to land it?)

kleisauke:
Just rebased, sorry for the delay.

sbc100 (approved):

Issue #1206, opened 2023-03-16
Docker image for ARM64 (Apple Silicon)


CodeWitchBella:
I discovered that official docker images at [emscripten/emsdk](https://hub.docker.com/r/emscripten/emsdk) only have AMD64 builds of emscripten. So while they work on apple's M1/M2, they are much slower than native. There are ARM64 builds of emsdk available (#547), so it shouldn't be hard at this point to produce docker images. I tried running `make version=3.1.33` in `docker` directory of this repo on my macbook and it produced an image which is ARM64. I did not yet have time to test if it works.

#1000 is closed as duplicate of #547, but it is not solved yet. It was more like a prerequisite than a solution.

While I can build the image myself, it would be much more convenient, if there was a prebuilt image of emscripten readily available. Could we get the community build bot to also push docker images?

CodeWitchBella:
According to [this](https://docs.docker.com/build/building/multi-platform/), it should be enough to specify `--platform linux/arm64` with the docker build command and push that. That assumes that you have working QEMU platform emulation (true for docker desktop but not necessarily for docker installed from Linux packages, I think). Then I assume that you can just push that to the registry and it will add new platform to the existing image, but I never actually tried this.

Anyway, thanks for the quick response. Much appreciated 🙂

lesomnus:
```
docker buidx build --platform linux/amd64,linux/arm64 --push ...
```
will build both _amd64_ and _arm64_ image in a single manifest and it will push automatically after build finished.
Im looking forward to get _arm64_ `emscripten/emsdk` on DockerHub soon.
Thanks for your work.

Issue #1211, opened 2023-04-23
Build linux/arm64 docker images


justjake:
The current docker images ([example](https://hub.docker.com/layers/emscripten/emsdk/3.1.36/images/sha256-36b1db2e65632778a5a2e2d41222fb166f02e218916ca3bd3733e0483a5bbe0d?context=explore)) are published only for linux/amd64.

On my M1 Macbook Pro, using these images tends to take 10-100x longer than using native ARM64 builds of Emscripten from homebrew. However, Homebrew will randomly upgrade my Emscripten version due to python changing or something.

It would be great to be able to use the EMSDK docker image for stability/reproducibility without suffering from super slow builds on my laptop.

It appears that CircleCI supports ARM, so perhaps all that's needed is adding an extra build variant for Dockerfile-related build jobs?

https://circleci.com/docs/using-arm/

sbc100:
While I agree we should provide arm64 docker images, if you want a solution that works today you could just install emsdk the normal way on macOS.   The emsdk binaries provided via `emsdk install latest` are built for arm64.

sbc100:
I see so the tags don't need to be different?   Would you be interested in sending a PR for that?  

Looks like a simple update to remove the specific `ARM64_IMAGE_TAG` would do it?

Are you sure it won't clobber the x86 version?

radiantly:
Sure, let me take a look at this over the weekend and open a PR. As long as we push the image correctly I don't see a reason for the existing amd64 version to be affected.

PR #1231, opened 2023-06-08
Auto-publish Arm64 images when emscripten-releases-tags.json is updated

jamsinclair:
Fixes: #1206 #1211

Hi Maintainers, firstly thank you for all your work around the emscripten project!

I would like to propose this PR as a potential solution to some issues identified in #1206 and #1211
1. Adds a CircleCI Job that will only run on pushes to the main branch. It will run on a CircleCI [Arm machine](https://circleci.com/docs/using-arm/).
1. The job will continue only if the `emscripten-releases-tags.json` file has changed and there isn't already a published version for the current `latest-arm64-linux` version
1. The job will then proceed to build an arm64 image, create an alias tag and push to dockerhub

⚠️ I haven't tested the pipeline on a CircleCI build. There might be some bugs or fixes to make still. 

I understand if this is not an extra complexity you want to support for the project yet, if so feel free to close the PR 🙇


.circleci/config.yml
```diff
@@ -172,6 +172,51 @@ jobs:
...
+          name: extract latest emscripten arm64 linux version
+          command: |
+            export ARM64_VERSION=$(grep -oP '(?<="latest-arm64-linux": ")[^"]*' emscripten-releases-tags.json)
+            export ARM64_IMAGE_TAG=${ARM64_VERSION}-arm64
+            echo "ARM64_VERSION=$ARM64_VERSION"
+            echo "ARM64_IMAGE_TAG=$ARM64_IMAGE_TAG"
+      - run:
+          name: check if docker image already exists
+          command: |
+            DOCKER_IMAGE_EXISTS=$(curl -s -o /dev/null -w "%{http_code}" https://hub.docker.com/v2/repositories/emscripten/emsdk/tags/$ARM64_IMAGE_TAG)
```
jamsinclair, line 199: These run steps could be moved to a separate shell script if you would prefer to keep the `config.yml` file cleaner.

.circleci/config.yml
```diff
@@ -172,6 +172,51 @@ jobs:
             docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
             make -C ./docker version=${CIRCLE_TAG} alias=latest push
 
+  publish-docker-image-arm64:
+    machine:
+      image: ubuntu-2004:current
+    resource_class: arm.medium
```
jamsinclair, line 178: This would configure the job to run in Arm environment https://circleci.com/docs/using-arm/

This would make building an Arm image considerably easier, as I'm unsure if the usual amd64 CircleCI images have qemu setup to allow building different platform images.

.circleci/config.yml
```diff
@@ -172,6 +172,51 @@ jobs:
...
+          name: build
+          command: make -C ./docker version=${ARM64_VERSION} build
+      - run:
+          name: test
+          command: make -C ./docker version=${ARM64_VERSION} test
+      - run:
+          name: push image
+          command: |
+            docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
+            make -C ./docker version=${ARM64_VERSION} alias=${ARM64_IMAGE_TAG} only_alias=true push
```
jamsinclair, line 218: What would the preferred tag be for these images? This PR would currently set it to `<emscripten-version>-arm64` e.g. `3.1.33-arm64`

sbc100: Does docker have a convention for this?     Is the architecture not an inherent part of the image?    What happens if you try to install the normal x64 imagine and arm host?

.circleci/config.yml
sbc100: Seems reasonable

.circleci/config.yml
jamsinclair: Great questions and thanks for prompt reply!

> Does docker have a convention for this? Is the architecture not an inherent part of the image?

Docker does offer an experimental solution with [Docker Manifest](https://docs.docker.com/engine/reference/commandline/manifest/).

[The official docker guide](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/) would suggest tagging your images with an arch suffix e.g. build each arch image separately and  tag with `3.1.33-amd64` and `3.1.33-arm64v8`. Then with the `docker manifest` tool you can alias these tags under `3.1.33` and push to the docker repo. When a consumer pulls an image like `emscripten/emsdk:3.1.33`, docker can auto select the appropriate arch for the host machine.

Moving to the manifest solution will increase the complexity of the current CI docker steps. Perhaps something that could be followed up with in the future?

> What happens if you try to install the normal x64 image and arm host?

Let me know if I've misunderstood. If a user installed both an `amd64` and `arm64v8` images they would have different docker tags so there's no conflict. The users would have to explicitly opt-in to using ARM based images, unless the manifest solution above was implemented.

.circleci/config.yml
```diff
@@ -248,6 +293,10 @@ workflows:
               ignore: /.*/
             tags:
               only: /.*/
+      - publish-docker-image-arm64:
+          filters:
+            branches:
+              only: main
```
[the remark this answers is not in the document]
jamsinclair: Good point, I didn't see that updates are always tagged. As seen at https://github.com/emscripten-core/emsdk/commit/a2ccccba58918706f815d806460d09310fe236d9

Will update 👍

jozefchutka:
Just checking, is this the issue to subscribe to observe the progress on latest emsdks being available for dockers? - Issue reported on https://github.com/emscripten-core/emscripten/issues/19275 ?

sbc100:
> Just checking, is this the issue to subscribe to observe the progress on latest emsdks being available for dockers? - Issue reported on [emscripten-core/emscripten#19275](https://github.com/emscripten-core/emscripten/issues/19275) ?

Specifically, this issue related to publish arm64 versions of the docker image.  The x86_64 images are already auto-published continuously.

jamsinclair:
Hey @sbc100 is there anything you would like to see to make this PR a possibility? 🤓

As it stands, this pull request would automatically publish a special image tag, `<version>-arm64`, when a change like #1288 is made. It would be an opt-in only experience, a user will have to use an `*-arm64` image tag.

I think the simplest long term solution would be to release multiarch docker images. This can be conveniently done in one command with BuildKit, `docker buildx build --platform linux/amd64,linux/arm64 .`.

.circleci/config.yml
```diff
@@ -209,6 +209,51 @@ jobs:
             docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
             make -C ./docker version=${CIRCLE_TAG} alias=latest push
 
+  publish-docker-image-arm64:
+    machine:
+      image: ubuntu-2004:current
```
sbc100, line 214: 2004?

sbc100: Oh I see that that means 20.04 not the year 2004!

.circleci/config.yml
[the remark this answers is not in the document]
sbc100: I see, thanks for the clarification.

.circleci/config.yml
```diff
@@ -209,6 +209,51 @@ jobs:
...
+            CHANGED_FILES=$(git diff --name-only <<pipeline.git.base_revision>>..<<pipeline.git.revision>>)
+            if [[ ! $CHANGED_FILES =~ "emscripten-releases-tags.json" ]]; then
+              echo "emscripten-releases-tags.json has not changed, skipping build and publish"
+              exit 0
+            fi
+      - run:
+          name: extract latest emscripten arm64 linux version
+          command: |
+            export ARM64_VERSION=$(grep -oP '(?<="latest-arm64-linux": ")[^"]*' emscripten-releases-tags.json)
+            export ARM64_IMAGE_TAG=${ARM64_VERSION}-arm64
```
sbc100, line 230: To these exported variables persist to the next `run` ?  My recollection was that they do not, and that you need to persist them somehow/somewhere?

sbc100 (approved):
lgm with a few comments

.circleci/config.yml
jamsinclair: You are correct! Thanks for pointing that out.

I've simulated it on my own CircleCI account to ensure that this job works as expected.

You can see the steps running at https://app.circleci.com/pipelines/github/jamsinclair/emsdk-circleci-test/8/workflows/8dffc43c-ef41-43b5-a973-c8778255612d/jobs/8

jamsinclair (commented):
Thanks for looking at the changes so swiftly @sbc100 – apologies I wasn't as quick to get back to your comments.

I've simulated the publish job on my own CircleCI account to ensure that this job works as expected. You can see the steps running at https://app.circleci.com/pipelines/github/jamsinclair/emsdk-circleci-test/8/workflows/8dffc43c-ef41-43b5-a973-c8778255612d/jobs/8

Feel free to tag me if there's any more changes required or ongoing support needed, thanks 🙇

PR #1239, opened 2023-06-23
Fix support fish shell in emsdk.py

shao4598:
Shell: Fish
OS: macOS
```shell
$ set -x EMSDK_QUIET 1
$ source ./emsdk_env.fish
```
fish: Unknown command: unset

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -114,15 +114,16 @@ else:
 UNIX = (MACOS or LINUX)
 
 
-# Pick which shell of 4 shells to use
+# Pick which shell of 5 shells to use
 POWERSHELL = bool(os.getenv('EMSDK_POWERSHELL'))
 CSH = bool(os.getenv('EMSDK_CSH'))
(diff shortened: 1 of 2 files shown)


emsdk.py
```diff
@@ -114,15 +114,16 @@ def exit_with_error(msg):
 UNIX = (MACOS or LINUX)
 
 
-# Pick which shell of 4 shells to use
+# Pick which shell of 5 shells to use
```
sbc100, line 117: Maybe just `# Pick which shell to use`?

sbc100 (approved):

emsdk.py
sbc100: I wonder if we should add some basic CI testing for other shells?

shao4598: 👌 I agree with you.

sbc100:
Any interesting in following up on this?  Or should we close it out?

PR #1248, opened 2023-06-27
bazel: add standalone defines during compile

walkingeyerobot:
specifically, define `STANDALONE_WASM` and `EMSCRIPTEN_STANDALONE_WASM` during compile actions for standalone mode

### DIFF
--- a/bazel/emscripten_toolchain/toolchain.bzl
+++ b/bazel/emscripten_toolchain/toolchain.bzl
@@ -1019,7 +1019,7 @@ def _impl(ctx):
             features = ["wasm_warnings_as_errors"],
         ),
         flag_set(
-            actions = ALL_COMPILE_ACTIONS,
+            actions = all_compile_actions,
             flags = ["-DSTANDALONE_WASM=1", "-DEMSCRIPTEN_STANDALONE_WASM=1"],
             features = ["wasm_standalone"],
         ),
--- a/bazel/emscripten_toolchain/toolchain.bzl
+++ b/bazel/emscripten_toolchain/toolchain.bzl
@@ -1019,6 +1019,11 @@ def _impl(ctx):
             features = ["wasm_warnings_as_errors"],
         ),
         flag_set(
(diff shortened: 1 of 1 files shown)


bazel/emscripten_toolchain/toolchain.bzl
```diff
@@ -1018,6 +1018,11 @@ def _impl(ctx):
             flags = ["-Werror"],
             features = ["wasm_warnings_as_errors"],
         ),
+        flag_set(
+            actions = ALL_COMPILE_ACTIONS,
+            flags = ["-DSTANDALONE_WASM=1", "-DEMSCRIPTEN_STANDALONE_WASM=1"],
+            features = ["wasm_standalone"],
+        ),
```
sbc100, line 1025: I wonder if project that want to build a special version of the code for standalone mode should just define this themselves?  This is what they would have to do in the non-bazel case where `-sSTANDALONE_WASM` is only a link-time flag.

walkingeyerobot: It's an ergonomic improvement. Setting defines in bazel is a bit unintuitive; many users end up setting defines for only a portion of their build instead of all of it like they would expect. Also, I believe most users who are setting `standalone = True` on `wasm_cc_binary` would either expect these to be defined and find them useful or not make use of them at all.

sbc100: It guess it maybe makes more sense for bazel to have different builds for different configurations, but in general I would hope that folks who what to build wasm libraries would just build a single library that could be used in both standalone and non-standalone mode, and they would not need to know at compile time.

We already have PIC and non-PIC and threads and non-threads builds, I would not want to encourage yet another axis on which to vary libraries.. but I guess maybe bazel hides all this so it matters less?

walkingeyerobot: What you're saying makes sense. I filed #19753. Once that's fixed this won't be necessary. I'll go ahead and delete this PR.

Issue #547, opened 2020-07-11
Set up community build bot for Linux/Aarch64 builds


bvibber:
Per recent email list discussion I've proposed setting up a community-run build bot for Linux/Aarch64 (ARM64) builds of LLVM + binaryen that can be downloaded by emsdk for direct installation on compatible machines.

(This is a niche platform today but we can expect it to become big as Apple is planning to convert their entire Mac line over to ARM64 CPUs in the next two years. Macs will eventually want native macOS/ARM64 builds too but Linux/ARM64 builds will also be used there in things like Docker or other virtualized Linux environments too, and it's easier to bootstrap now with easy availability of Linux server VMs)

I'm seeing a small regression in building with emsdk on ARM64 that I'll check over, then I'll see if I can build using the same tool that's used for the CI builds that emsdk downloads and make sure we've got something compatible.

kripken:
Talking to @brion , wasm2c may be able to help here. While we don't like the idea of end users needing to build C builds, if the main infra produces C builds then those could be built by @brion on an ARM machine, which as a single C file is much simpler than LLVM with the whole build system there. And then those builds would be uploaded, so end users get fully compiled builds, and wasm2c is just a step in the middle.

This will still have some downsides (like there is no simd or pthreads in wasm2c builds atm), so it would be best to get normal builds up. But I think it's worth seeing about wasm2c builds on the main bots - just one more build target, and it doesn't need a special machine to do so (the fast Linux machines can do it).

sbc100:
The problem not that that target machine can't build llvm from source (its full linux/arm installation).  He can build the whole SDK just fine.  The problem @brion wants to solve is having a pre-built set of binaries uploaded by the bots.  So it mostly about infa maintenance.  Converting to wasm2c doesn't seem to solve this specific problem AFAICT.

kripken:
@sbc100 See https://twitter.com/brionv/status/1283165113248133121 and earlier tweets right before it - my understanding is that there are some difficulties with building from source.

It would be much easier to just build a C file - `clang clang.c` and that's it.

sbc100:
Bit wait, I though the original discussion for this issue came of out request to have pre-built binaries because building from source was too slow?    Am I remembering wrong?

If the problem is "I can't build from source on arm linux", then that seems like something that should be solvable.  I'm sure llvm itself must be buildable on arm linux.

kripken:
[...]

PR #1256, opened 2023-07-10
Update linux arm64 to 3.1.41

sbc100:


sbc100:
See https://github.com/emscripten-core/emsdk/issues/547

PR #1284, opened 2023-10-08
Upgrade to Node.js 18.18.0

RobLoach:
This aims to update to [Node.js 18.18.0](https://nodejs.org/en/blog/release/v18.18.0), as discussed in #1173 . I'm not sure if it updates everything needed, but does update the manifest, as well all references. We'll need to do some aggressive testing to ensure it's correct.

### DIFF
--- a/bazel/emscripten_deps.bzl
+++ b/bazel/emscripten_deps.bzl
@@ -112,9 +112,8 @@ def emscripten_deps(emscripten_version = "latest"):
     # dependencies that use different emscripten versions
     excludes = native.existing_rules().keys()
     if "nodejs_toolchains" not in excludes:
-        # Node 16 is the first version that supports darwin_arm64
         node_repositories(
-            node_version = "16.6.2",
+            node_version = "18.18.0",
         )
 
     if "emscripten_bin_linux" not in excludes:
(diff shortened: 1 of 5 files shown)


sbc100:
Out of interest, is there some reason that you care about this mostly-internally-used version of node?   It shouldn't effect the version of node on which you can run the generated code.

Are you perhaps relying on the emsdk-provided version of node to run your generated code in some environment?  (testing? production?)

RobLoach:
Node.js 16 was EOLed last month: https://nodejs.dev/en/about/releases/

While it's not critical to update it, it would be good to push it forwards within emsdk.

sbc100:
> Node.js 16 was EOLed last month: https://nodejs.dev/en/about/releases/
> 
> While it's not critical to update it, it would be good to push it forwards within emsdk. 

emsdk does not depend on security updates or support for nodejs.  We use node internally as part of compiler process but it not exposed directly to end users or running untrusted code.   For this reason, I consider the end-of-life argument less relevant to our project than most.

The downside of dropping support for distros like Ubuntu/Bionic seems more real to me (node v18 doesn't support the version of glibc that ships in Ubuntu/Bionic: https://github.com/nodesource/distributions/issues/1392).

> Getting the following error since the install is not on Emscripten's release-build deps storage:
> 
> ```
> Installing tool 'node-18.18.0-64bit'..
> Error: Downloading URL 'https://storage.googleapis.com/webassembly/emscripten-releases-builds/deps/node-v18.18.0-linux-x64.tar.xz': HTTP Error 404: Not Found
> error: installation failed!
> ```

sbc100:
This was fixed in #1387

PR #1290, opened 2023-10-21
[README] fix fork emscripten docs

ericoporto:


ericoporto:
Actually, this is wrong too

PR #1295, opened 2023-10-29
[bazel] Add an :all target to binary repos

### LINKED ISSUE #1294: Export all files as @emscripten_bin_{platform}//:all
The `BUILD_FILE_CONTENT_TEMPLATE` in `bazel/emscripten_deps.bzl currently` currently exports a selection of files used by the CC toolchain that is configured here.

I'm facing a problem while creating Configure, Make & CMake toolchains using Emscripten via [rules_foreign_cc](https://github.com/bazelbuild/rules_foreign_cc), where I need access to other files from the `@emscripten_bin_linux` repository, namely `emconfigure[.py]`, `emmake[.py]` and `emcmake[.py]`. My current approach is to patch this repo (`@emsdk`) to modify the `BUILD_FILE_CONTENT_TEMPLATE` to export these files too.

Would you be accepting a pull request that simply adds an `:all` target to `@emscripten_bin_linux` and friends, which could then be used by toolchain developers to access other files from the binary repo? That would make my life easier.

attilaolah:
This is the simplest solution to my problem described in #1294.

I'm happy to change this to something more specific, e.g. by exporting only the files I currently need, but it seems like having an :all target is the most future-proof way here.

Closes #1294.

### DIFF
--- a/bazel/emscripten_deps.bzl
+++ b/bazel/emscripten_deps.bzl
@@ -9,6 +9,11 @@ BUILD_FILE_CONTENT_TEMPLATE = """
 package(default_visibility = ['//visibility:public'])
 
 filegroup(
+    name = "all",
+    srcs = glob(["**"]),
+)
+
+filegroup(
     name = "includes",
     srcs = glob([
         "emscripten/cache/sysroot/include/c++/v1/**",


walkingeyerobot (approved):

kripken:
Test error is not related and should not block this landing.

@tlively the error is in recent Lattice code in Binaryen,

```cpp
In file included from /root/project/binaryen/main/src/tools/wasm-fuzz-lattices.cpp:21:0:
/usr/include/c++/7/variant: In instantiation of 'constexpr const size_t std::variant_size_v<wasm::RandomFullLattice::L>':
/usr/include/c++/7/variant:702:10:   required from 'struct std::__detail::__variant::__gen_vtable<wasm::RandomFullLattice::ElementImpl, wasm::RandomFullLattice::getBottom() const::<lambda(const auto:5&)>&&, wasm::RandomFullLattice::L&>'
/usr/include/c++/7/variant:1255:23:   required from 'constexpr decltype(auto) std::visit(_Visitor&&, _Variants&& ...) [with _Visitor = wasm::RandomFullLattice::getBottom() const::<lambda(const auto:5&)>; _Variants = {wasm::RandomFullLattice::L&}]'
/root/project/binaryen/main/src/tools/wasm-fuzz-lattices.cpp:666:29:   required from here
/usr/include/c++/7/variant:97:29: error: incomplete type 'std::variant_size<wasm::RandomFullLattice::L>' used in nested name specifier
     inline constexpr size_t variant_size_v = variant_size<_Variant>::value;
                             ^~~~~~~~~~~~~~
```

PR #1326, opened 2023-12-27
Fix FileExistsError on subsequent bazel builds on Windows

### LINKED ISSUE #1181: [BUILD FAILURE] Bazel builds are always failing! emsdk (3.1.31)
Hi,

I am having an issue with Bazel and emscripten.
I simply want to link a static library which builds fine natively (using cc_binary directly) but always fails with wasm_cc_binary:
I am using the latest version of emsdk (3.1.31)

I tried to remove all the bazel folders and even started from scratch using the hello world example (https://github.com/emscripten-core/emsdk/tree/main/bazel/hello-world).

As soon as I add `deps` to `cc_binary` the emscripten build fails!

### The error

```sh
# Execution platform: @local_config_platform//:host
  File "C:\Users\admin\_bazel_admin\m6eps4yl\execroot\hello_world\external\emsdk\emscripten_toolchain\link_wrapper.py", line 64, in <module>
    os.rename(output_file, output_file + '.' + oformat)
FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'bazel-out/wasm-opt-ST-b4dbd4a2bfc4/bin/hello-world/hello-world' -> 'bazel-out/wasm-opt-ST-b4dbd4a2bfc4/bin/hello-world/hello-world.js'
```

### Files

```sh
/hello-world
    /BUILD
    /hello-world.cc
/lib
    /localtime
        /localtime.cpp
        /localtime.h
        /BUILD
```

### My BUILD file (/hello-world/BUILD): 

```c
load("@rules_cc//cc:defs.bzl", "cc_binary")
load("@emsdk//emscripten_toolchain:wasm_rules.bzl", "wasm_cc_binary")

cc_binary(
    name = "hello-world",
    srcs = ["hello-world.cc"],

    # ISSUE HERE... "Cannot create a file when that file already exists" when building
    deps = [
        "//lib/localtime:localtime",
    ],
)

wasm_cc_binary(
    name = "hello-world-wasm",
    cc_target = ":hello-world",
)
```

### My main (/hello-world/hello-world.cc): 

```cpp
#include <iostream>

#include "lib/localtime/localtime.h"

int main(int argc, char** argv) {
  LocalTime::Now();
  std::cout << "hello world!" << std::endl;
  return 0;
}
```

### /lib/localtime/localtime.h
```cpp
#pragma once

namespace HelloWorld {

    class LocalTime {
        public:

        /*
        * Prints the current time to stdout
        */
        static void Now();
    };
}

```

### /lib/localtime/localtime.cpp
```cpp
#include "localtime.h"

#include <ctime>
#include <stdio.h>

namespace HelloWorld  {

    void LocalTime::Now() {
        std::time_t result = std::time(nullptr);
        printf("%s", std::asctime(std::localtime(&result)));
    }

}
```

### /lib/localtime/BUILD
```cpp
load("@rules_cc//cc:defs.bzl", "cc_library")

package(default_visibility = ["//visibility:public"])

cc_library(
    name = "localtime",
    srcs = ["localtime.cpp"],
    hdrs = ["localtime.h"]
)
```

Is this a bug or am I doing something wrong? Again this works great when compiling natively WITHOUT emscripten

NovaSagittarii:
Fixes #1181. `os.rename()` has different behaviors on Windows vs Linux, using `os.replace()` instead fixes the
issue when a new build is created when previous files exist from an older build.

I'm not sure if this is the best way to do this though since there's the method `move_with_overwrite()` in emsdk.py which does what `os.replace()` is supposed to do. Regarding atomicity, according to [python documentation](https://docs.python.org/3/library/os.html#os.replace), `os.replace()`'s renaming is atomic as a POSIX requirement, but Windows isn't fully POSIX compliant.

https://github.com/emscripten-core/emsdk/blob/8822664a145391eee7cb57f9db7ba3705e19f2fe/emsdk.py#L547-L553

### DIFF
--- a/bazel/emscripten_toolchain/link_wrapper.py
+++ b/bazel/emscripten_toolchain/link_wrapper.py
@@ -61,7 +61,7 @@ if oformat is not None:
 
   # If the output name has no extension, give it the appropriate extension.
   if not base_name_split[1]:
-    os.rename(output_file, output_file + '.' + oformat)
+    os.replace(output_file, output_file + '.' + oformat)
 
   # If the output name does have an extension and it matches the output format,
   # change the base_name so it doesn't have an extension.
@@ -77,7 +77,7 @@ if oformat is not None:
   # Please don't do that.
   else:
     base_name = base_name_split[0]
-    os.rename(output_file, os.path.join(outdir, base_name + '.' + oformat))
+    os.replace(output_file, os.path.join(outdir, base_name + '.' + oformat))
 
 files = []
 extensions = [
@@ -161,6 +161,6 @@ if not len(files):
 # cc_binary must output exactly one file; put all the output files in a tarball.
 cmd = ['tar', 'cf', 'tmp.tar'] + files
 subprocess.check_call(cmd, cwd=outdir)
-os.rename(os.path.join(outdir, 'tmp.tar'), output_file)
+os.replace(os.path.join(outdir, 'tmp.tar'), output_file)
 
 sys.exit(0)


walkingeyerobot (approved):
thanks very much!

PR #1329, opened 2024-01-03
Restore npm install step removed in #1325

sbc100:


kripken (approved):
Do we not have tests for this?

sbc100:
> Do we not have tests for this?

I guess not.. I'm not even sure these install targets are even still supported at all.  I will followup and see if we can remove them maybe.

PR #1333, opened 2024-01-20
[bazel] Use hermetic Python

kon72:
This change removes the dependency on system Python.

Also fixes an issue where the build fails on Windows with `--incompatible_strict_action_env` flag specified:

```
'py' is not recognized as an internal or external command,
```


bazel/emscripten_toolchain/toolchain.bzl
```diff
@@ -1040,6 +1040,10 @@ def _impl(ctx):
                     key = "EM_CONFIG_PATH",
                     value = ctx.file.em_config.path,
                 ),
+                env_entry(
```
sbc100, line 1043: Why not just set `EMSDK_PYTHON` set here and avoid re-exporting it in env.sh?

sbc100 (commented):
Nice!

bazel/emscripten_toolchain/toolchain.bzl
kon72: Done!

walkingeyerobot (approved):
Thanks very much!

kalmard0:
Would changing the CI user to non-root be an acceptable solution here?

allsey87:
> If you wanted to send a PR that updates the bazel version, I'd be happy to accept it. The ubuntu image thing would be more of a question for @sbc100 I think.

I'll put it together tomorrow. Bazel 5.4.0 (used by CircleCI on Windows) is not covered by [`rules_python`'s support policy](https://rules-python.readthedocs.io/en/latest/support.html#supported-bazel-versions).

@sbc100 @walkingeyerobot could you two decide on what the minimum supported version of Bazel should be for emsdk? This [support matrix](https://bazel.build/release#support-matrix) might help.

logankaser:
> Let's bump it to bazel 7 then. Seems to be the oldest version that's still under active support.

Both 6.5.0 and 7.5.0 are supported until Dec 2025: https://bazel.build/release#support-matrix

sbc100:
Is this change still needed/useful?

@DoDoENT ?

DoDoENT:
> Is this change still needed/useful?
> 
> 
> 
> @DoDoENT ?

This is implemented, AFAIK. I added support for this half a year ago when system python on mac stopped being supported. 

And the current implementation is more modern and uses bzlmod. 

I think this one can be closed as completed.

PR #1334, opened 2024-01-24
Only pass -DLLVM_TEMPORARILY_ALLOW_OLD_TOOLCHAIN=ON when building llvm. NFC

sbc100:


dschuff (approved):
LGTM but is this still needed? LLVM has required a C++17 toolchain for a long time, and I would be surprised if it builds with a really old toolchain

sbc100:
> LGTM but is this still needed? LLVM has required a C++17 toolchain for a long time, and I would be surprised if it builds with a really old toolchain

I don't use this code path enough to know.. so for now I'll just land this as is.

PR #1345, opened 2024-02-16
Add activated_cfg for ccache

nornagon:
With this missing, ccache was never being added to PATH, and _EMCC_CCACHE wasn't being set.

### DIFF
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -633,6 +633,7 @@
     "bitness": 64,
     "url": "https://github.com/juj/ccache.git",
     "git_branch": "emscripten",
+    "activated_cfg": "CCACHE=%installation_dir%/bin",
     "activated_path": "%installation_dir%/bin",
     "activated_env": "_EMCC_CCACHE=1;CCACHE_CONFIGPATH=%installation_dir%/emcc_ccache.conf",
     "cmake_build_type": "Release",


emsdk_manifest.json
```diff
@@ -633,6 +633,7 @@
     "bitness": 64,
     "url": "https://github.com/juj/ccache.git",
     "git_branch": "emscripten",
+    "activated_cfg": "CCACHE=%installation_dir%/bin",
```
sbc100, line 636: CCACHE is not a recognized setting the emscripten config file as far as I can tell.

The valid config settings are listed here: https://github.com/emscripten-core/emscripten/blob/7cdfa8d9f22562e7c08789a5ae29ad9a1025a8e4/tools/config.py#L122-L140  (not the `CACHE` is unrelated to ccache).

emsdk_manifest.json
```diff
@@ -633,6 +633,7 @@
...
     "activated_path": "%installation_dir%/bin",
```
sbc100, line 637: This line should add ccache to your PATH.  You would need to be sure to include it in your `emsdk activate` command line though I think.

emsdk_manifest.json
nornagon: Ah, should I add a CCACHE setting there then? FWIW, when I made this change locally, it worked fine without adding it to the config.py list.

emsdk_manifest.json
nornagon: Without `activated_cfg`, this clause causes ccache to not be added to the path:

https://github.com/emscripten-core/emsdk/blob/c18280c8f3f8f088b4337f0f6a5fdb2be0d67179/emsdk.py#L1733-L1735

sbc100: `activated_cfg` doesn't have anything to do with the PATH.   `activated_cfg` is used to add things to the emscripten config file.  As far as I know ccache has not place in the emscripten config file.

The things that adds to your PATH is `activated_path`.

nornagon: Okay, then it would seem we need to change L1733-L1735 here, as this results in any tool without an activated_cfg and no deps being excluded. ccache is the only tool without an activated_cfg as far as I can tell.

nornagon: I _think_ if we change L1735 to `return True` then this will result in ccache always being active if it's installed? I'm not super familiar with how emsdk works—where is the list of currently activated tools stored?

sbc100: Yes, it looks like we can completely remove those lines I think.

I'm not sure why would declare a tool inactive simply because it has no dependencies.

sbc100: @juj any idea why these lines exist?

sbc100:
My hope is that #1766 will remove the need for this

sbc100:
Closing in favor of #1766

PR #1354, opened 2024-03-13
Remove JS_ENGINES and COMPILER_ENGINE from config file

sbc100:
`COMPILER_ENGINE` was completely removed from emscripten in https://github.com/emscripten-core/emscripten/pull/9469.

`JS_ENGINES` is only used/needed for testing and is no longer needed as of https://github.com/emscripten-core/emscripten/pull/9542.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1464,6 +1464,21 @@ def load_em_config():
       pass
 
 
+def find_emscripten_root(active_tools):
+  """Find the currently active emscripten root.
+
+  If there is more than one tool that defines EMSCRIPTEN_ROOT (this
+  should not happen under normal circumstances), assume the last one takes
+  precedence.
+  """
+  root = None
+  for tool in active_tools:
+    config = tool.activated_config()
+    if 'EMSCRIPTEN_ROOT' in config:
+      root = config['EMSCRIPTEN_ROOT']
+  return root
+
+
 def generate_em_config(active_tools, permanently_activate, system):
   cfg = 'import os\n'
   cfg += "emsdk_path = os.path.dirname(os.getenv('EM_CONFIG')).replace('\\\\', '/')\n"
@@ -1484,10 +1499,17 @@ def generate_em_config(active_tools, permanently_activate, system):
   for name, value in activated_config.items():
     cfg += name + " = '" + value + "'\n"
 
-  cfg += '''\
-COMPILER_ENGINE = NODE_JS
-JS_ENGINES = [NODE_JS]
-'''
+  emroot = find_emscripten_root(active_tools)
+  if emroot:
+    version = parse_emscripten_version(emroot)
+    # Older emscripten versions of emscripten depend on certain config
+    # keys that are no longer used.
+    # See https://github.com/emscripten-core/emscripten/pull/9469
+    if version < [1, 38, 46]:
+      cfg += 'COMPILER_ENGINE = NODE_JS\n'
+    # See https://github.com/emscripten-core/emscripten/pull/9542
+    if version < [1, 38, 48]:
+      cfg += 'JS_ENGINES = [NODE_JS]\n'
 
   cfg = cfg.replace("'" + EMSDK_PATH, "emsdk_path + '")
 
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -1484,10 +1484,18 @@ def generate_em_config(active_tools, permanently_activate, system):
...
+    if 'EMSCRIPTEN_ROOT' in config:
+      version = parse_emscripten_version(config['EMSCRIPTEN_ROOT'])
+      # Older emscripten versions of emscripten depend on certain config
+      # keys that are no longer used.
+      # See https://github.com/emscripten-core/emscripten/pull/9469
+      if version < [1, 38, 46]:
+        cfg += 'COMPILER_ENGINE = NODE_JS\n'
+      # See https://github.com/emscripten-core/emscripten/pull/9542
+      if version < [1, 38, 48]:
+        cfg += 'JS_ENGINES = [NODE_JS]\n'
```
kripken, line 1498: This is in a loop - is it not possible that we get these more than once? Or do we guarantee that `EMSCRIPTEN_ROOT` appears exactly once in a single activated config?

kripken (commented):
lgtm otherwise, versions look right.

emsdk.py
sbc100: I suppose we could have more than one version of emscripten installed?    But in that case the desired behaviour is that the config file should be a super-set of all the settings required I think.

The exact same pattern is used here: https://github.com/emscripten-core/emsdk/blob/7815dcaa5c97ee387e925137a42dd4219e7cec82/emsdk.py#L2490-L2494

kripken: The superset makes sense, but I think the current code can end up with duplication? That is,
```python
JS_ENGINES = [NODE_JS]
JS_ENGINES = [NODE_JS]
JS_ENGINES = [NODE_JS]
```

sbc100: But I don't think it matters because you can't really have more than one version of emscripten installed can you?

Also, those repeated keys would be harmless if you somehow did mange to activate more than one emscripten install at once

kripken: > But I don't think it matters because you can't really have more than one version of emscripten installed can you?

Yeah, I think only one can be active at once, but just wanted to check that my understanding was the same as yours. However, reading the code I can't easily confirm that, as in the manifest file `EMSCRIPTEN_ROOT` appears in things with id "releases" and also id "emscripten". I think a release contains emscripten among other things, so that suggests we use that field in more than one active thing?

> Also, those repeated keys would be harmless if you somehow did mange to activate more than one emscripten install at once

I agree, it should not break anything, but would just look weird.

sbc100: What solution would your propose here?

Remember this only effects very old version of emscripten.  So in order to see this "wierd" duplication one would need to:

1. Somehow install more than one version of emscripten (I'm pretty sure this is not support and/or would never work)
2. Both version of emscripten would need to be these very old versions.

kripken: Perhaps we could in the loop set a boolean "need legacy vars", and then outside of it if that is set then we add those two? That doesn't seem much more complex, and it seems safer to me.

sbc100: But there is no safety issue so I'm not sure adding complexity here is worth it.  Have those duplicate keys would be a very minor cosmetic issue (more folks are not going to look at this config file ever).

sbc100: I refactored and added `find_emscripten_root()` which returns the currently active emscripten root.

emsdk.py
```diff
@@ -1463,6 +1463,19 @@ def load_em_config():
     except:
       pass
 
+def find_emscripten_root(active_tools):
+  """Find the currnetly active emscripten root.
+
+  If there is more than one tool that defines EMSCRIPTEN_ROOT (this
+  should not happen under normal circumstances), assume the last one takes
+  precidence.
```
kripken, line 1471:
```suggestion
  precedence.
```

emsdk.py
```diff
@@ -1463,6 +1463,19 @@ def load_em_config():
...
+  """
+  for tool in active_tools:
+    config = tool.activated_config()
```
kripken, line 1474:
```suggestion
    root = None
    config = tool.activated_config()
```
I think without that Python will error if none is found (the `return root` will say it is read before written).

kripken (commented):
lgtm % comments

kripken (approved):

PR #1359, opened 2024-03-14
CI: Build `latest` docker image rather than `tot`

sbc100:
This means that CI run that update `latest` actually test the thing we are about to ship.

We recently had a case where `latest` was broken but `tot` was fixes and we accidentally shipped a broken SDK version (#1353).


sbc100:
I expect the CI for this to fail because `latest` is currently broken, however I think we should land this anyway so that when we update `latest` we can see that it gets fixed.

aheejin (approved):

PR #1363, opened 2024-03-29
(bazel) Set @platforms//os:emscripten for platform_wasm

trzeciak:
I noticed that @googlewalt added `@platforms//os:emscripten` [LINK](https://github.com/bazelbuild/platforms/commit/6b04b816a0edc09e1a3ad4a094177e181a2a1803), now we can update the `platform_wasm`.

### DIFF
--- a/bazel/MODULE.bazel
+++ b/bazel/MODULE.bazel
@@ -1,0 +1,1 @@
+bazel_dep(name = "platforms", version = "0.0.9")
--- a/bazel/README.md
+++ b/bazel/README.md
@@ -26,6 +26,11 @@ The SHA1 hash in the above `strip_prefix` and `url` parameters correspond to the
 newer versions, you'll need to update those. To make use of older versions, change the
 parameter of `emsdk_emscripten_deps()`. Supported versions are listed in `revisions.bzl`
 
+Bazel 7+ additionally requires `platforms` dependencies in the `MODULE.bazel` file.
+```starlark
+bazel_dep(name = "platforms", version = "0.0.9")
+```
+
 
 ## Building
 
--- a/bazel/BUILD
+++ b/bazel/BUILD
@@ -82,6 +82,7 @@ platform(
     name = "platform_wasm",
     constraint_values = [
         "@platforms//cpu:wasm32",
+        "@platforms//os:emscripten",
     ],
 )
 
--- a/bazel/deps.bzl
+++ b/bazel/deps.bzl
@@ -4,6 +4,15 @@ load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 def deps():
     maybe(
         http_archive,
+        name = "platforms",
+        urls = [
+            "https://mirror.bazel.build/github.com/bazelbuild/platforms/releases/download/0.0.9/platforms-0.0.9.tar.gz",
+            "https://github.com/bazelbuild/platforms/releases/download/0.0.9/platforms-0.0.9.tar.gz",
+        ],
+        sha256 = "5eda539c841265031c2f82d8ae7a3a6490bd62176e0c038fc469eabf91f6149b",
+    )
+    maybe(
+        http_archive,
         name = "bazel_skylib",
         sha256 = "c6966ec828da198c5d9adbaa94c05e3a1c7f21bd012a0b29ba8ddbccb2c93b0d",
         urls = [
--- a/bazel/test_external/MODULE.bazel
+++ b/bazel/test_external/MODULE.bazel
@@ -1,0 +1,1 @@
+bazel_dep(name = "platforms", version = "0.0.9")


trzeciak:
This change require use platforms in version 0.0.9+, in my own project, I set it in `MODULE.bazel`:
```
bazel_dep(name = "platforms", version = "0.0.9")
```

walkingeyerobot (approved):
Thanks very much!

sbc100:
CI is reporting `ERROR: /root/project/bazel/BUILD:81:9: no such target '@@platforms//os:emscripten': target 'emscripten' not declared in package 'os' defined by /root/.cache/bazel/_bazel_root/04d6701d4e894b3c38d7a891fc3b3ded/external/platforms/os/BUILD and referenced by '//:platform_wasm'`

I guess something else needs updating?

trzeciak:
If I understand the operation of bazel correctly, bazel has a built-in dependency for `platforms` (and several others), but it can be changed with an entry in the `MODULE` or `WORKSPACE` file.  
I don't know when platform@0.0.9 will become built-in to bazel.  
Therefore, I don't know what to do next with this MR so as not to spoil the work of others.

fmeum:
Since essentially everything depends on `platforms` and the first definition in WORKSPACE wins, most WORKSPACE users won't be on a `platforms` version with this target. However, most users will know that the error they get can be solved by updating their dep on `platforms` - simply because this breaks all the time with WORKSPACE and even a Bazel update doesn't help if something else brings in an outdated version.

With Bzlmod, bumping the `bazel_dep` version is all you need.

All in all, I would say go for the update.

sbc100:
I don't think there is a question as to whether we want to upgrade, but more of question of how to fix the CI tests which are failing.  We can't land this until we fix those.

fmeum:
Could you try adding
```starlark
maybe(
    http_archive,
    name = "platforms",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/platforms/releases/download/0.0.9/platforms-0.0.9.tar.gz",
        "https://github.com/bazelbuild/platforms/releases/download/0.0.9/platforms-0.0.9.tar.gz",
    ],
    sha256 = "5eda539c841265031c2f82d8ae7a3a6490bd62176e0c038fc469eabf91f6149b",
)
```
 to [this macro](https://github.com/emscripten-core/emsdk/blob/main/bazel/deps.bzl)?

trzeciak:
Oooo, this looks very promising, I will try it locally first.

trzeciak:
@fmeum Ok, I test it locally on macOS, and after add your snippet to deps macro, I tested it for few different version of bazel (using bazelisk):
```
# USE_BAZEL_VERSION=5.4.0  # OK
# USE_BAZEL_VERSION=6.4.0  # OK
# USE_BAZEL_VERSION=6.5.0  # OK
# USE_BAZEL_VERSION=7.0.0  # ERR
# USE_BAZEL_VERSION=7.0.2  # ERR
# USE_BAZEL_VERSION=7.1.1  # ERR
```
and it seems that not work for 7+ version.

I also followed the `circleci` logs as @walkingeyerobot mentioned, and now I see that on CI was used latest bazel version (7.1.1) on linux and macOS, and and for some reason 5.4.0 on windows.

Following this line of reasoning, I suppose I can add platform@0.0.9 dependency in a new MODULE file in this repo against to WORKSPACE (I will try this way).

By the way, it makes sense that this doesn't work in Bazel 7, because the maybe method is described as adding a dependency if it doesn't exist at all:
```py
def maybe(repo_rule, name, **kwargs):
    """Utility function for only adding a repository if it's not already present.
    …
```

Going further, it might be nice to update emsdk to fully support bzlmod, but maybe not in this MR.

bazel/README.md
```diff
@@ -26,6 +26,11 @@ The SHA1 hash in the above `strip_prefix` and `url` parameters correspond to the
 newer versions, you'll need to update those. To make use of older versions, change the
 parameter of `emsdk_emscripten_deps()`. Supported versions are listed in `revisions.bzl`
 
+Bazel 7+ additionally requires `platforms` dependencies in the `MODULE.bazel` file.
+```starlark
+bazel_dep(name = "platforms", version = "0.0.9")
+```
+
```
fmeum, line 33:
```suggestion
```

This isn't needed: Bzlmod resolves transitive dependencies via Minimum Version Selection, so adding the `platforms` dep to the emsdk's `MODULE.bazel` file is sufficient for dependent modules to pick it up.

bazel/MODULE.bazel
```diff
@@ -0,0 +1 @@
+bazel_dep(name = "platforms", version = "0.0.9")
```
fmeum, line 1: To properly Bzlmod-ify this so that other Bazel projects can depend on it, you also need to add something like this to the top:

```starlark
module(
    name = "emsdk",
    version = "<current version>",
)
```

Instead of manually maintaining `<current version>`, you can also use the [Publish to BCR app](https://github.com/bazel-contrib/publish-to-bcr) to automate the version bump and releases.

trzeciak: I hope I understand what you mean xD (I'm still new in bazel).  
If it's not a problem, I'll try to do it in the next/separate MR.  
Thanks for the tip!

bazel/README.md
fmeum: I see, this snippet is meant for users pulling in `emsdk` via WORKSPACE but who do have `--enable_bzlmod` set, e.g. because it is the default in Bazel 7+? That makes sense if you don't want to fully set up `emsdk` for Bzlmod.

If you do, then this wouldn't be necessary and such users could just load `emsdk` via a `bazel_dep` in MODULE.bazel instead. But that could be done in a separate PR.

PR #1383, opened 2024-05-15
[bazel] add tarball as an optional output

### LINKED ISSUE #1382: Version of `wasm_cc_binary` that doesn't untar the result?
I'm working on updating our build pipeline to use the upstream version of the emscripten bazel toolchain (our current implementation of using Bazel to drive emscripten predates having upstream support).

Our current build pipeline publishes the `*.tar` file produced by emscripten and downloads it elsewhere in the build pipeline. After changing our pipeline to use `wasm_cc_binary`, which is nice because of the platform transition it does, the output is now a multitude of files, which we have to tar back up again.

It would be great if there were an option or another function I could call that only did the transition, and didn't extract the tarball.

walkingeyerobot:
Usage would be something like:
```
filegroup(
    name = "hello-world-tarball",
    srcs = [":hello-world-wasm"],
    output_group = "_wasm_tar",
)
```

Fixes #1382

### DIFF
--- a/bazel/emscripten_toolchain/wasm_cc_binary.bzl
+++ b/bazel/emscripten_toolchain/wasm_cc_binary.bzl
@@ -128,12 +128,15 @@ def _wasm_cc_binary_impl(ctx):
         executable = ctx.executable._wasm_binary_extractor,
     )
 
-    return DefaultInfo(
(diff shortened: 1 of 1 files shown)


sbc100 (approved):
Rubberstamp lgtm

PR #1388, opened 2024-05-27
bazel: bump rules_nodejs and migrate to rules_js

linzhiq:
Bazel's Node.js dependency comes from [rules_nodejs](https://github.com/bazelbuild/rules_nodejs/). Previously, bazel/deps.bzl was using rules_nodejs 5.8.0, released in 2022 and only supported Node.js toolchains up to 18.12.1. 

This PR bumps rules_nodejs to latest 6.1.1. It also replaces build_bazel_rules_nodejs with [rules_js](https://github.com/aspect-build/rules_js), since `npm_install` that bazel/emscripten_deps.bzl used was deprecated. The README of rules_nodejs now recommends migrating to rules_js for everything other than the Node.js toolchain: (https://github.com/bazelbuild/rules_nodejs/commit/371e8cab15f9db444217703144b2eb71d0e13235)

#### Impetus

Our repo builds with Bazel and uses Emscripten and Node.js. Tried to upgrade Node.js 18 to Node.js 20 and saw that emsdk didn't support rules_nodejs 6+ in the same workspace.

### DIFF
--- a/bazel/deps.bzl
+++ b/bazel/deps.bzl
@@ -23,12 +23,12 @@ def deps():
     maybe(
         http_archive,
         name = "rules_nodejs",
-        sha256 = "08337d4fffc78f7fe648a93be12ea2fc4e8eb9795a4e6aa48595b66b34555626",
-        urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.8.0/rules_nodejs-core-5.8.0.tar.gz"],
+        sha256 = "3e8369256ad63197959d2253c473a9dcc57c2841d176190e59b91d25d4fe9e67",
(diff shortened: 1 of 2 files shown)


keith:
I think we probably need to fix the CI here first

allsey87:
@walkingeyerobot I would also like to upgrade rules_nodejs and nodejs since embuilder is complaining about the current version used in the Bazel workspace and this might be the source of the errors that I am seeing in #1405
```
embuilder: warning: node version appears too old (seeing "v16.6.2", expected "v16.20.0") [-Wversion-check]
```
Should I take over this PR or perhaps open a new one? I don't think I can push commits here unless I am a "contributor" to this repository, right?

sbc100:
It seems that in the interm we have already updated to rules_nodejs 6.7.3, so I think we can close this?

PR #1389, opened 2024-05-28
[Bazel] Add feature to support WASM Relaxed SIMD

yolanda15:
[WASM Relaxed SIMD](https://github.com/WebAssembly/relaxed-simd/blob/main/proposals/relaxed-simd/Overview.md) has been supported by LLVM and used in the XNNPACK acceleration library in bazel build, see [reference](https://github.com/google/XNNPACK/blob/master/build_config/BUILD.bazel#L169).

Add this feature in bazel toolchain so that it can be recoginized in config similarly to the WASM SIMD feature.


kripken (approved):
Looks good on the wasm backend side (LLVM flag looks correct).

walkingeyerobot (approved):

Issue #1401, opened 2024-06-11
[Bazel] Dynamically generate Emscripten cache


allsey87:
I am investigating a solution for solving the problem of the frozen cache + wasm64, lto, and pic once and for all. The idea is to pass both the settings and the ports that you want built to `emscripten_deps` which will generate the appropriate genrules and pull the generated files into the correct file groups so that everything is made available once during toolchain set up.

To make this work, I would change `emscripten_deps` to take a two addition arguments as follows with defaults set up for backwards compatibility:
```python
emscripten_deps(
   emscripten_version = "latest",
   features = ["wasm32"], # lto, lto-thin, pic, wasm64
   targets = ["crtbegin", ...]
)
```
The idea would be to then loop over the targets, creating a genrule for each and calling embuilder to make the asset available in the cache. The string of all genrules would then be injected into `BUILD_FILE_CONTENT_TEMPLATE`. I started work on a rough prototype but before taking it further, **please let me know if there are any reservations that would prevent this getting merged**.

At the moment, my rough prototype sits external to `@emscripten_bin_linux//` but the idea will be to move this code inside it so that I can tweak what is available in the cache.

```python
genrule(
    name = "test-embuilder",
    tools = [
        "@emscripten_bin_linux//:emscripten_config_upstream",
        "@emscripten_bin_linux//:emscripten/embuilder.py",
        "@emscripten_bin_linux//:compiler_files",
        "@emscripten_bin_linux//:linker_files",
        "@emscripten_bin_linux//:ar_files",
    ],
    cmd = """
echo "import os" >> embuilder_config
echo "CACHE = '$$(realpath $(RULEDIR))'"
echo "BINARYEN_ROOT = '$$(realpath $$(dirname $(location @emscripten_bin_linux//:emscripten_config_upstream)))'" >> embuilder_config
echo "LLVM_ROOT = os.path.join(BINARYEN_ROOT, 'bin')" >> embuilder_config
echo "EMSCRIPTEN_ROOT = os.path.join(BINARYEN_ROOT, 'emscripten')" >> embuilder_config

$(location @emscripten_bin_linux//:emscripten/embuilder.py) \
    --em-config embuilder_config \
    --pic \
    build crtbegin
""",
    # This rule crashes here at the moment since crtbegin.o is not created in RULEDIR and embuilder seems to be ignoring
    # the CACHE environment variable
    outs = ["crtbegin.o"]
)
```

sbc100:
[...]

Issue #1362, opened 2024-03-28
compile wams64 using bazel mode


phoenix20162016:
in default bazel config， the build env is under 32bit mode, how can i switch to 64bit mode

sbc100:
To build for wasm64 you can add `-sMEMORY64` to you compile and link commands.

phoenix20162016:
![image](https://github.com/emscripten-core/emsdk/assets/18574396/d83dfb70-8dc1-4dd2-aecc-5fe560a11d78)
follows your suggestion， another error occurs, the following is my bazel configuration
![image](https://github.com/emscripten-core/emsdk/assets/18574396/b463551f-0870-4a7b-b992-387eab79cf5c)

in the fold of the compile tools, the wasm64-emscripten dir is really  not exist

sbc100:
I see, yes we don't currently wasm64 versions of any core libraries in the bazel toolchain.    I think this is the same as  https://github.com/emscripten-core/emsdk/issues/807.    Perhaps we can close this and continue the discussion there?

wzheng21:
@phoenix20162016 I had the same problem as yours. For whoever having similar issues, I had to patch emsdk setting `FROZEN_CACHE = False` in config.py, then the issue becomes `read-only blabla` error, basically indicating emscripten tries to compile libraries and writes into a temporary cache folder but failed.

This of course would fail, due to bazel's idea of keeping builds hermitian. The following are the two different workarounds I tested with my work:

1. Disable bazel sandbox mode. Just specify `--spawn_strategy=local`. Be careful when using this, especially if this is not a toy project
2. Don't play with sandbox, but specify `--action_env=EM_CACHE=~/.cache/emscripten_cache`. When FROZEN_CACHE is False, emscripten will try to write to `~/.emscripten_cache`, which will definitely fail in bazel sandbox mode due to permissions. Writing to `~/.cache` folder, however, is allowed.

Hope this is helpful for anyone getting stuck with this issue.

Issue #1099, opened 2022-09-05
Bazel: @emsdk cannot pick up SDL2 headers and libraries.


sohanjg:
Bazel build with emsdk (for wasm_cc_binary) is unable to fetch and discover the SDL2 file on specifying them in linkopts ("-sUSE_SDL=2", "-sUSE_SDL_IMAGE=2", "-sUSE_SDL_TTF=2", "-sUSE_SDL_NET=2") and coopts ("-DUSE_SDL=2", "-DUSE_SDL_IMAGE=2",  "-DUSE_SDL_TTF=2", "-DUSE_SDL_NET=2",) as mentioned in https://emscripten.org/docs/compiling/Building-Projects.html?#emscripten-ports 
Compilation fails with "SDL2/SDL.h" headers not found errors. 

Note : SDL1 headers and libs/API, work out of the box. We are experimenting in Skia bazel environment.

sbc100:
The `-s` options are needed at compile time too.    For example, you should use `-sUSE_SDL=2` and not `-DUSE_SDL2`

sohanjg:
Thanks ! Now it fails with some permission error, about RO file system. I am on Ubuntu 22.04 with sudo access.
Pasting the build log.

`Starting local Bazel server and connecting to it...
INFO: Analyzed target //example:hello_world_wasm (133 packages loaded, 12128 targets configured).
INFO: Found 1 target...
ERROR: /home/ubuntu/skia/tools/timer/BUILD.bazel:7:16: Compiling tools/timer/Timer.cpp failed: (Exit 1): emcc.sh failed: error executing command external/emsdk/emscripten_toolchain/emcc.sh '--sysroot=external/emscripten_bin_linux/emscripten/cache/sysroot' -fdiagnostics-color -fno-exceptions -fno-strict-aliasing -funsigned-char ... (remaining 155 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
Traceback (most recent call last):
  File "/home/ubuntu/.cache/bazel/_bazel_ubuntu/756dee633784ef959e49802380c6f32b/sandbox/linux-sandbox/4/execroot/skia/external/emscripten_bin_linux/emscripten/emcc.py", line 4010, in <module>
    sys.exit(main(sys.argv))
  File "/usr/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
[...]

PR #1402, opened 2024-06-12
[Bazel] Dynamically generate the Emscripten cache

allsey87:
This is a draft PR for sharing my progress on this feature. The idea is to use embuilder to generate the Emscripten cache dynamically by specifying what you want upfront and in which configuration (i.e., wasm64, lto, pic etc).

For now I would really appreciate any input on these remaining problems:

1. **Output names**
Some libraries generate `.a` and some `.o`, I suspect that it is just the c runtime libraries that emit the `.o` files, so I could tell Bazel that if the name of the library begins with crt then its output is `.o`?

2. **Overlaying the cache**
I added my output to `link_files` but this seems to be ignored by the Emscripten linker. When I compile a program, Emscripten considers the cache to be: `external/emscripten_bin_linux/emscripten/cache/` while, the cache assets from my genrule are actually stored at: `execroot/_main/bazel-out/k8-opt-exec-ST-13d3ddad9198/bin/external/emscripten_bin_linux/emscripten/cache/`. I am not sure what the best approach for moving forward here is.

3. **Portability**
I haven't found away to get rid of the `dirname` and `realpath` for the moment. It seems almost impossible to provide a rule in Bazel with a path which means it is very difficult to set the values of `BINARYEN_ROOT` etc for `embuilder`. The best solution I can think of for the moment is to leave the Bash-isms in place and just use `cmd_bash` and `cmd_bat` (provided by genrule) to set the environment variables.

Related issues:
- #807
- #971
- #1099
- #1362
- #1401

### DIFF
--- a/bazel/emscripten_deps.bzl
+++ b/bazel/emscripten_deps.bzl
@@ -7,6 +7,7 @@ def _parse_version(v):
 
 BUILD_FILE_CONTENT_TEMPLATE = """
 package(default_visibility = ['//visibility:public'])
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
 
 filegroup(
     name = "all",
@@ -26,6 +27,7 @@ filegroup(
 filegroup(
     name = "emcc_common",
     srcs = [
+        "emscripten/embuilder.py",
         "emscripten/emcc.py",
         "emscripten/emscripten-version.txt",
         "emscripten/cache/sysroot_install.stamp",
@@ -68,6 +70,7 @@ filegroup(
         "bin/wasm-split{bin_extension}",
         "bin/wasm2js{bin_extension}",
         ":emcc_common",
+        ":embuilder",
     ] + glob(
         include = [
             "emscripten/cache/sysroot/lib/**",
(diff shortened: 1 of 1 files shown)


allsey87:
TODO: Set NODEJS environment variable to embuilder

allsey87:
@walkingeyerobot let me know if you think this is going in a reasonable direction or if you would do something differently

bazel/emscripten_deps.bzl
```diff
@@ -94,11 +97,65 @@ filegroup(
...
+        ":emscripten/embuilder.py",
+        ":compiler_files",
+        ":ar_files"
+    ],
+    srcs = [":embuilder_config"],
+    cmd = \"\"\"
+export EM_BINARYEN_ROOT=$$(realpath $$(dirname $$(dirname $(location :emscripten/embuilder.py))))
+export EM_LLVM_ROOT=$$EM_BINARYEN_ROOT/bin
+export EM_EMSCRIPTEN_ROOT=$$EM_BINARYEN_ROOT/emscripten
+export EM_CACHE=$(RULEDIR)/emscripten/cache
```
sbc100, line 118: These values normally part of the config file.   See https://github.com/emscripten-core/emsdk/blob/main/bazel/emscripten_toolchain/emscripten_config

I think the only one you want to override here is EM_CACHE.   The others should all come from that config file.

sbc100:
Is there some shared, writable location that we can use is the one true location for the emscripten cache?   Somewhere that is both readable and writable when the build rules run?

bazel/emscripten_deps.bzl
allsey87: The problem there (I think/I will test this) is that `emscripten_config` is part of the `@emsdk//` while this code is inside `@emscripten_bin_XXX//` so this might create a circular dependency.

> I think the only one you want to override here is EM_CACHE. The others should all come from that config file.

I would also need to set `EM_FROZEN_CACHE=0` since at this point the cache should be writable.

allsey87:
> Somewhere that is both readable and writable when the build rules run?

I guess Bazel wouldn't allow this since it could result in irreproducible builds, right?

allsey87:
I thought I had a really nice alternative solution to what I have done so far. The idea would have been to move cache generation from inside `emscripten_bin_XXX` to `emsdk` after the toolchain has been set up and then during builds override `EM_CACHE` on a per target basis.

For example, you set up a secondary cache in your WORKSPACE with something like:
```python
emsdk_emscripten_cache(
    name = "pic_cache"
    mode = ["--pic"],
    libraries = ["crtbegin", "libprintf_long_double-debug"]
)
```
and then for each target that needs to use this cache, you could set `EM_CACHE=//:pic_cache` via `env`. The problem, the meaning of `env` seems to vary a lot in Bazel. For foreign cc rules like `configure_make`, this works since the entries in `env` are set during build. However, for the inbuilt `cc_binary`, `env` appears to be only set at runtime...

allsey87:
Closed in favour of #1405

Issue #1399, opened 2024-06-09
[Bazel] Main bootstrap script for Wasm Audio Worklets not generated


WilliamIzzo83:
Trying to build the basic [noise generator example](https://github.com/emscripten-core/emscripten/blob/main/test/webaudio/audioworklet.c) with Bazel results in the audio worklet js file not generated.

Suppose we have a target defined like so

```python
load("@emsdk//emscripten_toolchain:wasm_rules.bzl", "wasm_cc_binary")

cc_binary(
    name = "audio_worklet",
    srcs = ["audio-worklet.c"],
    copts = [
        "-sWASM_WORKERS=1",
    ],
    linkopts = [
        "-sAUDIO_WORKLET",
        "-sWASM_WORKERS",
        "--oformat=html",
    ],
)

wasm_cc_binary(
    name = "audio_worklet_wasm",
    cc_target = "audio_worklet",
    outputs = [
        "index.html",
        "audio_worklet.js",
        "audio_worklet.wasm",
        "audio_worklet.aw.js"
    ],
)
```

Results in the following build error

```
Starting local Bazel server and connecting to it...
INFO: Analyzed target //:audio_worklet_wasm (80 packages loaded, 7258 targets configured).
INFO: From Action index.html:
[ERROR] Archive does not contain file with extname: .aw.js
ERROR: <redacted>/bazel_audio_worklet/BUILD:16:15: output 'audio_worklet.aw.js' was not created
ERROR: <redacted>/bazel_audio_worklet/BUILD:16:15: Action index.html failed: not all outputs were created or valid
Target //:audio_worklet_wasm failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 23.288s, Critical Path: 3.69s
INFO: 10 processes: 7 internal, 3 darwin-sandbox.
```

Here's a minimal example to reproduce the issue: 
[bazel_audio_worklet_ko.zip](https://github.com/user-attachments/files/15754202/bazel_audio_worklet_ko.zip)

[...]

PR #1404, opened 2024-06-12
[Bazel] Let audio workout bootstrap js file be generated.

WilliamIzzo83:
The allowed extension list in `link_wrapper.py` and in `wasm_cc_library.bzl` have been updated in order to support files with the `.aw.js` extensions that are generated when users set `-sAUDIO_WORKLET` compiler/linker flag.

Related issues:
- #1399


walkingeyerobot (approved):
thanks very much!

PR #1405, opened 2024-06-14
[Bazel] Support generating a secondary cache

### LINKED ISSUE #807: Bazel build with LTO fails
When using the Bazel emscripten toolchain as outlined [here](https://github.com/emscripten-core/emsdk/tree/master/bazel#bazel-emscripten-toolchain), when building and linking with LTO via `--copt="-flto" --linkopt="-flto"`, the linking step seems to fail because of the frozen cache setting in use [here](https://github.com/emscripten-core/emsdk/blob/master/bazel/emscripten_toolchain/emscripten_config#L8).

Reproducible so far on MacOS and Linux. This is an example error trace with a bit of internal nomenclature removed.
```
Linking of rule '<redacted>' failed (Exit 1): emcc_link.sh failed: error executing command external/emsdk/emscripten_toolchain/emcc_link.sh @bazel-out/wasm-opt-ST-53a11ea50094/bin/<path to target>-2.params
2021-04-29 13:02:20 
2021-04-29 13:02:20 Use --sandbox_debug to see verbose messages from the sandbox emcc_link.sh failed: error executing command external/emsdk/emscripten_toolchain/emcc_link.sh @bazel-out/wasm-opt-ST-53a11ea50094/bin/<path to target>-2.params
2021-04-29 13:02:20 
2021-04-29 13:02:20 Use --sandbox_debug to see verbose messages from the sandbox
2021-04-29 13:02:20 Traceback (most recent call last):
2021-04-29 13:02:20   File "emscripten_bin_linux/emscripten/emcc.py", line 3304, in <module>
2021-04-29 13:02:20     sys.exit(main(sys.argv))
2021-04-29 13:02:20   File "emscripten_bin_linux/emscripten/emcc.py", line 3297, in main
2021-04-29 13:02:20     ret = run(args)
2021-04-29 13:02:20   File "emscripten_bin_linux/emscripten/emcc.py", line 2105, in run
2021-04-29 13:02:20     extra_files_to_link += system_libs.calculate([f for _, f in sorted(temp_files)] + extra_files_to_link, link_as_cxx, forced=forced_stdlibs)
2021-04-29 13:02:20   File "emscripten_bin_linux/emscripten/tools/system_libs.py", line 1461, in calculate
2021-04-29 13:02:20     add_library(system_libs_map[forced])
2021-04-29 13:02:20   File "emscripten_bin_linux/emscripten/tools/system_libs.py", line 1452, in add_library
2021-04-29 13:02:20     libs_to_link.append((lib.get_path(), need_whole_archive))
2021-04-29 13:02:20   File "emscripten_bin_linux/emscripten/tools/system_libs.py", line 318, in get_path
2021-04-29 13:02:20     return shared.Cache.get_lib(self.get_filename(), self.build)
2021-04-29 13:02:20   File "emscripten_bin_linux/emscripten/tools/cache.py", line 130, in get_lib
2021-04-29 13:02:20     return self.get(name, *args, **kwargs)
2021-04-29 13:02:20   File "emscripten_bin_linux/emscripten/tools/cache.py", line 145, in get
2021-04-29 13:02:20     raise Exception('FROZEN_CACHE is set, but cache file is missing: %s' % shortname)
2021-04-29 13:02:20 Exception: FROZEN_CACHE is set, but cache file is missing: sysroot/lib/wasm32-emscripten/lto/libembind-rtti.a
```

### LINKED ISSUE #971: Unable to use LTO in Bazel.

### LINKED ISSUE #1099: Bazel: @emsdk cannot pick up SDL2 headers and libraries.

### LINKED ISSUE #1362: compile wams64 using bazel mode

### LINKED ISSUE #1401: [Bazel] Dynamically generate Emscripten cache

allsey87:
This is a working solution for generating a separate Emscripten cache. Note that this requires an additional entry in the workspace as follows:
```starlark
load("@emsdk//:emscripten_cache.bzl", emsdk_emscripten_cache = "emscripten_cache")
emsdk_emscripten_cache()
```
When used like this, the default Emscripten cache will be used. However, if the entry is as follows:
```starlark
load("@emsdk//:emscripten_cache.bzl", emsdk_emscripten_cache = "emscripten_cache")
emsdk_emscripten_cache(flags = ["--lto"])
```
Then embuilder will be called to build all system libraries and ports (i.e., the `ALL` option to embuilder) with the LTO option enabled. This can take awhile, so I have also made possible to specify which libraries you want to build explicitly:
```starlark
load("@emsdk//:emscripten_cache.bzl", emsdk_emscripten_cache = "emscripten_cache")
emsdk_emscripten_cache(
    flags = ["--lto"],
    libraries = [
        "crtbegin",
        "libprintf_long_double-debug",
        "libstubs-debug",
        "libnoexit",
        "libc-debug",
        "libdlmalloc",
        "libcompiler_rt",
        "libc++-noexcept",
        "libc++abi-debug-noexcept",
        "libsockets"
    ]
)
```

Resolves #807, resolves #971, resolves #1099, resolves #1362, resolves #1401


allsey87:
```
embuilder: warning: node version appears too old (seeing "v16.6.2", expected "v16.20.0") [-Wversion-check]
shared:INFO: (Emscripten: Running sanity checks)
embuilder:INFO: building crtbegin
cache:INFO: generating system asset: sysroot/lib/wasm32-emscripten/lto/crtbegin.o... (this will be cached in "/home/developer/.cache/bazel/_bazel_developer/ed66fc7424839955a359edda00eecd9b/external/emscripten_cache/cache/sysroot/lib/wasm32-emscripten/lto/crtbegin.o" for subsequent builds)
cache:INFO: generating system headers: sysroot_install.stamp... (this will be cached in "/home/developer/.cache/bazel/_bazel_developer/ed66fc7424839955a359edda00eecd9b/external/emscripten_cache/cache/sysroot_install.stamp" for subsequent builds)
cache:INFO:  - ok
  File "/home/developer/.cache/bazel/_bazel_developer/ed66fc7424839955a359edda00eecd9b/external/emscripten_bin_linux/emscripten/emcc.py", line 1602, in <module>
    sys.exit(main(sys.argv))
             ^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/developer/.cache/bazel/_bazel_developer/ed66fc7424839955a359edda00eecd9b/external/emscripten_bin_linux/emscripten/emcc.py", line 1595, in main
    ret = run(args)
          ^^^^^^^^^
  File "/home/developer/.cache/bazel/_bazel_developer/ed66fc7424839955a359edda00eecd9b/external/emscripten_bin_linux/emscripten/emcc.py", line 655, in run
    linker_inputs = phase_compile_inputs(options, state, newargs, input_files)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/developer/.cache/bazel/_bazel_developer/ed66fc7424839955a359edda00eecd9b/external/emscripten_bin_linux/emscripten/emcc.py", line 950, in phase_compile_inputs
    system_libs.ensure_sysroot()
  File "/usr/lib/python3.11/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/developer/.cache/bazel/_bazel_developer/ed66fc7424839955a359edda00eecd9b/external/emscripten_bin_linux/emscripten/tools/system_libs.py", line 2490, in ensure_sysroot
    cache.get('sysroot_install.stamp', install_system_headers, what='system headers')
  File "/home/developer/.cache/bazel/_bazel_developer/ed66fc7424839955a359edda00eecd9b/external/emscripten_bin_linux/emscripten/tools/cache.py", line 157, in get
    with lock(shortname):
  File "/usr/lib/python3.11/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "/home/developer/.cache/bazel/_bazel_developer/ed66fc7424839955a359edda00eecd9b/external/emscripten_bin_linux/emscripten/tools/cache.py", line 61, in lock
    acquire_cache_lock(reason)
  File "/home/developer/.cache/bazel/_bazel_developer/ed66fc7424839955a359edda00eecd9b/external/emscripten_bin_linux/emscripten/tools/cache.py", line 35, in acquire_cache_lock
    assert 'EM_CACHE_IS_LOCKED' not in os.environ, f'attempt to lock the cache while a parent process is holding the lock ({reason})'
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: attempt to lock the cache while a parent process is holding the lock (sysroot_install.stamp)
embuilder: error: subprocess 1/1 failed (returned 1)!
```

allsey87:
Note that although all checks are currently passing, this is only testing the pass-through behaviour for the moment. I will add additional checks that test the secondary cache shortly.

allsey87:
@sbc100, @walkingeyerobot, this is now ready for review. I came close to being able to support secondary caches on Windows, but there are some quirks to when Node.js is installed under `external` that do not apply to Linux and Mac. I would have liked to have gotten this working, but I have already spent a whole day on it and I really cannot spend any more of my time on this.

The steps for enabing Windows support in its current state are:
1. Uncomment line 37 and remove line 36 of `emscripten_cache.bzl`
2. Uncomment the test at the bottom of `test_bazel.ps1`

In this state, the test on Windows will fail on line 78 of `emscripten_cache.bzl` since Node.js is not available at `external/nodejs_windows_amd64/bin/nodejs/node.exe`.

bazel/emscripten_cache.bzl
```diff
@@ -0,0 +1,103 @@
+BUILD_FILE_CONTENT_TEMPLATE = """
+package(default_visibility = ['//visibility:public'])
+exports_files(['emscripten_config'])
+"""
+
+EMBUILDER_FILE_CONTENT_TEMPLATE = """
+CACHE = '{cache}'
+EMSCRIPTEN_ROOT = '{emscripten_root}'
```
sbc100, line 8: `EMSCRIPTEN_ROT` should not be needed here.

bazel/emscripten_cache.bzl
```diff
@@ -0,0 +1,103 @@
...
+    else:
+        fail('Unsupported operating system')
+
+def _emscripten_cache_impl(repository_ctx):
+    # Read the default emscripten configuration file
+    default_config = repository_ctx.read(
+        repository_ctx.path(
+            Label("@emsdk//emscripten_toolchain:default_config")
+        )
+    )
```
sbc100, line 47: If you can read the default config then the easiest thing to do is just to copy it.. and and a new `CACHE = ` line at the bottom.   Then you should be able to avoid the whole template thing above?

sbc100:
> Does this degrade Windows support in any way? Or does building non-prebuilt system libraries simply not work on Windows?

IIUC this doesn't degrade existing features but adds a new feature that simply doesn't work on windows (yet).   Seems like its probably worth addressing that before we land.

bazel/emscripten_cache.bzl
allsey87: That's not really possible. The default configuration relies on the environment variables `EM_BIN_PATH` and `ROOT_DIR` which are only set after the toolchain has been set up

allsey87:
> > Does this degrade Windows support in any way? Or does building non-prebuilt system libraries simply not work on Windows?
> 
> IIUC this doesn't degrade existing features but adds a new feature that simply doesn't work on windows (yet).

This is correct. Windows still works as normal with the prebuilt cache as long as `emscripten_cache` does not contain any arguments.

> Seems like its probably worth addressing that before we land.

The only path forward that I am aware of which might solve the problem of Node.js not being available on Windows in time would be to migrate to rules_js as proposed in #1388 since the newer repo uses the bzlmod approach (MODULE.bazel) instead of the WORKSPACE file to set up Node.js. I suspect that the bzlmod initialisation may happen earlier than the WORKSPACE which could mean that Node.js would be available on all platforms by time I need to run embuilder.

Would it be ok to add this on to this PR or should it be in a separate PR?

sbc100:
To be honest I'm surprised you need node_js configured at all to run embuilder... can't you just using a config file that doesn't contain node_js at all when you call embuilder?

Then create a new config file which is a copy of the core on, just with a new CACHE at the end, for use in the rest of the toolchain?   Also, it might be possible to just stick with the existing config file and set the EM_CACHE environment variable (for use in the rest of the toolchain that is.. i.e. after the toolchain has been setup).

allsey87:
> To be honest I'm surprised you need node_js configured at all to run embuilder... can't you just using a config file that doesn't contain node_js at all when you call embuilder?
> 
> Then create a new config file which is a copy of the core on, just with a new CACHE at the end, for use in the rest of the toolchain? Also, it might be possible to just stick with the existing config file and set the EM_CACHE environment variable (for use in the rest of the toolchain that is.. i.e. after the toolchain has been setup).

I just had a quick look into this. It might actually just be the call to `check_sanity` that results in the node version check and requirement on the `NODE_JS` variable. Looking at those functions, it might be possible to bypass that check by setting `EM_IGNORE_SANITY` in the environment and hence removing the dependency on Node.js...

sbc100:
> > To be honest I'm surprised you need node_js configured at all to run embuilder... can't you just using a config file that doesn't contain node_js at all when you call embuilder?
> > Then create a new config file which is a copy of the core on, just with a new CACHE at the end, for use in the rest of the toolchain? Also, it might be possible to just stick with the existing config file and set the EM_CACHE environment variable (for use in the rest of the toolchain that is.. i.e. after the toolchain has been setup).
> 
> I just had a quick look into this. It might actually just be the call to `check_sanity` in embuilder.py that results in the node version check and requirement on the `NODE_JS` variable. Looking at those functions, it might be possible to bypass that check by setting `EM_IGNORE_SANITY` in the environment and hence removing the dependency on Node.js...

We shouldn't be checking for NODE_JS untill we actually need it.   I can take a look at fixing that on the emscripten side if that is true.

sbc100:
Looks like you can probably just ignore the node version-check warnings if embuilder generates any.

allsey87:
> > > To be honest I'm surprised you need node_js configured at all to run embuilder... can't you just using a config file that doesn't contain node_js at all when you call embuilder?
> > > Then create a new config file which is a copy of the core on, just with a new CACHE at the end, for use in the rest of the toolchain? Also, it might be possible to just stick with the existing config file and set the EM_CACHE environment variable (for use in the rest of the toolchain that is.. i.e. after the toolchain has been setup).
> > 
> > 
> > I just had a quick look into this. It might actually just be the call to `check_sanity` in embuilder.py that results in the node version check and requirement on the `NODE_JS` variable. Looking at those functions, it might be possible to bypass that check by setting `EM_IGNORE_SANITY` in the environment and hence removing the dependency on Node.js...
> 
> We shouldn't be checking for NODE_JS untill we actually need it. I can take a look at fixing that on the emscripten side if that is true.

It is true, the call stack is main -> check_sanity -> perform_sanity_checks -> check_node_version.

> Then create a new config file which is a copy of the core on, just with a new CACHE at the end, for use in the rest of the toolchain?

This is already how things are working. The CACHE line is added to the end of `default_config` which is written out to `emscripten_config` and used by the toolchain.

> Also, it might be possible to just stick with the existing config file and set the EM_CACHE environment variable (for use in the rest of the toolchain that is.. i.e. after the toolchain has been setup).

I was experimenting with this approach initially, I will investigate and try to remember why I don't use the environment variable anymore.

allsey87:
> Looks like you can probably just ignore the node version-check warnings if embuilder generates any.

Wrong version is a warning. Node.js missing results in a crash.

sbc100:
> > Looks like you can probably just ignore the node version-check warnings if embuilder generates any.
> 
> Wrong version is a warning. Node.js missing results in a crash.

If necessary we could probably fix upstream emscripten so that we don't depend on node JS when running embuilder.   But fixing it downstream (here) seems reasonable too.. especially if we want to support older versions of emscripten.

allsey87:
> > > Looks like you can probably just ignore the node version-check warnings if embuilder generates any.
> > 
> > 
> > Wrong version is a warning. Node.js missing results in a crash.
> 
> If necessary we could probably fix upstream emscripten so that we don't depend on node JS when running embuilder. But fixing it downstream (here) seems reasonable too.. especially if we want to support older versions of emscripten.

So it turns out there are two places where Node.js is being checked:
1. [check_node_version](https://github.com/emscripten-core/emscripten/blob/main/tools/shared.py#L364)
2. [set_config_from_tool_location](https://github.com/emscripten-core/emscripten/blob/main/tools/config.py#L184)

But it is possible to work around both these checks with a small hack by setting the following environment variables:
```bash
EM_IGNORE_SANITY=1
EM_NODE_JS=empty
```
The `empty` bit is the small hack, it can be any non-empty string and is used to get us past the checks inside [`set_config_from_tool_location`](https://github.com/emscripten-core/emscripten/blob/main/tools/config.py#L97) by setting `val` to something that isn't "falsy".

allsey87:
And we have Windows support!! I am not sure why the `test-bazel-mac-arm64` is stuck at "Waiting for status to be reported" and not starting, but I think that is unrelated to this PR. Perhaps one of you two can start/trigger this check manually?

bazel/README.md
```diff
@@ -18,6 +18,9 @@ emsdk_deps()
 load("@emsdk//:emscripten_deps.bzl", emsdk_emscripten_deps = "emscripten_deps")
 emsdk_emscripten_deps(emscripten_version = "2.0.31")
 
+load("@emsdk//:emscripten_cache.bzl", emsdk_emscripten_cache = "emscripten_cache")
+emsdk_emscripten_cache()
+
```
sbc100, line 23: Do all users need to specify this?  I would hope that users who want to default cache can skip this?   In which case we can delete these two lines?

allsey87: I had a quick look and I suspect that it would require a new approach to solving the problem (and I don't really have the time for that right now).

Considering that [Bzlmod is replacing the WORKSPACE system](https://bazel.build/external/migration), I would like to ask to have this merged as is and when it comes time to transition emsdk to Bzlmod, we can find a new way to initialize emsdk and all of its components from the MODULE file, which I believe is more flexible.

sbc100: That seems unfortunate since won't all existing users now need to add these two lines?

allsey87: I slept on it and came up with another solution which works around this problem. See the latest commit, but in essence, we can move the cache initialization inside of the toolchain registration and use a default argument to maintain backwards compatibility. Let me know what you think!

bazel/test_secondary_lto_cache/BUILD
```diff
@@ -0,0 +1,21 @@
...
+wasm_cc_binary(
+    name = "hello-world-wasm",
+    cc_target = ":hello-world",
+    outputs = [
+        "hello-world.js",
+        "hello-world.wasm",
+    ],
+)
+
+
```
sbc100, line 21: Remove trailing lines here.

sbc100 (commented):
This looks way better now.  Thanks.

I leave it to @walkingeyerobot to have the final say on whether this makes sense to land.

allsey87:
I noticed today that my headers were still being included from the default cache location instead of the generated cache location. I think that is due to [toolchain.bzl#483](https://github.com/emscripten-core/emsdk/blob/main/bazel/emscripten_toolchain/toolchain.bzl#L483). Am I correct in thinking this path should be updated to point to the newly generated cache? I guess it doesn't really matter since we are just looking at header files?

sbc100:
The header files should be exactly the same so it shouldn't matter which sysroot is used for headers.

walkingeyerobot (approved):

allsey87:
@walkingeyerobot @sbc100 I think this PR will need to be manually merged due to a CircleCI gitch (`test-bazel-mac-arm64 is just waiting indefinitely`).

PR #1415, opened 2024-06-24
Ensure temporary files have proper unique names for bazel builds.

### LINKED ISSUE #1386: [Bazel] [Potential solution] Cannot build multiple emscripten binaries in parallel with bazel
On bazel with sandboxing disabled (or when not available, e.g. Windows), multiple parallel binaries will fail to build due to a shared file being written by the compiler: tmp.tar.

This filename comes from https://github.com/emscripten-core/emsdk/blob/ce74ca2b1c968f897150bdc55daa9e3c12a3fefc/bazel/emscripten_toolchain/link_wrapper.py#L162

I believe this can be easily solved by changing tmp.tar to a filename that includes the current build rule's name, to make it unique. But I haven't managed to test this.

kalmard0:
This attempts to fix #1386.


walkingeyerobot (approved):
Thanks very much!

PR #1419, opened 2024-06-28
Use "command -v" instead of "which" to detect commands

akoeplinger:
On a Linux distro that doesn't have the `which` program installed we're getting the following error:

    $ ./emsdk install latest
    ./emsdk: line 39: exec: python: not found

It's failing to detect the installed `python3` and falls back to using `python`, but this distro doesn't provide a python -> python3 symlink so we fail.

Fix this by using `command -v` instead which is a POSIX standard.
The same change went into emscripten a couple years ago: https://github.com/emscripten-core/emscripten/pull/15071

### DIFF
--- a/emsdk
+++ b/emsdk
@@ -26,7 +26,7 @@ fi
 # If bundled python is not found, look for `python3` in PATH.  This is especially important on macOS (See:
 # https://github.com/emscripten-core/emsdk/pull/273)
 if [ -z "$EMSDK_PYTHON" ]; then
-  if PYTHON3="$(which python3 2>/dev/null)"; then
+  if PYTHON3="$(command -v python3 2>/dev/null)"; then
     EMSDK_PYTHON=$PYTHON3
   fi
 fi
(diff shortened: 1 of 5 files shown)


emsdk
```diff
@@ -26,7 +26,7 @@ fi
 # If bundled python is not found, look for `python3` in PATH.  This is especially important on macOS (See:
 # https://github.com/emscripten-core/emsdk/pull/273)
 if [ -z "$EMSDK_PYTHON" ]; then
-  if PYTHON3="$(which python3 2>/dev/null)"; then
+  if PYTHON3="$(command -v python3 2>/dev/null)"; then
```
sbc100, line 29: I think all the other changes apart from this line can be reverted.   

This is the only place where the code really needs to be portable and run on many different distros.  All the other places are basically build scripts where we control the environment.

sbc100 (approved):
lgtm, with a limiting of the scope of this change.

emsdk
akoeplinger: @sbc100 done

allsey87:
I think this is fine for now, but eventually it would be good to migrate to `rules_python` which (I believe) should be more portable and eliminate the need for most of the `.sh` and `.bat` scripts.

sbc100 (approved):

Issue #1263, opened 2023-08-07
Bazel wrapper uses system python instead of py_binary


logankaser:
It seems like the bazel wrapper escapes the bazel sandbox and relies on having the correct version of python installed on the system. Is there a way to force it to use the bazel python tool chain?

allsey87:
> It seems like the bazel wrapper escapes the bazel sandbox

Could you go into a bit more detail here? What sort of issues has this caused you? Referring to #1422, I am just using the native implementation of `py_binary` (not the one from `rules_python`) which I suspect still uses the system installed Python.

If I get time to work on #1422, I intend to switch it over to `python_rules` since that work around https://github.com/bazelbuild/bazel/issues/17629.

logankaser:
> > It seems like the bazel wrapper escapes the bazel sandbox
> 
> Could you go into a bit more detail here? What sort of issues has this caused you? Referring to #1422, I am just using the native implementation of `py_binary` (not the one from `rules_python`) which I suspect still uses the system installed Python.
> 
> If I get time to work on #1422, I intend to switch it over to `python_rules` since that might work around [bazelbuild/bazel#17629](https://github.com/bazelbuild/bazel/issues/17629).

Fro example, https://github.com/emscripten-core/emsdk/blob/ce1d281ce65b5c7398b17ed6484efc6a3901df6f/bazel/emscripten_toolchain/emcc.sh
just invokes system python3.

allsey87:
> For example, https://github.com/emscripten-core/emsdk/blob/ce1d281ce65b5c7398b17ed6484efc6a3901df6f/bazel/emscripten_toolchain/emcc.sh
just invokes system python3.

Indeed it does, I am not sure what I was on about in my previous comment.

Issue #1420, opened 2024-07-01
[bazel] [rules_foreign_cc] ModuleNotFoundError: No module named 'tools'


allsey87:
When building a simple hello world binary with Meson, I am hitting the following error
```
mesonbuild.utils.universal.EnvironmentException: Unknown compiler(s): [['/home/developer/.cache/bazel/_bazel_developer/25e07d78077dfe1eca932359d50e41ef/sandbox/processwrapper-sandbox/9/execroot/_main/external/emsdk/emscripten_toolchain/emcc.sh']]
```
Digging a bit deeper, it seems `emcc.py --version` is failing and unable to find the tools module.

```
The Meson build system
Version: 1.1.1
Source dir: /home/developer/.cache/bazel/_bazel_developer/25e07d78077dfe1eca932359d50e41ef/sandbox/processwrapper-sandbox/9/execroot/_main/testing/hellomeson
Build dir: /home/developer/.cache/bazel/_bazel_developer/25e07d78077dfe1eca932359d50e41ef/sandbox/processwrapper-sandbox/9/execroot/_main/bazel-out/k8-fastbuild/bin/testing/test_meson.build_tmpdir
Build type: native build
Project name: tutorial
Project version: undefined
running cmd: ['/home/developer/.cache/bazel/_bazel_developer/25e07d78077dfe1eca932359d50e41ef/sandbox/processwrapper-sandbox/9/execroot/_main/external/emsdk/emscripten_toolchain/emcc.sh', '--version']
out: 
err: Traceback (most recent call last):
  File "/home/developer/.cache/bazel/_bazel_developer/25e07d78077dfe1eca932359d50e41ef/sandbox/processwrapper-sandbox/9/execroot/_main/external/emscripten_bin_linux/emscripten/emcc.py", line 23, in <module>
    from tools.toolchain_profiler import ToolchainProfiler
ModuleNotFoundError: No module named 'tools'
```
Perhaps this isn't included in the `compiler_files` filegroup which is passed to the toolchain?

allsey87:
> Perhaps this isn't included in the compiler_files filegroup which is passed to the toolchain?

This can't be the case, this module import is the first statement in emcc.py...

allsey87:
[...]

PR #1422, opened 2024-07-04
[Bazel] Use py_binary for emcc

allsey87:
This PR was an investigation into clean up the Emscripten toolchain in Bazel using `py_binary` to wrap and run `emcc` and `emar` etc. This would have been nice since I think it would have allowed us to drop all the sh/bat scripts (and OS detection logic around it) that just call the Python interpreter anyway. The motivation for this is #1420 and I was hoping that by wrapping `emcc` with `py_binary` that the paths/modules/libraries would have been a bit more robust to whatever is going wrong in this issue.

Unfortunately I hit the roadblock described in this [SO question](https://stackoverflow.com/questions/78703242/how-do-i-provide-a-py-binarys-runfiles-to-a-c-toolchain-in-bazel) that cannot be worked around without a hack. The main blocker is https://github.com/bazelbuild/bazel/issues/17629, however, if there is interest in this approach, we could conditionally apply a genrule on Linux and Mac to add in the missing shebang as suggested in https://github.com/bazelbuild/bazel/issues/17629#issuecomment-2207375102.

What do you think @sbc100 and @walkingeyerobot? Do you think this is worth pursuing further? Incidentally, this PR would make #1333 redundant in my opinion.

### DIFF
--- a/bazel/emscripten_toolchain/BUILD.bazel
+++ b/bazel/emscripten_toolchain/BUILD.bazel
@@ -12,6 +12,14 @@ filegroup(
     ],
 )
 
+py_binary(
+    name = "emcc_compile",
+    main = "emscripten/emcc.py",
+    srcs = [
+        "@emscripten_bin_linux//:emscripten/emcc.py",
+    ],
+)
+
 filegroup(
     name = "compiler_files",
     srcs = [
@@ -19,6 +27,7 @@ filegroup(
         "emcc.bat",
         "@emsdk//:compiler_files",
         ":common_files",
+        ":emcc_compile"
     ],
 )
 
@@ -66,6 +75,7 @@ emscripten_cc_toolchain_config_rule(
         "@bazel_tools//src/conditions:host_windows": "bat",
         "//conditions:default": "sh",
     }),
+    emcc = ":emcc_compile",
 )
 
 cc_toolchain(
--- a/bazel/emscripten_toolchain/toolchain.bzl
+++ b/bazel/emscripten_toolchain/toolchain.bzl
@@ -81,7 +81,7 @@ def _impl(ctx):
     ################################################################
     # Tools
     ################################################################
-    clang_tool = tool(path = emcc_script)
+    clang_tool = tool(tool = ctx.executable.emcc)
     clif_match_tool = tool(path = "dummy_clif_matcher")
     link_tool = tool(path = emcc_link_script)
     archive_tool = tool(path = emar_script)
@@ -1135,6 +1135,7 @@ emscripten_cc_toolchain_config_rule = rule(
         "em_config": attr.label(mandatory = True, allow_single_file = True),
         "emscripten_binaries": attr.label(mandatory = True, cfg = "exec"),
         "script_extension": attr.string(mandatory = True, values = ["sh", "bat"]),
+        "emcc": attr.label(mandatory = True, executable = True, allow_files = True, cfg = "exec"),
     },
     provides = [CcToolchainConfigInfo],
 )


allsey87:
Referring to the CI results, on Linux and Mac this currently fails as described in the [SO question](https://stackoverflow.com/questions/78703242/how-do-i-provide-a-py-binarys-runfiles-to-a-c-toolchain-in-bazel) and as follows:
```
Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
  File "bazel-out/k8-opt-exec-ST-d57f47055a04/bin/emscripten_toolchain/emcc_compile", line 559, in <module>
    Main()
  File "bazel-out/k8-opt-exec-ST-d57f47055a04/bin/emscripten_toolchain/emcc_compile", line 457, in Main
    module_space = FindModuleSpace(main_rel_path)
  File "bazel-out/k8-opt-exec-ST-d57f47055a04/bin/emscripten_toolchain/emcc_compile", line 172, in FindModuleSpace
    raise AssertionError('Cannot find .runfiles directory for %s' % sys.argv[0])
AssertionError: Cannot find .runfiles directory for bazel-out/k8-opt-exec-ST-d57f47055a04/bin/emscripten_toolchain/emcc_compile
Target //hello-world:hello-world-wasm failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 47.031s, Critical Path: 0.11s
INFO: 9 processes: 9 internal.
ERROR: Build did NOT complete successfully
```
Interestingly enough, it gets further on Windows and actually starts emcc.py but then fails since I haven't added the required `deps` and `data` to the `py_binary` yet. Although this could be a quirk of using a very old Bazel version on Windows (5.4.0).
```
ERROR: C:/users/circleci/project/bazel/hello-world/BUILD:4:10: Compiling hello-world/hello-world.cc failed: (Exit 1): emcc_compile.exe failed: error executing command bazel-out\x64_windows-opt-exec-2B5CBBC6-ST-4a8fd3122b04\bin\emscripten_toolchain\emcc_compile.exe @bazel-out/wasm-fastbuild-ST-7eb0ea6b7782/bin/hello-world/_objs/hello-world/hello-world.o.params
  File "\\?\C:\Users\circleci\AppData\Local\Temp\Bazel.runfiles_mbfhpz1e\runfiles\emscripten_bin_linux\emscripten\emcc.py", line 23, in <module>
    from tools.toolchain_profiler import ToolchainProfiler
ModuleNotFoundError: No module named 'tools'
Target //hello-world:hello-world-wasm failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 91.945s, Critical Path: 0.54s
INFO: 11 processes: 9 internal, 2 local.
FAILED: Build did NOT complete successfully
FAILED: Build did NOT complete successfully
```

sbc100:
`py_binary` seems like it might be little overkill for this situation, but if it works I won't object.   I assume there is an easier easy to get a hermetic python runtime and run a script without having to package that script as a py_binary/

allsey87:
> `py_binary` seems like it might be little overkill for this situation, but if it works I won't object. I assume there is an easier easy to get a hermetic python runtime and run a script without having to package that script as a py_binary/

Well, I've dug pretty deep into Bazel over the past weeks and while I could be mistaking, I am under the impression that what we are trying to do with `emcc` etc is actually one of the primary use cases for `py_binary`...

sbc100:
> > `py_binary` seems like it might be little overkill for this situation, but if it works I won't object. I assume there is an easier easy to get a hermetic python runtime and run a script without having to package that script as a py_binary/
> 
> Well, I've dug pretty deep into Bazel over the past weeks and while I could be mistaking, I am under the impression that what we are trying to do with `emcc` etc is actually one of the primary use cases for `py_binary`...

Fair enough.  Sounds like you know way more than me at this point about this stuff :)

I though that `py_binary` was for producing toolchain outputs, but if it also makes sense to managing build tools within the toolchain itself then that makes sense to me.

allsey87:
> I though that `py_binary` was for producing toolchain outputs, but if it also makes sense to managing build tools within the toolchain itself then that makes sense to me.

You are not wrong, it is for both. Ironically it seems, however, that people actually run into more problems using `py_binary` externally (i.e., as an output of Bazel) than using it internally as a tool or via `bazel run`.

logankaser:
This would also resolve #1263

logankaser:
Would also resolve #1263

sbc100:
Are you still interesting in working on this?

@DoDoENT do you think is path makes sense?

DoDoENT:
> Are you still interesting in working on this?
> 
> 
> 
> @DoDoENT do you think is path makes sense?

I doubt. I tried solving that problem with with the same approach some half a year ago (when I was fixing issue where macOS python was too old for smack), and the most I managed to do is the current situation. 

I think that anything more would probably require changes in Hazel or in python tool chain or maybe even in rules_cc 🤷🏻‍♂️. I really don't know.

sbc100:
OK, closing for now.  Feel free to re-open if needed.

PR #1429, opened 2024-07-24
fix: improve Fish shell support (and Cshell)

mtb0x1:
I faced the same issue mentioned in #1299, this PR fixes the issue by taking in account specific `set` and `unset` syntax for Fish shell.

The post install instruction (i.e : `./emsdk activate latest`) provides wrong instructions in Fish shell and Cshell cases. 
This should be fixed by this PR too, although the code is a bit redundant and it could be more cleaner. 

The [website instruction](https://emscripten.org/docs/getting_started/downloads.html#installation-instructions-using-the-emsdk-recommended)  have the same issue, but I guess user with specific shell will be able to manage.

PR #1433, opened 2024-08-01
Add missing comma from #1429

sbc100:
Turns out this is what was causing the OOM when building binaryen since we were clobbering the EMSDK_NUM_CORES environment variable by mistake.

See #1429


mtb0x1:
my bad, at least it was fixed fast. Thanks !

PR #1436, opened 2024-08-18
bazel: update rules_nodejs and migrate to rules_js

eric-higgins-ai:
This finishes the work started in https://github.com/emscripten-core/emsdk/pull/1388 by fixing CI. It avoids a breaking change by:
* Using the latest rules_js 1.x.x version, instead of updating to rules_js 2 (which removes support for bazel 5).
* Copying the contents of [rules_js_dependencies](https://github.com/aspect-build/rules_js/blob/main/js/repositories.bzl) instead of calling it, as the call would need to be added by users in their `WORKSPACE` files

Context from the previous PR:

> Bazel's Node.js dependency comes from [rules_nodejs](https://github.com/bazelbuild/rules_nodejs/). Previously, bazel/deps.bzl was using rules_nodejs 5.8.0, released in 2022 and only supported Node.js toolchains up to 18.12.1.

> This PR bumps rules_nodejs to latest 6.1.1. It also replaces build_bazel_rules_nodejs with [rules_js](https://github.com/aspect-build/rules_js), since npm_install that bazel/emscripten_deps.bzl used was deprecated. The README of rules_nodejs now recommends migrating to rules_js for everything other than the Node.js toolchain: (https://github.com/bazelbuild/rules_nodejs/commit/371e8cab15f9db444217703144b2eb71d0e13235)

> Impetus
Our repo builds with Bazel and uses Emscripten and Node.js. Tried to upgrade Node.js 18 to Node.js 20 and saw that emsdk didn't support rules_nodejs 6+ in the same workspace.

Similarly, it's not possible to update to rules_js v2 in a workspace that also references `emsdk`.

### DIFF
--- a/bazel/README.md
+++ b/bazel/README.md
@@ -15,9 +15,6 @@ git_repository(
 load("@emsdk//:deps.bzl", emsdk_deps = "deps")
 emsdk_deps()
 
-load("@aspect_rules_js//js:repositories.bzl", "rules_js_dependencies")
-rules_js_dependencies()
-
 load("@emsdk//:emscripten_deps.bzl", emsdk_emscripten_deps = "emscripten_deps")
 emsdk_emscripten_deps(emscripten_version = "2.0.31")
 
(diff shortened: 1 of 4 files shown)


walkingeyerobot (approved):
aspect is fine to use

PR #1444, opened 2024-09-03
Fix 'emsdk install sdk-main-64bit' to produce a functioning installation.

juj:


emsdk.py
```diff
@@ -1332,6 +1332,16 @@ def emscripten_npm_install(tool, directory):
...
 
   print('Done running: npm ci')
+
+  try:
+    subprocess.check_output([sys.executable, 'bootstrap.py'],
+      cwd=directory, stderr=subprocess.STDOUT, env=env,
+      universal_newlines=True)
+  except subprocess.CalledProcessError as e:
+    errlog('Error running %s:\n%s' % (e.cmd, e.output))
+    return False
```
sbc100, line 1342: I don't think the try/catch here should be necessary.

`check_output`, on its own, should already report the name of the script that failed (I think).

Alternatively, perhaps this try/catch with `check_output` can be factored out since its a pattern used above.

lgtm either way though.

Its a shame we can't really a CI test for this because I think it requires compiling llvm and binaryen from source, right?

emsdk.py
```diff
@@ -1332,6 +1332,16 @@ def emscripten_npm_install(tool, directory):
       return False
 
   print('Done running: npm ci')
+
```
sbc100, line 1335: Do we still all the code above for patching closure compiler.   

The comment says "delete the Java version, since that takes up ~12.5MB of installation space that is no longer needed.".  Is it really worth the extra complexity to save 12.5mb locally?   Or are you/we shipping the resulting local emsdk installation?  Even so, is 12.5mb worth it?

juj: We are shipping the resulting emsdk installation.

I think it is worth the save, given that it is "easy" dead code. Our WebGL support module gets downloaded by thousands and thousands of developers, it all accumulates up in CDN costs.

sbc100: SGTM.

emsdk.py
juj: The idea here I think is to stop propagating the exception - I reused the same pattern that the above existing try-catch uses.

Yeah, there is probably no good way to avoid testing this unless we would compile from source. (we could inject some kind of fake mocking, but that would have some brittle)

emsdk.py
sbc100, line 1335: Since the bootstrap process involves running `npm ci`, I wonder if this above running of `npm ci` is redundant, or if the addition use of the `--production` and `--no-optional` is somehow undone by running `bootstrap` here.

juj: I don't know, however, such a change of removing `npm ci` would be best for another commit for bisectability.

> if the addition use of the `--production` and `--no-optional` is somehow undone by running `bootstrap` here.

If that is the case, then there is something to fix in `bootstrap.py`, as currently Emscripten cannot be used without calling bootstrap.py.

sbc100 (approved):

PR #1470, opened 2024-10-18
Remove temp zip after uncompressing in update_python.py

juj:
Remove temp zip after uncompressing in update_python.py to avoid confusing it with the output artifact zip. Adjust comments.

### DIFF
--- a/scripts/update_python.py
+++ b/scripts/update_python.py
@@ -34,6 +34,8 @@ from subprocess import check_call
 version = '3.9.2'
 major_minor_version = '.'.join(version.split('.')[:2])  # e.g. '3.9.2' -> '3.9'
 download_url = 'https://www.nuget.org/api/v2/package/python/%s' % version
+# This is not part of official Python version, but a repackaging number appended by emsdk
+# when a version of Python needs to be redownloaded.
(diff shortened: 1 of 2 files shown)


scripts/update_python.py
```diff
@@ -153,6 +156,10 @@ def build_python():
     pybin = os.path.join(src_dir, 'install', 'usr', 'local', 'bin', 'python3')
     pip = os.path.join(src_dir, 'install', 'usr', 'local', 'bin', 'pip3')
     check_call([pybin, '-m', 'ensurepip', '--upgrade'])
+    # TODO: Potential bug: the following cmdline does not pin down a version
+    # of requests module, resulting in possibly different version of the module
+    # being installed on future runs. Switch to pip install requests==<version> to
+    # to download a pinned version.
```
sbc100, line 162: Maybe just add a version here instead of this comment?

sbc100 (approved):

scripts/update_python.py
juj: I don't know which version to add. Maybe if I get time to run this on a mac I'll check what version it was.

PR #1471, opened 2024-10-18
Use zip_cmd() and unzip_cmd() in update_node.py as well.

juj:
This way update_node.py can be called on Windows host machines as well.

### DIFF
--- a/scripts/update_node.py
+++ b/scripts/update_node.py
@@ -15,6 +15,7 @@ import urllib.request
 import subprocess
 import os
 import shutil
+from zip import unzip_cmd, zip_cmd
 
 version = '18.20.3'
 base = 'https://nodejs.org/dist/latest-v18.x/'
@@ -37,13 +38,13 @@ for suffix in suffixes:
     urllib.request.urlretrieve(download_url, filename)
 
     if '-win-' in suffix:
-      subprocess.check_call(['unzip', '-q', filename])
+      subprocess.check_call(unzip_cmd() + [filename])
       dirname = os.path.splitext(os.path.basename(filename))[0]
       shutil.move(dirname, 'bin')
       os.mkdir(dirname)
       shutil.move('bin', dirname)
       os.remove(filename)
-      subprocess.check_call(['zip', '-rq', filename, dirname])
+      subprocess.check_call(zip_cmd() + [filename, dirname])
       shutil.rmtree(dirname)
 
     upload_url = upload_base + filename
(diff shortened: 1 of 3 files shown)


scripts/update_node.py
```diff
@@ -37,13 +38,13 @@
...
     if '-win-' in suffix:
-      subprocess.check_call(['unzip', '-q', filename])
+      subprocess.check_call(unzip_cmd() + [filename])
       dirname = os.path.splitext(os.path.basename(filename))[0]
       shutil.move(dirname, 'bin')
       os.mkdir(dirname)
       shutil.move('bin', dirname)
       os.remove(filename)
-      subprocess.check_call(['zip', '-rq', filename, dirname])
+      subprocess.check_call(zip_cmd() + [filename, dirname])
```
sbc100, line 47: Since this is only ever run once every years or two a developer's machine (normally my machine) I think its reasonable to have `zip` in the path.  

I'd rather not split emsdk.py into multiple files for this reason.

sbc100: Also, I wonder if we should just update emsdk so that it doesn't require this modification to the windows node archive?

juj: I vaguely recall that is how it was originally, but someone reported an issue that windows doesn't have bin/ but linux and macos have. I prefer the consistency, so that the tools are in the same directory and can be run with `bin/node%.exe%` without needing to write platform-specific paths for each.

juj: > I think its reasonable to have `zip` in the path.

The fix is to be able to run this tool on Windows with 7z. I don't know where to obtain zip on Windows, and we've only used 7z.exe ever. If it's good for update_python.py, it should be good for update_node.py as well?

> I'd rather not split emsdk.py into multiple files for this reason.

These are not used in emsdk.py, but just in these two python files in scripts/ . I needed to be able to run update_node.py to develop Windows on ARM support.

sbc100 (approved):
Sorry, I thought there functions were being split out from emsdk.py.

Thanks change LGTM.

BTW could you make this change simpler/smaller by doing `from update_python import ...`, avoiding the new file completely?

juj:
> BTW could you make this change simpler/smaller by doing `from update_python import ...`, avoiding the new file completely?

Is the new file problematic in this case? I would think that it would be nicer to have the "active" scripts update_python and update_node be independent of each other, and only depend on a common tool script? Having update_node depend on update_python seems a bit off?

juj:
If you feel strongly, I can follow up, though I thought this could be more idiomatically structured?

PR #1472, opened 2024-10-18
Only upload node with gsutil in update_node.py if --upload is passed.

juj:
This allows running the script locally.

Is there some kind of automated script that needs a matching update, or was this always run manually?

### DIFF
--- a/scripts/update_node.py
+++ b/scripts/update_node.py
@@ -13,6 +13,7 @@ direcotry.
 
 import urllib.request
 import subprocess
+import sys
 import os
 import shutil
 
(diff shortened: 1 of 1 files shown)


sbc100:
This script is always run manually.   I'm happy to run it if you need a new version uploaded.

sbc100 (commented):
Is there any point in running this script if you don't then upload the binaries? 

i.e. the whole point of this script is to upload the binaries I think.   Are you imagining some other purpose?

juj:
Yes, I can run the script locally to verify that the PR I am proposing to you works, even though I am not the one doing the uploads.

Without this PR, I could not verify that my addition of arm64 version of node would actually work, since in the inner loop it would abort running this script with an exception. update_python.py already had `--upload`, so having the same in `update_node.py` seems reasonable?

sbc100:
In that case maybe you could call the new flag `-n/--try-run`?

sbc100 (approved):
Actually, since the python script already has that then this lgtm as is.

juj:
> In that case maybe you could call the new flag `-n/--dry-run`?

I didn't see update_node.py having this flag.. so I monkeyed the same structure in update_node.py from update_python.py.

(I'll put up a PR to update python for ARM64 separately)

PR #1473, opened 2024-10-18
Update Node to 20.18.0 LTS and add support for Windows on ARM64 to Node packages.

juj:
Earlier Node 18.20.3 LTS did not yet support Windows on ARM64 builds.

Node 20.18.0 LTS is released on https://nodejs.org/en/blog/release/v20.18.0.

This update currently only affects the "compiled from source" builds, and not the Google precompiled releases.

### DIFF
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -107,6 +107,56 @@
 
 
   {
+    "id": "node",
+    "version": "20.18.0",
+    "bitness": 32,
+    "arch": "x86",
+    "windows_url": "node-v20.18.0-win-x86.zip",
+    "activated_path": "%installation_dir%/bin",
+    "activated_path_skip": "node",
+    "activated_cfg": "NODE_JS='%installation_dir%/bin/node%.exe%'",
+    "activated_env": "EMSDK_NODE=%installation_dir%/bin/node%.exe%"
+  },
+  {
+    "id": "node",
+    "version": "20.18.0",
+    "arch": "arm",
+    "bitness": 32,
+    "linux_url": "node-v20.18.0-linux-armv7l.tar.xz",
+    "activated_path": "%installation_dir%/bin",
+    "activated_path_skip": "node",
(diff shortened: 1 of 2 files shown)


sbc100:
Do we/I need also upload these binaries to cloud storage bucket?

sbc100:
I think we should probably update the binary distribution too.  i.e. I think it would be good to have just one version that we all use.

juj:
> Do we/I need also upload these binaries to cloud storage bucket?

Yes, these versions will then need to be uploaded.

> I think we should probably update the binary distribution too. i.e. I think it would be good to have just one version that we all use.

I was kind of hoping that maybe you could find a moment to work on that? The problem is that I don't know how to split so that the old binary versions keep using the existing node, and new binary versions will pick up the new node? I.e. how do the google waterfalls decide on the versioning for the binary releases you produce.

sbc100:
I don't know of an easy way to have different versions of the emscripten-releases binaries have different dependencies.   So either we would need to invent such a mechanism, or we would just have to accept that older emsdk versions also get a newer node version (which I think in this case is probably fine).

I'll start running the script to upload the binaries now.

sbc100:
node v20.18.0 binaries are now uploaded the cloud storage.

juj:
> we would just have to accept that older emsdk versions also get a newer node version (which I think in this case is probably fine).

Personally that would be fine to me. I recommend bumping that in the manifest as a separate PR.

> node v20.18.0 binaries are now uploaded the cloud storage.

Awesome, thanks! Let me run a double check install to see that this works in this branch.

sbc100 (commented):
Fair enough.   This change LGTM then.  Can you followup with another PR to update the emscripten-releases binaries?

The good thing about updating the emscripten-releases dependencies in this PR is that you would get some decent test coverage in the CI.

sbc100 (approved):

juj:
> Can you followup with another PR to update the emscripten-releases binaries?

Sure. I'll land this PR after I've locally built to verify that the SDK installs ok.

PR #1474, opened 2024-10-18
Add downloaded and generated archive files to gitignore

juj:


.gitignore
```diff
@@ -33,3 +33,8 @@ __pycache__
 /mingw
 /spidermonkey
 /binaryen
+
+# Generated and downloaded packages
+*.zip
+*.tar.gz
+*.tar.xz
```
sbc100, line 40: Shouldn't this already be covered by `downloads/`?

juj: I got this with

```
10/18/2024  12:41 PM        42,013,950 node-v20.18.0-darwin-arm64.tar.gz
10/18/2024  12:41 PM        43,660,520 node-v20.18.0-darwin-x64.tar.gz
10/18/2024  12:41 PM        24,884,024 node-v20.18.0-linux-arm64.tar.xz
10/18/2024  12:41 PM        22,088,076 node-v20.18.0-linux-armv7l.tar.xz
10/18/2024  12:41 PM        25,738,964 node-v20.18.0-linux-x64.tar.xz
10/18/2024  12:41 PM        26,229,598 node-v20.18.0-win-arm64.zip
10/18/2024  12:41 PM        29,622,807 node-v20.18.0-win-x64.zip
10/18/2024  12:40 PM        27,572,492 node-v20.18.0-win-x86.zip
```
in scripts/

sbc100: I think I'd rather know that they are there so that I know to go an clean up them when I'm done (e.g. with `git clean`).   

The update_node script is very rarely run so it shouldn't happen very often.  

Alternatively we could have the update scripts download in the `downloads` folder.  But personally I'm fine with these rarely-run script creating files like this that show up it git status.

sbc100: I actually quote like being able to do `git status` to see the names of the different files that were downloaded when updating node/python but maybe thats just me?

juj: Ok, sure - not critical in any way.

PR #1477, opened 2024-10-18
Update Python to 3.13.0 and add support to Windows on AMD64.

juj:
Update Python to 3.13.0 and add support to Windows on AMD64. Rewrite packaging of pywin32 because it is no longer shipped via separate binary installers, but has migrated to use pip.


scripts/update_python.py
```diff
@@ -32,29 +32,35 @@
 from subprocess import check_call
 from zip import unzip_cmd, zip_cmd
 
-version = '3.9.2'
+version = '3.13.0'
```
sbc100, line 35: Do we need to upgrade to such a recent version?  i.e. what as the first version of which arm64 packages are available.

The slight downside of using a bleeding edge version of python like this is that we accidentally start depending of very new python features that folks on older/stable versions might not have.

juj: I was pondering about this too. My thinking here is that since we do these updates so rarely, maybe we would for once update to a modern version.

Python 3.9, 3.10 and 3.11 will no longer receive bugfixes to any issues one might run into. I don't think it makes sense to update to any of these versions, even if one of these would be the earliest ones to support Windows on ARM64.

Python 3.14 is in feature development. So we definitely will not want to update to that branch.

Python 3.12 and Python 3.13 have finished feature development and are actively receiving bufixes. It feels that either of these would be the ones to focus on.

I.e.

image

I picked 3.13 since it is newer, and I could not find any mentions of OS support cutoff differences between 3.12 and 3.13.

> we accidentally start depending of very new python features that folks on older/stable versions might not have.

3.13 is stable. (or rather, actually, every python version is broken in different ways. But Emscripten test suite is fortunately evolved to be good at finding ways that Python works or doesn't work out)

We do bundle python for the very purpose to normalize which version we provide to people? So this covers Windows and macOS users. On Linuxes where this currently is not the case, I understand updating to a newer version should be possible easier with their package managers?

juj: I suppose the other alternative could be [3.12.7](https://www.python.org/downloads/release/python-3127/) . At the end of the day I don't feel that strongly, but generally would strive to standardize to latest version, especially if Emscripten test suite will pass with 3.13.0. (Testing that out on Windows, although I am seeing Windows failures already before this update, making it a bit more time consuming to compare)

sbc100: sgtm

sbc100:
I guess to perform that actual upload I should download this PR and run the update script?

README.md
```diff
@@ -65,7 +65,7 @@ https://emscripten.org/docs/building_from_source/toolchain_what_is_needed.html.
 
 ### Linux
 
-- `python`: Version 3.9.2 or above.
+- `python`: Version 3.13.0 or above.
```
sbc100, line 68: I'm not sure we want to drop support for python 3.9 just because emsdk bundles a bleeding edge version.   I think the idea is that we still want to support the system version of python on older/stable linux releases.

sbc100: In fact I suggest we keep the linux version of emsdk pinned to the older version of python to avoid accidentally breaking compat.

scripts/update_python.py
```diff
@@ -10,15 +10,19 @@
 We only supply binaries for windows and macOS, but we do it very different ways for those two OSes.
```
sbc100, line 10: Perhaps add another sentence here:  "On linux, we simply depend on the system version of python".

sbc100 (approved):
lgtm % comments

scripts/update_python.py
cclauss: ;-)  https://www.python.org/downloads/release/python-3131/
```suggestion
version = '3.13.1'
```

juj:
If it helps, I can also generate the artifacts and upload them to my dev box for you to pull?

juj:
https://oummg.com/dump/emscripten/python-3.13.3-0-win-amd64.zip
https://oummg.com/dump/emscripten/python-3.13.3-0-win-arm64.zip

sbc100:
> https://oummg.com/dump/emscripten/python-3.13.3-0-win-amd64.zip https://oummg.com/dump/emscripten/python-3.13.3-0-win-arm64.zip

Don't we need the macOS binaries too?

emsdk_manifest.json
```diff
@@ -350,13 +388,13 @@
   {
     "version": "main",
     "bitness": 64,
-    "uses": ["python-3.9.2-nuget-64bit", "llvm-git-main-64bit", "node-20.18.0-64bit", "emscripten-main-64bit", "binaryen-main-64bit"],
+    "uses": ["python-3.13.0-64bit", "llvm-git-main-64bit", "node-20.18.0-64bit", "emscripten-main-64bit", "binaryen-main-64bit"],
```
sbc100, line 391: 3.13.3?

scripts/update_python.py
```diff
@@ -8,17 +8,22 @@
...
+  3. Re-zip and upload to storage.google.com
 
 macOS recipe:
   1. Clone cpython
   2. Use homebrew to install and configure openssl (for static linking!)
   3. Build cpython from source and use `make install` to create archive.
+
+Raspberry Pi Debian 12 (Bookworm):
+  1. Before calling this script, run "sudo apt install libssl-dev", or otherwise
+     Python won't be able to use SSL.
```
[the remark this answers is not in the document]
juj: I must say I don't remember anymore - I may have tried to make emsdk work on Linux on ARM as well, and given that Raspberry Pi was the only ARM Linux device I hard around, I probably used that... and then banged my head against trying to figure out what's going wrong with it.

I can delete this though.

scripts/update_python.py
```diff
@@ -32,27 +33,35 @@
...
 
-pywin32_version = '227'
-pywin32_base = 'https://github.com/mhammond/pywin32/releases/download/b%s/' % pywin32_version
+PSUTIL = 'psutil==7.0.0'
 
 upload_base = 'gs://webassembly/emscripten-releases-builds/deps/'
 
 
+# Detects whether current python interpreter architecture is ARM64 or AMD64
+# If running AMD64 python on an ARM64 Windows, this still intentionally returns AMD64
```
sbc100, line 48: I'm not sure I understand the second part of this comment.

Do we really want to report `AMD64` if we are running amd64 python on arm64 windows?

sbc100: I see, so effectively this script cannot cross build the arm64 python package.

sbc100:
I don't understand how that CI can be passing here...

juj:
Oh, I needed to bump mininum macOS version from 10.13 to 11.0, or ctypes would no longer compile on x86 Apple Macs.

10.13 macOS supported:

Laptops: [MacBook](https://en.wikipedia.org/wiki/MacBook) (Late 2009 or later), [MacBook Air](https://en.wikipedia.org/wiki/MacBook_Air) (Late 2010 or later), [MacBook Pro](https://en.wikipedia.org/wiki/MacBook_Pro) (Mid 2010 or later)
    Desktops: [Mac Mini](https://en.wikipedia.org/wiki/Mac_Mini) (Mid 2010 or later), [iMac](https://en.wikipedia.org/wiki/IMac) (Late 2009 or later), [iMac Pro](https://en.wikipedia.org/wiki/IMac_Pro) (2017) (macOS 10.13), [Mac Pro](https://en.wikipedia.org/wiki/Mac_Pro) (Mid 2010 or later)

Whereas 11.0 macOS requires at minimum:

Laptops: [MacBook](https://en.wikipedia.org/wiki/MacBook) (Early 2015 or later), [MacBook Air](https://en.wikipedia.org/wiki/MacBook_Air) (Mid 2013 or later), [MacBook Pro](https://en.wikipedia.org/wiki/MacBook_Pro) (Late 2013 or later)
    Desktops: [Mac Mini](https://en.wikipedia.org/wiki/Mac_Mini) (Late 2014 or later), [iMac](https://en.wikipedia.org/wiki/IMac) (Mid 2014 or later), [iMac Pro](https://en.wikipedia.org/wiki/IMac_Pro) (2017), [Mac Pro](https://en.wikipedia.org/wiki/Mac_Pro) (Late 2013 or later)

At Unity we currently require macOS 11.0 as the minimum target for macOS for both Arm and x64 Macs. (macOS 11.0 was the first macOS version that shipped with Arm support)

sbc100:
> > I don't understand how that CI can be passing here...
> 
> Does the CI run the update_python.py script?

No, but I would expect them to try to download the latest SDK which should now be pointing a non-existent (at least not yet) python archives.

Does `./emsdk install latest` install your new python?

emsdk_manifest.json
```diff
@@ -350,13 +388,13 @@
...
     "bitness": 64,
-    "uses": ["python-3.9.2-nuget-64bit", "llvm-git-main-64bit", "node-20.18.0-64bit", "emscripten-main-64bit", "binaryen-main-64bit"],
+    "uses": ["python-3.13.3-64bit", "llvm-git-main-64bit", "node-20.18.0-64bit", "emscripten-main-64bit", "binaryen-main-64bit"],
     "os": "win"
   },
   {
     "version": "main",
     "bitness": 64,
-    "uses": ["python-3.9.2-64bit", "llvm-git-main-64bit", "node-20.18.0-64bit", "emscripten-main-64bit", "binaryen-main-64bit"],
+    "uses": ["python-3.13.3-64bit", "llvm-git-main-64bit", "node-20.18.0-64bit", "emscripten-main-64bit", "binaryen-main-64bit"],
```
sbc100, line 397: Ah, these are just the `main` SDKs i think you need to also update the `"releases-%releases-tag%` SDKs below.  Then CI should fail.

juj: That makes sense. Updated releases now python-3.13.3-64bit.

mcbarton:
After this is in can I access this change by using emsdk install latest, or do I need to wait for the next release? I recently added a Windows x86 Emscripten build of a project called CppInterOp, and am interested in extending it to build on Windows arm too (mostly just to give developers the opportunity to build the Emscripten build from any platform they might have access too)

juj:
> Is there some where we need to update the new macOS min version? Otherwise this looks ready to land.

Updated CMake configure step to also target macOS 11.0 at minimum.

emsdk_manifest.json
```diff
@@ -350,13 +388,13 @@
   {
     "version": "main",
     "bitness": 64,
-    "uses": ["python-3.9.2-nuget-64bit", "llvm-git-main-64bit", "node-20.18.0-64bit", "emscripten-main-64bit", "binaryen-main-64bit"],
```
sbc100, line 353: I think you can also delete the python-3.9.2 entries in this file

sbc100 (approved):
lgtm % comments

README.md
```diff
@@ -65,7 +65,7 @@ https://emscripten.org/docs/building_from_source/toolchain_what_is_needed.html.
 
 ### Linux
 
-- `python`: Version 3.9.2 or above.
+- `python`: Version 3.9.2 or above. Version 3.13.3 or a newer 3.13.x is recommended.
```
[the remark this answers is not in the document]
juj: ok

emsdk_manifest.json
juj: If we did that, then people cannot manually choose to install the previous python 3.9.2 to e.g. compare if updating python to 3.13.3 was the root cause of a regression.

i.e. with retaining that, users can still install combinations of the individual tools that they like.

README.md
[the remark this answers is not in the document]
juj: Err, forgot to push before merge. https://github.com/emscripten-core/emsdk/pull/1562

PR #1497, opened 2024-12-06
Python lint: Upgrade to flake8=7.1.1

cclauss:
https://pypi.org/project/flake8
```diff
-     - run: python3 -m pip install flake8==3.9.2
+     - run: python3 -m pip install flake8==7.1.1
```

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,24 +1,67 @@
 [flake8]
-ignore =
-   E111, # Indentation is not a multiple of four
-   E114, # Indentation is not a multiple of four (comment)
-   E501, # Line too long
-   E121, # Continuation line under-indented for hanging indent
-   E722  # bare excepts
-   E741, # Variable names such as 'l', 'O', or 'I'
+   # E111: Indentation is not a multiple of four
+   # E114: Indentation is not a multiple of four (comment)
+   # E121: Continuation line under-indented for hanging indent
+   # E122: continuation line missing indentation or outdented
+   # E126: continuation line over-indented for hanging indent
+   # E127: continuation line over-indented for visual indent
+   # E128: continuation line under-indented for visual indent
+   # E201: whitespace after '['
+   # E202: whitespace before ']'
+   # E203: whitespace before ':'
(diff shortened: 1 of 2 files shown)


.flake8
```diff
@@ -1,24 +1,67 @@
...
   ./git,
+  ./gnu,
+  ./llvm,
   ./node,
   ./python,
+  ./releases,
   ./temp,
-  ./downloads,
-  ./crunch,
+  ./upstream,
```
sbc100, line 67: Is this list just being sorted, or are there changes here too?

.flake8
```diff
@@ -1,24 +1,67 @@
...
+   # F821: undefined name 'basestring'
+   # F841: local variable 'foos' is assigned to but never used
+   # W291: trailing whitespace
+   # W391: blank line at end of file
+   # W503: line break before binary operator
+   # W504: line break after binary operator
+ignore = E111, E114, E121, E122, E126, E127, E128, E201, E202, E203, E211, E221, E226,
+   E228, E231, E241, E251, E261, E262, E266, E271, E301, E302, E303, E305, E306, E402,
+   E501, E704, E713, E721, E722, E731, E741, F401, F403, F405, F541, F811, F821, F841,
+   W291, W391, W503, W504
```
sbc100, line 50: Seems a little harder to maintain than the old method since now we have to keep the list in the comments in sync with the list of numbers here.

With the old method the comment and the error code are and the same line together as one.

sbc100 (approved):
lgtm % comments

.flake8
cclauss: Sorted (only) to make it easy to spot missing items and make it nearly impossible to add duplicates.

.flake8
sbc100: So how does the existing code work?  Is this something that used to work but no longer does?

cclauss: These must be new lint rules added between v3 (2021) and v7 (2024).
* https://flake8.pycqa.org/en/latest/release-notes

PR #1498, opened 2024-12-06
Python lint: Use flake8 --extend-ignore instead of --ignore

cclauss:
Related to
* #1497 

After re-reading the [flake8 configuration docs](https://flake8.pycqa.org/en/latest/user/configuration.html), I remembered that using `flake8 --extend-ignore` instead of `--ignore` greatly reduces the number of active lint rules.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,55 +1,6 @@
 [flake8]
-   # E111: Indentation is not a multiple of four
-   # E114: Indentation is not a multiple of four (comment)
-   # E121: Continuation line under-indented for hanging indent
-   # E122: continuation line missing indentation or outdented
-   # E126: continuation line over-indented for hanging indent
-   # E127: continuation line over-indented for visual indent
-   # E128: continuation line under-indented for visual indent
-   # E201: whitespace after '['
-   # E202: whitespace before ']'
-   # E203: whitespace before ':'
-   # E211: whitespace before '('
-   # E221: multiple spaces before operator
-   # E226: missing whitespace around arithmetic operator
-   # E228: missing whitespace around modulo operator
-   # E231: missing whitespace after ','
-   # E241: multiple spaces after ','
-   # E251: unexpected spaces around keyword / parameter equals
-   # E261: at least two spaces before inline comment
-   # E262: inline comment should start with '# '
-   # E266: too many leading '#' for block comment
-   # E271: multiple spaces after keyword
-   # E301: expected 1 blank line, found 0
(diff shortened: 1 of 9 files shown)


emsdk.py
```diff
@@ -784,7 +784,7 @@ def GIT(must_succeed=True):
       if ret == 0:
         cached_git_executable = git
         return git
-    except:
+    except Exception:
```
sbc100, line 787: When we are trying to catch everything, why not use the shorter form?

cclauss: % [`ruff rule E722`](https://docs.astral.sh/ruff/rules/bare-except)
## Why is this bad?
A bare `except` catches `BaseException` which includes `KeyboardInterrupt`, `SystemExit`, `Exception`, and others. Catching `BaseException` can make it hard to interrupt the program (e.g., with Ctrl-C) and can disguise other problems.

---

https://peps.python.org/pep-0008/#programming-recommendations
> When catching exceptions, mention specific exceptions whenever possible instead of using a bare except: clause:
```python
try:
    import platform_specific_module
except ImportError:
    platform_specific_module = None
```
> A bare except: clause will catch SystemExit and KeyboardInterrupt exceptions, making it harder to interrupt a program with Control-C, and can disguise other problems. If you want to catch all exceptions that signal program errors, use except Exception: (bare except is equivalent to except BaseException:).

> A good rule of thumb is to limit use of bare ‘except’ clauses to two cases:
> * If the exception handler will be printing out or logging the traceback; at least the user will be aware that an error has occurred.
> * If the code needs to do some cleanup work, but then lets the exception propagate upwards with raise. try...finally can be a better way to handle this case.

sbc100: Ah yes we don't want to catch KeyboardInterrupt here I guess

.flake8
```diff
@@ -1,55 +1,6 @@
...
-   # W391: blank line at end of file
-   # W503: line break before binary operator
-   # W504: line break after binary operator
-ignore = E111, E114, E121, E122, E126, E127, E128, E201, E202, E203, E211, E221, E226,
-   E228, E231, E241, E251, E261, E262, E266, E271, E301, E302, E303, E305, E306, E402,
-   E501, E704, E713, E721, E722, E731, E741, F401, F403, F405, F541, F811, F821, F841,
-   W291, W391, W503, W504
-
-exclude =
+extend-exclude =
```
sbc100, line 2: This wasn't needed before?

sbc100 (approved):

PR #1499, opened 2024-12-07
pyupgrade: Upgrade syntax for newer versions of Python

cclauss:
https://github.com/asottile/pyupgrade

% `ruff check --select=RUF010,UP --ignore=UP015,UP031 --fix --unsafe-fixes`

% `git grep is_string`  # to ensure it was only used once in the codebase.

f-strings are shorter, faster, and easier to read.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -4,7 +4,6 @@
 # University of Illinois/NCSA Open Source License.  Both these licenses can be
 # found in the LICENSE file.
 
-from __future__ import print_function
 
 import copy
 from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


emsdk.py
```diff
@@ -1038,8 +1033,7 @@ def quote_parens(x):
 def xcode_sdk_version():
   try:
     output = subprocess.check_output(['xcrun', '--show-sdk-version'])
-    if sys.version_info >= (3,):
-      output = output.decode('utf8')
+    output = output.decode('utf8')
     return output.strip().split('.')
```
cclauss, line 1037: Possible alternative:
```suggestion
    return output.decode('utf8').strip().split('.')
```

emsdk.py
```diff
@@ -29,13 +28,9 @@
     import _winreg as winreg
   import ctypes.wintypes
 
-if sys.version_info >= (3,):
-  from urllib.parse import urljoin
-  from urllib.request import urlopen
-  import functools
-else:
-  from urlparse import urljoin
-  from urllib2 import urlopen
```
sbc100, line 38: This one script in particular we try to keep running on python2.  The reason is that this script actually downloads the version of python3 that we use in emsdk.   So its good to be able to bootstrap the system on systems that have python2 install by default (like older versions of macOS apparently).

Issue #1509, opened 2025-01-09
Is there a way to use emsdk using bzlmod?


ShapeLayer:
I am now trying to use `emsdk/bazel` with bzlmod. However, the contents of `/bazel` seem to not support bzlmod. In more detail, I'm trying to use emsdk using only `MODULE.bazel`, but it is not working well.

Is there a way to use emsdk with `MODULE.bazel` not `WORKSPACE`?

Ref:
- [Bzl Migration Guide](https://bazel.build/external/migration)
- [WORKSPACE files](https://bazel.build/rules/lib/globals/workspace) - _Available APIs for WORKSPACE_

```
ERROR: Skipping '//includes:libgfm2pst-wasm.js': error loading package 'includes': Unable to find package for
 @@[unknown repo 'emsdk' requested from @@]//emscripten_toolchain:wasm_rules.bzl: The repository '@@
[unknown repo 'emsdk' requested from @@]' could not be resolved: No repository visible as '@emsdk' from 
main repository. Was the repository introduced in WORKSPACE? The WORKSPACE file is disabled by default in 
Bazel 8 (late 2024) and will be removed in Bazel 9 (late 2025), please migrate to Bzlmod. See https://bazel.build/external/migration.
ERROR: error loading package 'includes': Unable to find package for @@[unknown repo 'emsdk' requested from 
@@]//emscripten_toolchain:wasm_rules.bzl: The repository '@@[unknown repo 'emsdk' requested from @@]' 
could not be resolved: No repository visible as '@emsdk' from main repository. Was the repository introduced in 
WORKSPACE? The WORKSPACE file is disabled by default in Bazel 8 (late 2024) and will be removed in Bazel 9 
(late 2025), please migrate to Bzlmod. See https://bazel.build/external/migration.
```

sbc100:
@walkingeyerobot ?

ShapeLayer:
Add: I want to know if there is an official manual, procedure, or works in progress about migration to bzlmod.

_I've been trying to migrate, and it seems to be increasing works. So I want to check other answers before going any further._

sbc100:
No plans that I know off.   I have no idea what bzlmod even is though.   

The bazel support for emsdk it mostly a best effort by interested parties so if you would like to contribute please go ahead.

cburchert:
[...]

PR #1518, opened 2025-02-02
Bazel: Pass nodejs binary path as an environment variable.

cburchert:
The current way to derive the binary path relies on a specific name for the nodejs repository. This blocks migrating to bzlmod, as bzlmod prefixes repository directories with the module name that created them. By asking bazel for the path instead we always get the correct path, so we can work with both bzlmod and WORKSPACE based dependencies at the same time.

The repository @nodejs, used in the build label, refers to nodejs for the host platform and is generated by the following macro (wasn't too obvious to me):
https://github.com/bazel-contrib/rules_nodejs/blob/d19d695275c5fd0d5224ddc7826d9f7f4b8186df/nodejs/repositories.bzl#L452

This is some work towards solving #1509.

### DIFF
--- a/bazel/emscripten_toolchain/toolchain.bzl
+++ b/bazel/emscripten_toolchain/toolchain.bzl
@@ -71,6 +71,7 @@ def _impl(ctx):
     cc_target_os = "emscripten"
 
     emscripten_dir = ctx.attr.emscripten_binaries.label.workspace_root
+    nodejs_path = ctx.file.nodejs_bin.path
 
     builtin_sysroot = emscripten_dir + "/emscripten/cache/sysroot"
 
@@ -1060,6 +1061,10 @@ def _impl(ctx):
                     key = "EM_CONFIG_PATH",
                     value = ctx.file.em_config.path,
                 ),
+                env_entry(
+                    key = "NODE_JS_PATH",
+                    value = nodejs_path,
+                ),
             ],
         ),
         # Use llvm backend.  Off by default, enabled via --features=llvm_backend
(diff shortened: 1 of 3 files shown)


bazel/emscripten_toolchain/toolchain.bzl
```diff
@@ -1060,6 +1061,10 @@ def _impl(ctx):
                     key = "EM_CONFIG_PATH",
                     value = ctx.file.em_config.path,
                 ),
+                env_entry(
+                    key = "NODE_JS_PATH",
+                    value = nodejs_path,
```
sbc100, line 1066: If this is is an absolute path (or can be) then you can just pass `EM_NODE_JS` and completely remove the line from the config file.

cburchert: I looked a bit more into this, but couldn't find a way to do this much better. When this code here is executed, then the absolute path is not known yet, as we aren't yet executing actions, which run in sandboxes. I looked for some variable that will get substituted later, but couldn't find any.

There is env.sh, which is called before we end up in the config and it could do the conversion to an absolute path, but that doesn't really save any work. If we do it in env.sh, it does not work on Windows, which has its own env.bat, so the conversion forks into two places.

cburchert:
Ok this seems to have fixed the Windows issue. I don't fully understand why this broke the closure compiler call and it wasn't broken before, but I guess having the same workaround on all platforms at least is better.

walkingeyerobot (approved):

sbc100:
lgtm if @walkingeyerobot is happy.  Are you able to hit the merge button @walkingeyerobot ?

PR #1530, opened 2025-02-11
Bazel: Migrate to bzlmod

### LINKED ISSUE #1509: Is there a way to use emsdk using bzlmod?

cburchert:
Note: This is based on https://github.com/emscripten-core/emsdk/pull/1518, so that needs to be merged first.

This migrates the bazel integration to the new Bazel dependency system "bzlmod". bzlmod is becoming mandatory this year (see second sentence here: https://bazel.build/external/migration).

This is a backwards incompatible migration, directly removing the old WORKSPACE based approach. Users will have to change how they depend on bzlmod, however I assume pretty much every user will be happy about it, because they are forced to use bzlmod anyway or add extra flags to continue to build with newer Bazel versions. Given that users normally depend on specific git commits in the old system, they won't be hit with this until they decide to upgrade emsdk.

The basic principle here is simple: I took everything that WORKSPACE did and searched an alternative in bzlmod. Some interesting bits:

- We have less worries about multiple versions and people depending on emscripten multiple times in different ways. This is resolved by the new system: Bazel first iterates through the MODULE.bazel files recursively, then finds the minimum version needed for each module and then executes the module extensions that define repos exactly once at that version. So no more ifs needed to detect multiple inclusions.
- A bunch of nodejs stuff moves to MODULE.bazel, because that is how the nodejs module works now. As their module extension gets executed only once you need to declare everything that you could need before that in the MODULE.bazel file. A side effect of that is that we have to make a fake repository when emscripten doesn't have an arm64 binary for linux, because we can't actually figure that out in MODULE.bazel, so we have to declare that it always exists and then create one in all cases.

There is a bunch of autoformatter changes in here as well, I could try to revert them if you prefer.

Closes #1509


bazel/emscripten_deps.bzl
```diff
@@ -98,113 +100,85 @@ filegroup(
...
 
-def emscripten_deps(emscripten_version = "latest"):
-    version = emscripten_version
+def _emscripten_deps_impl(ctx):
+    version = None
+
+    for mod in ctx.modules:
+        for config in mod.tags.config:
+            if config.version and version != None:
+                fail("More than one emscripten version specified!")
```
cburchert, line 109: Not sure about this one. We could also pick the maximum version that was specified.

.circleci/config.yml
```diff
@@ -262,7 +262,7 @@ jobs:
     environment:
       PYTHONUNBUFFERED: "1"
       EMSDK_NOTTY: "1"
-      USE_BAZEL_VERSION: "5.4.0"
+      USE_BAZEL_VERSION: "7.4.1"
```
walkingeyerobot, line 265: if we can switch this to using the latest bazel instead, that would be ideal. then when bazel inevitably introduces a new feature that breaks us, we'll be notified.

bazel/emscripten_deps.bzl
walkingeyerobot: failing here seems fine.

walkingeyerobot (changes requested):
Overall this looks awesome, thanks very much! Just need the Windows test to pass, and if possible let's use the latest bazel instead of pinning to a slightly newer version than what we have. Thanks again!

.circleci/config.yml
trzeciak: While analyzing something, it occurred to me that it would be nice if the tests were run on at least two major versions of bazel, for today it would be latest 8.x and latest 7.x (8.1.0 and 7.5.0), for all platforms.

bazel/README.md
trzeciak, line 1: I think it would be nice to add information in the readme that emsdk no longer supports workspace, along with a link to the commit/version where workspace support was last available.

grefab:
Hey, I see that your test-bazel-linux workflow fails for the same reason my did. This was the solution: https://github.com/bazelbuild/rules_python/pull/713#issuecomment-1885628496

jjudd:
I was in the process of migrating a large repository to bzlmod when I came across this PR. Big thanks for all the work done on it so far! Super helpful.

I tried using it, ran into a few issues, and then fixed those issues. I'll do a bit more testing + polishing and if things still look good, I'll open a PR to this PR.

Most of the trouble had to do with the fact that the emscripten rules were `select`ing files and only setting up one toolchain. This caused troubles with missing files among other things.

I moved to instead setting up multiple toolchains, using `exec_compatible_with`, and letting Bazel do its thing.

The `test_bazel.sh` script passes and our targets that use emscripten seem to build successfully, so hopefully I'll have a PR open here pretty soon.

jjudd:
PR is up here: https://github.com/cburchert/emsdk/pull/1

Happy to respond to any feedback y'all provide.

jjudd:
I'm interested in doing some work that depends on this PR being merged.

In an effort to help get this merged I believe I've fixed the CI errors, rebased & got things working post-rebase.

I'm going to open another draft PR to confirm, so I can get CI to run in order to confirm that it is fixed. If it is, I'll open another PR to this PR and hopefully at that point it should be ready to merge.

jjudd:
@cburchert I've got a branch that is rebased on main and has builds passing. I opened a PR to your branch here: https://github.com/cburchert/emsdk/pull/2

The PR is funky because of the rebase. It may be easier to just rebase your branch and cherry pick the last 4 commits on that branch. Whatever is easiest for you.

At that point I think this will be ready for review & merging.

cburchert:
Awesome! I had some trouble resolving the build failures, especially the Windows one, as I don't have access to a Windows system. Then also other work distracted me and I lost track a bit, so I am really happy that you picked this up!

I will see if I can slightly improve the README somewhere today or tomorrow, otherwise this would probably be ready. It has grown a lot, so if the repository owners want this, we could also try to split into a few smaller pull requests. Especially the autoformatter fixes and bazel lock files clutter the diffs here and are probably easy to separate if wished.

jjudd:
@trzeciak @walkingeyerobot builds/checks are now passing on this PR. Any chance y'all could please take a look through it as I believe it should now be ready for review/merging.

.circleci/config.yml
cburchert: Added this now.

bazel/README.md
cburchert: Done.

walkingeyerobot (approved):

bazel/revisions.bzl
```diff
@@ -5,174 +5,199 @@ EMSCRIPTEN_TAGS = {
...
         sha_linux = "b188249ecb939dadc679aaf2d3d9afd0fe19ab942f91b7bc926b4f252915dd1a",
+        sha_linux_arm64 = "4aedc8ca641b40d9bd82d85b1dc3458fe1afc9a132da06a09384a5f89c058969",
         sha_mac = "2092aa4bef3b9f88d3f343b042a417ba617d4e04454656d8f2e101ba53f854e8",
         sha_mac_arm64 = "7a9a15845257629b7602d15bdf7633a8e10472b0fa9b3d9ee7149938aa2f2039",
         sha_win = "9fe76b6189566d56f0cf9aecbd23a006778530aa87184a900f5662e39ce7272a",
     ),
     "3.1.57": struct(
         hash = "523b29e1b99a61069a2fa9f9d3cc9be1c4c53d4d",
         sha_linux = "5bc444132258d4404d396f2044a4a334064ad0f1022555cad5ec72804a98ba5a",
+        sha_linux_arm64 = "f0022413afcc1610deff10921b3f5938bf4d01eba46ce96655f2295bdd84bd6a",
```
sbc100, line 200: Are these `sha_linux_arm64` related?   How did you generate these?   Assuming they are not related perhaps they could go in their own PR?

bazel/MODULE.bazel
```diff
@@ -1 +1,79 @@
-bazel_dep(name = "platforms", version = "0.0.9")
+module(
+    name = "emsdk",
+    version = "4.0.6",
```
sbc100, line 3: It looks like this version will need to be updated for each release? 

Does this also mean that only one version of the emsdk module can be used with any given commit in the emsdk repo?  i.e. I can't checkout tip-of-tree and use/install and older version?

bazel/test_secondary_lto_cache/MODULE.bazel.lock
```diff
@@ -0,0 +1,3710 @@
+{
```
sbc100, line 1: Does it make sense to check in these large lock files into source control given that they are just tests?  I guess so?

I wonder if were should move these test directories under the top level `test/` (not has part of this PR).

bazel/remote_emscripten_repository.bzl
```diff
@@ -0,0 +1,148 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load(":emscripten_build_file.bzl", "EMSCRIPTEN_BUILD_FILE_CONTENT_TEMPLATE")
+load(":revisions.bzl", "EMSCRIPTEN_TAGS")
+load("//emscripten_toolchain:toolchain.bzl", "emscripten_cc_toolchain_config_rule")
+
+def remote_emscripten_repository(
```
sbc100, line 6: Is "repository" the right name here?  Maybe this has some bazel-specific meaning?   Perhaps something like "external_emscripten_tools" or "external_emsdk_archive"?   I'm probably not best placed to know what the best option here is, but "repository" here makes me thing of a git reprository which this is not (right?).

bazel/revisions.bzl
jjudd: They are related to the toolchain `"//emscripten_toolchain:cc-toolchain-wasm-emscripten_linux_arm64"` that supports arm64.

I generated these by modifying the `scripts/update_bazel_workspace.py` file and re-running it for all versions after the emscripten added arm64 support: https://github.com/emscripten-core/emsdk/issues/547#issuecomment-2052449214

Modifications to the Python file are here: https://github.com/emscripten-core/emsdk/pull/1530/files#diff-5cfef6f2e098de6c8778bf82600a38ae35b43fc44b414dcb920ab2fbbd40dfc3R45

I believe they are relevant to this PR because if you want to use the ruleset on arm and these aren't defined, then things will break.

sbc100: I see. So this change adds linux arm64 support at the same time as converting to bzlmod?  Its not reasonable to land linux arm64 support first?

bazel/remote_emscripten_repository.bzl
jjudd: It's Bazel specific term. I named it `repository` because it currently creates a Bazel repository.

https://bazel.build/concepts/build-ref#repositories

> Source files used in a Bazel build are organized in repositories (often shortened to repos). A repo is a directory tree with a boundary marker file at its root; such a boundary marker file could be MODULE.bazel, REPO.bazel, or in legacy contexts, WORKSPACE or WORKSPACE.bazel.
>
> The repo in which the current Bazel command is being run is called the main repo. Other, (external) repos are defined by repo rules; see [external dependencies overview](https://bazel.build/external/overview) for more information.

http_archive is a `repository_rule` and the emcripten repositories created are referred to as repositories elsewhere in the ruleset. For example, `@emscripten_bin_mac_arm64//...`

sbc100: I see.  Thanks for the explanation.  Naming is hard I guess.

bazel/test_secondary_lto_cache/MODULE.bazel.lock
sbc100: Lets do that as a followup then.  Thanks.

bazel/MODULE.bazel
jjudd: AFAIK you can use whatever version you want by doing the following in the `MODULE.bazel` file:

```
emscripten_deps = use_extension(
    "@emsdk//:emscripten_deps.bzl",
    "emscripten_deps",
)
emscripten_deps.config(version = "4.0.1")
```

Meaning if you're using module version `4.0.6` you could use up to emsdk version `4.0.6`. 

I just tested using emsdk 4.0.3 with module 4.0.6 and it worked in our main repo.

If you want to create support for a newer version of emsdk or a custom version of it, then you could call `remote_emscripten_repository` directly with `create_toolchains` and `register_toolchains`, and in theory that should also work.

I haven't tested that use case, though.

bazel/revisions.bzl
jjudd: I believe the rule set currently supports arm64, or at least tries to: https://github.com/emscripten-core/emsdk/blob/24fc909c0da13ef641d5ae75e89b5a97f25e37aa/bazel/BUILD#L11-L17

I'm not confident it was working, which is why I made these changes when updating things to use toolchains.

I believe it should be fine to not include these, as I've created a fallback to an empty repository in `emcripten_deps.bzl`:

```
    # Not all versions have a linux/arm64 release: https://github.com/emscripten-core/emsdk/issues/547
    if hasattr(revision, "sha_linux_arm64"):
        remote_emscripten_repository(
            name = emscripten_repo_name("linux_arm64"),
            bin_extension = "",
            sha256 = revision.sha_linux_arm64,
            strip_prefix = "install",
            type = "tar.xz",
            url = emscripten_url.format("linux", revision.hash, "-arm64", "tar.xz"),
        )
    else:
        _empty_repository(
            name = emscripten_repo_name("linux_arm64"),
        )
```

That said, considering the rule set currently has config for arm64, I figure including it as part of this PR won't hurt anything.

bazel/MODULE.bazel
sbc100: Perhaps we should not tie this version number to the emsdk version number then?  Maybe it should have its own unrelated version.. to avoid confusing these two unrelated versions?

bazel/emscripten_toolchain/wasm_cc_binary.bzl
```diff
@@ -32,12 +32,13 @@ def _wasm_transition_impl(settings, attr):
 
     return {
         "//command_line_option:compiler": "emscripten",
-        "//command_line_option:crosstool_top": "@emsdk//emscripten_toolchain:everything",
         "//command_line_option:cpu": "wasm",
         "//command_line_option:features": features,
         "//command_line_option:dynamic_mode": "off",
         "//command_line_option:linkopt": linkopts,
         "//command_line_option:platforms": [platform],
+        # This is hardcoded to an empty cc_library because dlmalloc.bc
```
sbc100, line 40: The `.bc` extension is no longer used by emscripten.  Perhaps just call say "... because the malloc library is implicitly added..." ?

jjudd:
@cburchert this line was added to `test_bazel.sh`: `grep ${VER} bazel/MODULE.bazel || (echo ${FAILMSG} && false)` should it also be added to `test_bazel_mac.sh`?

bazel/MODULE.bazel
jjudd: I'm torn. I like the idea of versioning schemes being different because they have different lifecycles. At the same time this check in `test_bazel.sh` is an easy way to make sure the module version gets bumped when a new version of emsdk is added: `grep ${VER} bazel/MODULE.bazel || (echo ${FAILMSG} && false)`

IMHO, ideally the versioning schemes would be separated and there would be some kind of check to make sure the module version gets bumped when adding a new emsdk version.

bazel/emscripten_toolchain/wasm_cc_binary.bzl
jjudd: Makes sense to me.

```suggestion
        # This is hardcoded to an empty cc_library because the malloc library
```

bazel/MODULE.bazel
cburchert: I implemented it this way to keep the current behavior, which also has separate versions for emsdk and the emscripten toolchain used.

There is already a check that the bazel module version is updated when adding a new emscripten release. It is the following line in test/test_bazel.sh. :
```
grep ${VER} bazel/MODULE.bazel || (echo ${FAILMSG} && false)
``

Note that scripts/update_bazel_workspace.py also implements automatically updating the module version, so if that script is used to update the repository the version number will also automatically get bumped.

**Comment by cburchert** [comment]
> @cburchert this line was added to `test_bazel.sh`: `grep ${VER} bazel/MODULE.bazel || (echo ${FAILMSG} && false)` should it also be added to `test_bazel_mac.sh`?

Makes sense, added.

**Comment by cburchert** [comment]
Regarding the README, I already added the part about the last release that supported WORKSPACE file based inclusion. I read through another time and everything seems correct to me.

File: bazel/MODULE.bazel (hunk 4 above)
Comment (sbc100) [reply bazel/MODULE.bazel:3]: > I implemented it this way to keep the current behavior, which also has separate versions for emsdk and the emscripten toolchain used.

But this seems like a change in behaviour:

old: The bazel toolchain version is unrelated to any emsdk version (0.0.9)
new: The bazel toolchain version matches exactly the latest emsdk version (4.0.6)

I'm not saying that new way is wrong, just the it seems like a departure from the old behaviour.

I think I'm find with it as is, but we might also want de-emphasis the ability to install older versions.. since it seems a little odd or use emsdk bazel version 4.0.3 to install emsdk version 4.0.1.  (e.g. the new scheme makes me think they should match more than the old scheme).

File: bazel/README.md [hunk 8]
```diff
@@ -2,36 +2,30 @@
...
-git_repository(
-    name = "emsdk",
+emsdk_version = "4.0.6"
+bazel_dep(name = "emsdk", version = emsdk_version)
+git_override(
+    module_name = "emsdk",
     remote = "https://github.com/emscripten-core/emsdk.git",
-    tag = "3.1.64",
     strip_prefix = "bazel",
+    tag = emsdk_version,
```
sbc100, line 15: Can/should we just recommend using main there (i.e. drop the `tag = `?).

BTW why do we need to specify the `bazel_dep` *and* the git repository?  Doesn't the former imply the later somehow via a registry?

Also, how does `bezel_dep` find `emsdk` if it comes before the `git_override` the provides it?

bazel/README.md
```diff
@@ -2,36 +2,30 @@
...
-parameter of `git_repository` and `emsdk_emscripten_deps()`. Supported versions are listed in `revisions.bzl`
-
-Bazel 7+ additionally requires `platforms` dependencies in the `MODULE.bazel` file.
-```starlark
-bazel_dep(name = "platforms", version = "0.0.9")
+emscripten_deps = use_extension(
+    "@emsdk//:emscripten_deps.bzl",
+    "emscripten_deps",
+)
+emscripten_deps.config(version = "4.0.1")
```
sbc100, line 26: If I want the default version should I just delete this line?  If this line is optional perhaps it should not be included in the default instructions, but just in side node.

bazel/README.md
cburchert: bazel_dep would by default go to the bazel central registry, however emsdk isn't registered there. We should fix that, however we first require this pull request to be merged, as you can only have bazel modules in the registry. As long as it is not in the registry, you need the git_override to stop the registry request via pulling from github directly.

So the solution is:

1. Merge this pull request.
2. Add emsdk to the bazel central registry.
3. Remove this part from the readme here.

Regarding the version: We need something stable, i.e. a commit hash or a tag. I don't think we can put main here or we will produce at least a warning. Bazel rebuilds everything that changed or that depends on something changed, so pointing to changing remote dependencies is a pretty bad idea for your build caches.

bazel/README.md
cburchert: The line right before this code block tells you to only include this, if you want to override to a different version than the emsdk version. Does that address the question?

bazel/MODULE.bazel
jjudd: If the `0.0.9` version you're referring to is the `bazel_dep(name = "platforms", version = "0.0.9")` then that is unrelated to the module version present in

```
module(
    name = "emsdk",
    version = "4.0.6",
)
```

`0.0.9` was the version of the platforms Bazel module that was being used. The source of which can be found here: https://github.com/bazelbuild/platforms

The other one is the version of the module defined by this rule set, i.e., the emsdk Bazel rules.

bazel/README.md
jjudd: To add on, in case it is helpful: `bazel_dep` says that the dependency is depended upon in this repo. If you don't specify an override, that dep is retrieved from the central Bazel registry: https://registry.bazel.build/

If you do specify an override, then that override defines where the dep is to be retrieved from.

Because emsdk is not published to the central registry, the override is necessary for the emsdk module to be depended upon.

bazel/MODULE.bazel
sbc100: Ah!  I see thanks for the explanation.

bazel/README.md
sbc100: sgtm

bazel/README.md
sbc100: Doh.  Thanks.

sbc100 (approved):

sbc100:
I can update the CI names once this lands.

@walkingeyerobot should we go for it?

Vertexwahn:
Awesome!

sbc100:
Looks like this caused some kind of issue with `update_bazel_workspace.py`.  See https://app.circleci.com/pipelines/github/emscripten-core/emsdk/2530/workflows/a2074f99-f6d8-4ce5-a0e2-69f57c836883/jobs/16369?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary, which happened as part of CI run for #1549

sbc100:
Found a fix: #1550

Issue #1479, opened 2024-10-23
[Bazel] Support for feature debug fission in emsdk-bazel-toolchain


trzeciak:
This is a feature request to add support for Debug Fission (dwarf-5 in separated `.dwo`/`.dwp` file), in emsdk-bazel-toolchain.

From similar reports I've learned that the emscripten team is not currently maintaining a toolchain for bazel, but maybe someone else has it partly ready?

The motivation for this fr is to improve the developer experience of debugging C++ applications in WebAssembly by reducing linking times and reducing the time needed to parse symbol files by devtools.

References:
- https://developer.chrome.com/blog/faster-wasm-debugging
- https://emscripten.org/docs/porting/Debugging.html#debugging-in-the-browser
- [bazel `--fission` flag](https://bazel.build/docs/user-manual#fission)
- https://gcc.gnu.org/wiki/DebugFission
- [clang `-gsplit-dwarf` flag](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang1-gsplit-dwarf)

PR #1531, opened 2025-02-19
[Bazel] Support for feature debug fission in emsdk-bazel-toolchain #1479

trzeciak:
Support for debug fission. In short, debugging fission should:
* reduce linking time, RAM usage and disk usage  
* speed up incremental builds
* speed up debugger work (reduce startup and breakpoint time)

To use this, follow the `--fission=yes` flag.

References:
* https://developer.chrome.com/blog/faster-wasm-debugging
* https://bazel.build/docs/user-manual#fission
* https://yurydelendik.github.io/webassembly-dwarf/
* https://www.tweag.io/blog/2023-11-23-debug-fission/#case-study-bazel

### DIFF
--- a/bazel/emscripten_deps.bzl
+++ b/bazel/emscripten_deps.bzl
@@ -96,6 +96,13 @@ filegroup(
         ],
     ),
 )
+
+filegroup(
+    name = "dwp_files",
+    srcs = [
+        "bin/llvm-dwp{bin_extension}",
+    ],
+)
 """
 
 def emscripten_deps(emscripten_version = "latest"):
--- a/bazel/emscripten_toolchain/BUILD.bazel
+++ b/bazel/emscripten_toolchain/BUILD.bazel
@@ -44,11 +44,24 @@ filegroup(
 )
 
 filegroup(
+    name = "dwp_files",
+    srcs = [
+        "emdwp-emscripten_bin_linux_arm64.sh",
+        "emdwp-emscripten_bin_linux.sh",
+        "emdwp-emscripten_bin_mac_arm64.sh",
+        "emdwp-emscripten_bin_mac.sh",
+        "emdwp-emscripten_bin_win.bat",
+        "@emsdk//:dwp_files",
+    ],
+)
+
+filegroup(
     name = "all_files",
     srcs = [
         ":ar_files",
         ":compiler_files",
         ":linker_files",
+        ":dwp_files",
     ],
 )
 
@@ -75,7 +88,7 @@ cc_toolchain(
     ar_files = ":ar_files",
     as_files = ":empty",
     compiler_files = ":compiler_files",
-    dwp_files = ":empty",
+    dwp_files = ":dwp_files",
     linker_files = ":linker_files",
     objcopy_files = ":empty",
     strip_files = ":empty",
--- a/bazel/emscripten_toolchain/emdwp-emscripten_bin_linux.sh
+++ b/bazel/emscripten_toolchain/emdwp-emscripten_bin_linux.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+exec external/emscripten_bin_linux/bin/llvm-dwp "$@"
--- a/bazel/emscripten_toolchain/emdwp-emscripten_bin_linux.sh
+++ b/bazel/emscripten_toolchain/emdwp-emscripten_bin_linux.sh
@@ -1,3 +1,10 @@
 #!/bin/bash
+#
+#  This script differs in form from emcc.{py,bat}/…, because bazel are limited/bugged in the way of executing dwp tool.
+#  Bazel dwp action configuration does not pass environment variables, so we cannot use them in this script.
+#  For more info, see PR discussion and bezel issue:
+#  - https://github.com/emscripten-core/emsdk/pull/1531#discussion_r1962090650
+#  - https://github.com/bazelbuild/bazel/issues/25336
+#
 
 exec external/emscripten_bin_linux/bin/llvm-dwp "$@"
--- a/bazel/emscripten_toolchain/emdwp-emscripten_bin_linux.sh
+++ b/bazel/emscripten_toolchain/emdwp-emscripten_bin_linux.sh
@@ -2,7 +2,7 @@
 #
 #  This script differs in form from emcc.{py,bat}/…, because bazel are limited/bugged in the way of executing dwp tool.
 #  Bazel dwp action configuration does not pass environment variables, so we cannot use them in this script.
-#  For more info, see PR discussion and bezel issue:
+#  For more info, see PR discussion and bazel issue:
 #  - https://github.com/emscripten-core/emsdk/pull/1531#discussion_r1962090650
 #  - https://github.com/bazelbuild/bazel/issues/25336
 #
--- a/bazel/emscripten_toolchain/wasm_binary.py
+++ b/bazel/emscripten_toolchain/wasm_binary.py
@@ -36,7 +36,7 @@ def main():
 
   if args.dwp_file:
     for idx, output in enumerate(args.outputs):
-      if output.endswith(".dwp"): # also update extension 'binary.dwp' to 'binary.wasm.dwp'
+      if output.endswith(".dwp"):  # also update extension 'binary.dwp' to 'binary.wasm.dwp'
         shutil.copy2(args.dwp_file, output)
         args.outputs.pop(idx)
         break
--- a/bazel/emscripten_toolchain/wasm_binary.py
+++ b/bazel/emscripten_toolchain/wasm_binary.py
@@ -14,7 +14,7 @@ WebAssembly binary into a larger web application.
 import argparse
 import os
 import tarfile
-
+import shutil
 
 def ensure(f):
   if not os.path.exists(f):
@@ -26,11 +26,20 @@ def main():
   parser = argparse.ArgumentParser()
   parser.add_argument('--archive', help='The archive to extract from.')
   parser.add_argument('--outputs', help='Comma separated list of files that should be extracted from the archive. Only the extname has to match a file in the archive.')
+  parser.add_argument('--dwp_file', help='Optional dwp input file, generated when fission flags set.')
   parser.add_argument('--allow_empty_outputs', help='If an output listed in --outputs does not exist, create it anyways.', action='store_true')
   args = parser.parse_args()
 
   args.archive = os.path.normpath(args.archive)
   args.outputs = args.outputs.split(",")
+  args.dwp_file = os.path.normpath(args.dwp_file) if args.dwp_file else None
+
+  if args.dwp_file:
+    for idx, output in enumerate(args.outputs):
+      if output.endswith(".dwp"): # also update extension 'binary.dwp' to 'binary.wasm.dwp'
+        shutil.copy2(args.dwp_file, output)
+        args.outputs.pop(idx)
+        break
 
   tar = tarfile.open(args.archive)
 
--- a/bazel/emscripten_toolchain/wasm_binary.py
+++ b/bazel/emscripten_toolchain/wasm_binary.py
@@ -16,6 +16,7 @@ import os
 import tarfile
 import shutil
 
+
 def ensure(f):
   if not os.path.exists(f):
     with open(f, 'w'):
--- a/bazel/emscripten_toolchain/wasm_cc_binary.bzl
+++ b/bazel/emscripten_toolchain/wasm_cc_binary.bzl
@@ -69,6 +69,7 @@ _ALLOW_OUTPUT_EXTNAMES = [
     ".fetch.js",
     ".js.symbols",
     ".wasm.debug.wasm",
+    ".wasm.dwp",
     ".html",
     ".aw.js",
 ]
@@ -107,10 +108,11 @@ _WASM_BINARY_COMMON_ATTRS = {
 }
 
 def _wasm_cc_binary_impl(ctx):
-    args = ctx.actions.args()
     cc_target = ctx.attr.cc_target[0]
+    dwp_file = cc_target[DebugPackageInfo].dwp_file if DebugPackageInfo in cc_target else None
+    outputs = ctx.outputs.outputs
 
-    for output in ctx.outputs.outputs:
+    for output in outputs:
         valid_extname = False
         for allowed_extname in _ALLOW_OUTPUT_EXTNAMES:
             if output.path.endswith(allowed_extname):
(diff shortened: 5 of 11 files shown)


trzeciak:
Can I sync bazel for windows (from old ones 5.4.0), to be sync with linux/mac?

bazel/emscripten_toolchain/BUILD.bazel
```diff
@@ -43,12 +43,25 @@ filegroup(
...
 )
 
+filegroup(
+    name = "dwp_files",
+    srcs = [
+        "emdwp-emscripten_bin_linux_arm64.sh",
+        "emdwp-emscripten_bin_linux.sh",
+        "emdwp-emscripten_bin_mac_arm64.sh",
+        "emdwp-emscripten_bin_mac.sh",
+        "emdwp-emscripten_bin_win.bat",
```
sbc100, line 53: Why does this tool require 5 files where the above tools all just use two (one .sh and one .bat)?

bazel/emscripten_deps.bzl
```diff
@@ -96,6 +96,13 @@ filegroup(
         ],
     ),
 )
+
+filegroup(
+    name = "dwp_files",
+    srcs = [
+        "bin/llvm-dwp{bin_extension}",
```
sbc100, line 103: Doesn't this need emdwp.py like the above uses emar.py?

bazel/emscripten_toolchain/emdwp-emscripten_bin_linux.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+exec external/emscripten_bin_linux/bin/llvm-dwp "$@"
```
sbc100, line 3: The emdwp.py wrapper should be able to find these files based on LLVM_ROOT, so I don't think you need 5 different wrappers here.

bazel/emscripten_toolchain/wasm_cc_binary.bzl
```diff
@@ -119,28 +121,35 @@ def _wasm_cc_binary_impl(ctx):
...
             fail("Invalid output '{}'. Allowed extnames: {}".format(output.basename, ", ".join(_ALLOW_OUTPUT_EXTNAMES)))
 
+    inputs = ctx.files.cc_target
+    args = ctx.actions.args()
     args.add_all("--archive", ctx.files.cc_target)
-    args.add_joined("--outputs", ctx.outputs.outputs, join_with = ",")
+    args.add_joined("--outputs", outputs, join_with = ",")
+
+    if dwp_file:
+        args.add("--dwp_file", dwp_file)
```
sbc100, line 130: How about `--dwp-file`?   At least for emscripten and emsdk we use the hyphen rather than underscore in the option name.

bazel/emscripten_toolchain/emdwp-emscripten_bin_linux.sh
trzeciak: I agree, but it doesn't work / or I don't know how to do it, https://github.com/emscripten-core/emsdk/pull/1531#discussion_r1961930494

bazel/emscripten_toolchain/wasm_cc_binary.bzl
trzeciak: I was guided by the `--allow_empty_outputs` option in the [`bazel/emscripten_toolchain/wasm_binary.py`](bazel/emscripten_toolchain/wasm_binary.py) file.

trzeciak: @sbc100 If you wish, I can change both.

bazel/emscripten_toolchain/BUILD.bazel
trzeciak: `printenv` from `dwp` tool:

```
...
SUBCOMMAND: # //:main [action 'CcGenerateDwp main.dwp', configuration: 77bdbcfa4ce6db4cf8539adb5465d4c7ee5d7734a59bd2c6e99a60f4e1c191e4, execution platform: @@platforms//host:host, mnemonic: CcGenerateDwp]
(cd /private/var/tmp/_bazel_damian/1c7f831e8ac7ba8a4146146efff137a8/execroot/_main && \
  exec env - \
  external/emsdk/emscripten_toolchain/emdwp-emscripten_bin_mac_arm64.sh bazel-out/wasm-dbg-ST-6bab6f2a10f7/bin/_objs/sdk/bar.dwo bazel-out/wasm-dbg-ST-6bab6f2a10f7/bin/_objs/sdk/foo.dwo bazel-out/wasm-dbg-ST-6bab6f2a10f7/bin/_objs/main/main.dwo -o bazel-out/wasm-dbg-ST-6bab6f2a10f7/bin/main.dwp)
# Configuration: 77bdbcfa4ce6db4cf8539adb5465d4c7ee5d7734a59bd2c6e99a60f4e1c191e4
# Execution platform: @@platforms//host:host
INFO: From CcGenerateDwp main.dwp:
+ printenv
TMPDIR=/tmp
__CF_USER_TEXT_ENCODING=0x1F5:0x0:0x0
PWD=/private/var/tmp/_bazel_damian/1c7f831e8ac7ba8a4146146efff137a8/sandbox/darwin-sandbox/30/execroot/_main
SHLVL=1
_=/usr/bin/printenv
+ exec external/emscripten_bin_mac_arm64/bin/llvm-dwp bazel-out/wasm-dbg-ST-6bab6f2a10f7/bin/_objs/sdk/bar.dwo bazel-out/wasm-dbg-ST-6bab6f2a10f7/bin/_objs/sdk/foo.dwo bazel-out/wasm-dbg-ST-6bab6f2a10f7/bin/_objs/main/main.dwo -o bazel-out/wasm-dbg-ST-6bab6f2a10f7/bin/main.dwp
…
```

trzeciak: I think the reason why environment variables are unavailable is in bazel itself.  
The action named CcGenerateDwp is declared here: [LINK](https://github.com/bazelbuild/bazel/blob/master/src/main/starlark/builtins_bzl/common/cc/cc_debug_helper.bzl#L76-L84)

>     ctx.actions.run(
>        mnemonic = "CcGenerateDwp",
>        tools = packager["tools"],
>        executable = packager["executable"],
>        toolchain = cc_helper.CPP_TOOLCHAIN_TYPE,
>        arguments = [packager["arguments"]],
>        inputs = packager["inputs"],
>        outputs = packager["outputs"],
>     )

As you can see, the `env` variable is not declared for this action.

For comparison, I found the `CppArchive` action where the `env` variable is set [LINK](https://github.com/bazelbuild/bazel/blob/master/src/main/starlark/builtins_bzl/common/cc/cc_import.bzl#L73-L95):


>        env = cc_common.get_environment_variables(
>            feature_configuration = feature_configuration,
>            action_name = CPP_LINK_STATIC_LIBRARY_ACTION_NAME,
>            variables = archiver_variables,
>        )
>
>        # TODO(bazel-team): PWD=/proc/self/cwd env var is missing, but it is present when an analogous archiving
>        # action is created by cc_library
>        ctx.actions.run(
>            executable = archiver_path,
>            toolchain = cc_helper.CPP_TOOLCHAIN_TYPE,
>            arguments = [args],
>            env = env,
>            inputs = depset(
>                direct = object_files,
>                transitive = [
>                    cc_toolchain.all_files,
>                ],
>            ),
>            use_default_shell_env = True,
>            outputs = [output_file],
>            mnemonic = "CppArchive",
>        )

----

So as you can see, on the master version (>8.1.0), these variables will still not be there.  
So I don't know if it's worth waiting for it at all, or using what I suggested, or improving it somehow.

bazel/emscripten_toolchain/wasm_cc_binary.bzl
sbc100: No worries, its all internal anyway, so keeping the local convention seems fine.

bazel/emscripten_toolchain/BUILD.bazel
sbc100: It we do end up landed this alternative method of have 5 different scripts, I think we should at least document why its needed, and perhaps link to a bazel bug, so we can potentially revert to the more common pattern in the future.

trzeciak: Okay, I can report this to bazel and add a link to the report here.  
The question is, where should I add a comment/TODO with a link in the sourcecode to make it clear enough?  

In a comment in each of the five scripts? Or somewhere in a collective place? Maybe link the comment to this PR there?

sbc100: Yes, a comment in each of the 5 scripts sounds good.

Something like "This scripts is needed because we cannot use the emscripten emdwp.py entry point because of a limiation/bug with bazel: <url>"

trzeciak: Okay, I'll prepare something for tomorrow.

trzeciak: If you think about it, it's probably possible to do it like this:
* add an `emdwp-template.py` file with the following content:
```py
#!/bin/bash
#
#  This script…
#

export EMCC_WASM_BACKEND=__EMCC_WASM_BACKEND__
export EM_BIN_PATH=__EM_BIN_PATH__
export EM_CONFIG_PATH=__EM_CONFIG_PATH__
export NODE_JS_PATH=__NODE_JS_PATH__

source $(dirname $0)/env.sh

exec python3 $EMSCRIPTEN/tools/emdwp.py "$@"
```
* add some bazel-stuff to extract those variable using `select` functionality to something like [TemplateVariableInfo](https://bazel.build/rules/lib/providers/TemplateVariableInfo
* add `genrule` to update `emdwp-template.py` based on `TemplateVariableInfo`
* and connect the output of this `genrule` as a input of `cc_toolchain.dwp_files`

I think this should work. Unfortunately it's a bit beyond the time I wanted to spend on it.  
On the other hand, I think it will be less readable than it is now _(although I may be biased/lazy)_.

trzeciak: I reported it to bazel: https://github.com/bazelbuild/bazel/issues/25336

I also updated all scripts to add this comment:

>     This script differs in form from emcc.{py,bat}/…, because bazel are limited/bugged in the way of executing dwp tool.
>     Bazel dwp action configuration does not pass environment variables, so we cannot use them in this script.
>     For more info, see PR discussion and bezel issue:
>     - https://github.com/emscripten-core/emsdk/pull/1531#discussion_r1962090650
>     - https://github.com/bazelbuild/bazel/issues/25336

phlax:
would be great for this to land it unblocks us (envoy)

sbc100:
> would be great for this to land it unblocks us (envoy)

@phlax do you know much about Bazel itself?  Would you be interested in sending them very small (maybe just one line) PR so that we don't need maintain all these extra wrapper scripts: https://github.com/bazelbuild/bazel/issues/25336

bazel/emscripten_toolchain/wasm_binary.py
```diff
@@ -26,11 +27,20 @@ def main():
...
   parser.add_argument('--outputs', help='Comma separated list of files that should be extracted from the archive. Only the extname has to match a file in the archive.')
+  parser.add_argument('--dwp_file', help='Optional dwp input file, generated when fission flags set.')
   parser.add_argument('--allow_empty_outputs', help='If an output listed in --outputs does not exist, create it anyways.', action='store_true')
   args = parser.parse_args()
 
   args.archive = os.path.normpath(args.archive)
   args.outputs = args.outputs.split(",")
+  args.dwp_file = os.path.normpath(args.dwp_file) if args.dwp_file else None
+
+  if args.dwp_file:
```
sbc100, line 38: Can you put `args.dwp_file = os.path.normpath(args.dwp_file)` here instead of line 36 (one less conditional).

bazel/emscripten_toolchain/wasm_binary.py
```diff
@@ -26,11 +27,20 @@ def main():
...
+    for idx, output in enumerate(args.outputs):
+      if output.endswith(".dwp"):  # also update extension 'binary.dwp' to 'binary.wasm.dwp'
```
sbc100, line 40: Use single quotes here

bazel/emscripten_toolchain/wasm_binary.py
```diff
@@ -26,11 +27,20 @@ def main():
   parser = argparse.ArgumentParser()
   parser.add_argument('--archive', help='The archive to extract from.')
   parser.add_argument('--outputs', help='Comma separated list of files that should be extracted from the archive. Only the extname has to match a file in the archive.')
+  parser.add_argument('--dwp_file', help='Optional dwp input file, generated when fission flags set.')
```
sbc100, line 30: Why doesn't the dwp file live in the archive like the rest on the files?

Is is because bazel actually supports the dwp file natively (unlike the other extra files we generate).   If there is support for dwp natively in bazel why do we need to involve the wasm_binary.py script at all here?  i.e. if its not part of the archive why we even involve the wasm-binary script at all?

phlax:
@sbc100 i can take a look if i get some spare cycles - but can i suggest we land this with the multiple scripts ... iiuc the fix would depend on the end users bazel version - and i know at least for us if it lands on bazel main now it may be a year+ before we are going to be up to that version

bazel/emscripten_toolchain/wasm_binary.py
trzeciak: Hm.. using `dwp` files is optional, won't that be a problem for the `normpath` method?

bazel/emscripten_toolchain/wasm_binary.py
trzeciak: This issue is more complex, although I am no expert, I will try to shed more light on it:
* during compilation we can generate .o files containing debugging information, which can be extracted from the binary after the linking phase (but this is not the case),
* debug fission, is the extraction of part of the debug information during compilation, then during each compilation, two files are created: .o and .dwo,
* a .dwp file is a collection of .dwo files
* bazel aggregates .dwo files into (multiple, if necessary) .dwp files, and multiple .dwp files into a single .dwp file (~100-leaf tree)
* dwp file is created as a separate target (file) next to the output file from the cc_binary target (the first time it wasn't obvious to me that if I have target `cc_binary(name = "foo", …)`, then to get a .dwp file I have to add a requirement to the target (file): `foo.dwp`)
* considering this, if we want to generate both files at once, we can add a filegroup that will have a dependency on both files (`foo` and `foo.dwp`)
* forwarding dependencies to the dwp file via the wasm_binary.py file (~via wasm_cc_binary), simply makes the same configuration apply to both files,
doing it differently would involve duplicating the `wasm_cc_binary` method.

A bit of text, but I hope you know what I mean.

trzeciak:
I would also suggest not to wait for changes https://github.com/bazelbuild/bazel/issues/25336 as it will block merging this MR only with the latest Bazel version (which is not necessary now) and in worst case this version will be released in the next major version.  
At the same time, although I tried to describe exactly what is needed, it has not been taken any further.
The solution itself may be "one-line", but someone may expect tests (which I generally agree with), which may require more understanding / more time.
At the same time, adding the requirement to upgrade bazel to the latest version (even though I think it's good and progressive), is no longer attractive to me (it's about other dependencies).

If we agreed not to wait for those changes, just add TODO in the code, then I can try to come back to this MR, merge the current branch, apply the reported fixes, etc.

sbc100:
Are you still interested in landing this?    If so can you rebase?

PR #1539, opened 2025-03-17
correction of problem 402

Sofija2:
takes into account that the file may be temporarily blocked (for example, by antivirus or another process), and tries to repeat the operation several times with an increase in waiting time.


kripken:
Is it not better to require the user to run the command again? A fixed number (10 in this PR) of retries seems arbitrary and may not be enough.

PR #1541, opened 2025-03-25
Remove closure compile hack when running `npm ci`

sbc100:
See https://chromium.googlesource.com/emscripten-releases/+/2ee2f60605a7774b4e194c235e372c29973038ab

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1260,76 +1260,14 @@ def emscripten_npm_install(tool, directory):
   env = os.environ.copy()
   env["PATH"] = node_path + os.pathsep + env["PATH"]
   print('Running post-install step: npm ci ...')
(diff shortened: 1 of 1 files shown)


dschuff (approved):

juj:
I was verifying this change locally, since this PR results in Java closure compiler no longer being deleted, but it looks like there is an earlier regression with the addition of the `bootstrap.py` script in Emscripten that masked the size diff.

Before `bootstrap.py` script, the `node_modules\` subdirectory on Windows would read `81.0 MB (85,015,488 bytes)` in size.

After addition of `bootstrap.py`, the subdirectory reads `169 MB (177,525,552 bytes)`, a 2x increase.

I wrote more details at https://github.com/emscripten-core/emsdk/issues/1546.

PR #1551, opened 2025-04-15
Don't build binaryen tests

sbc100:
These seems to be some issue with building gtest on ubuntu/focal.  Looks like maybe the libstdc++ is too old for it.


sbc100:
Landing TBR to get the tree green

tlively:
lgtm

PR #1563, opened 2025-05-29
Update Node.js to LTS 22.16.0.

juj:
This matches the minimum OS requirements set of macOS 11.0 and Windows 10:

https://github.com/nodejs/node/blob/v22.x/BUILDING.md#platform-list

### DIFF
--- a/bazel/MODULE.bazel
+++ b/bazel/MODULE.bazel
@@ -16,7 +16,7 @@ python.toolchain(
 )
 
 node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")
-node.toolchain(node_version = "22.16.0")
+node.toolchain(node_version = "20.18.0")
 use_repo(node, "nodejs")
 
 emscripten_deps = use_extension(
--- a/bazel/MODULE.bazel.lock
+++ b/bazel/MODULE.bazel.lock
@@ -897,7 +897,7 @@
               "node_urls": [
                 "https://nodejs.org/dist/v{version}/{filename}"
               ],
-              "node_version": "22.16.0",
+              "node_version": "20.18.0",
               "include_headers": false,
               "platform": "linux_amd64"
             }
@@ -910,7 +910,7 @@
               "node_urls": [
                 "https://nodejs.org/dist/v{version}/{filename}"
               ],
-              "node_version": "22.16.0",
+              "node_version": "20.18.0",
               "include_headers": false,
               "platform": "linux_arm64"
             }
@@ -923,7 +923,7 @@
               "node_urls": [
                 "https://nodejs.org/dist/v{version}/{filename}"
               ],
-              "node_version": "22.16.0",
+              "node_version": "20.18.0",
               "include_headers": false,
               "platform": "linux_s390x"
             }
@@ -936,7 +936,7 @@
               "node_urls": [
                 "https://nodejs.org/dist/v{version}/{filename}"
               ],
-              "node_version": "22.16.0",
+              "node_version": "20.18.0",
               "include_headers": false,
               "platform": "linux_ppc64le"
             }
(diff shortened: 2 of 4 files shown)


juj:
> What is the motivation for this update?

The main driver for me is to get security people off my neck, who are running Snyk scanner on the node.js we are bundling, and flagging Emscripten as "unsafe" on the basis of "so many vulnerabilities": https://security.snyk.io/package/linux/rhel%3A8/nodejs%3A20%2Fnodejs

Node 22 currently has only low severity issues: https://security.snyk.io/package/linux/chainguard%3Alatest/nodejs-22 which those people fortunately don't care about.

sbc100:
Fair enough.  Sounds reasonable.  Kind of.   I meant do the also care about possible exploits in llvm which I'm sure has many many more.  They probably don't understand we only run on trusted inputs.

In any case lgtm rather than trying to fight that battle I guess.

juj:
Hmm, `build-docker-image` fails with the following:

```
0.062 ## Update and install packages
31.62 E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/f/fontconfig/fontconfig-config_2.13.1-4.2ubuntu5_all.deb  403  Forbidden [IP: 185.125.190.82 80]
31.62 E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```
I don't know if that could indirectly be caused by this PR, or if that is something that is currently broken by some unrelated cause? The previous python PR landed without a 403 Forbidden at least.

juj:
There is a later error also

```
ERROR: failed to solve: process "/bin/sh -c echo \"## Update and install packages\"     && apt-get -qq -y update     && DEBIAN_FRONTEND=\"noninteractive\" TZ=\"America/San_Francisco\" apt-get -qq install -y --no-install-recommends         sudo         libxml2         ca-certificates         python3         python3-pip         wget         curl         zip         unzip         git         git-lfs         ssh-client         build-essential         make         ant         libidn12         cmake         openjdk-11-jre-headless     && apt-get -y clean     && apt-get -y autoclean     && apt-get -y autoremove     && rm -rf /var/lib/apt/lists/*     && rm -rf /var/cache/debconf/*-old     && rm -rf /usr/share/doc/*     && rm -rf /usr/share/man/??     && rm -rf /usr/share/man/??_*     && echo \"## Done\"" did not complete successfully: exit code: 100
```
that I don't know if that's a follow-up. Do you know of a way to debug these on the CI locally?

sbc100:
That looks like an unrelated network issue with the ubuntu package server.   Will restart..

sbc100:
The new error `test_dockerimage.sh: line 9: node: command not found` does look related.. although I'm not sure how exactly.

juj:
> The new error `test_dockerimage.sh: line 9: node: command not found` does look related.. although I'm not sure how exactly.

Ah thanks, found the error. There was a missing coded path to node version in `docker/Dockerfile`.

juj:
Search-replaced other references to 20.18.0 to 22.16.0 as well.

juj:
Ok, now all tests pass.

emsdk_manifest.json
```diff
@@ -156,6 +156,56 @@
...
+  {
+    "id": "node",
+    "version": "22.16.0",
+    "bitness": 32,
+    "arch": "x86",
+    "windows_url": "node-v22.16.0-win-x86.zip",
+    "activated_path": "%installation_dir%/bin",
+    "activated_path_skip": "node",
+    "activated_cfg": "NODE_JS='%installation_dir%/bin/node%.exe%'",
+    "activated_env": "EMSDK_NODE=%installation_dir%/bin/node%.exe%"
```
sbc100, line 168: At some point (just like with python) I do think we should remove the older version of these tools from this json file.  I don't see any point is keeping stuff we don't use around.   Maybe after 6 months or so?

PR #1567, opened 2025-06-13
Build and push multi-platform docker image

radiantly:
This PR adds the `publish-docker-image-multiplatform` CI job step that pushes a multi-platform docker image when a new tag is created. #1211 

Unfortunately, this is not as simple as pushing two images targeting different platforms to the same tag. There are a couple of ways to create a multi-platform image:
 - Using `docker buildx` to build images for all platforms. To do this images targeting non-native platforms would need to be built using QEMU (slower), a remote machine or cross-compilation.
 - Building and publishing platform-specific images `emscripten/emsdk-x64`, `emscripten/emsdk-arm64` and then creating a manifest that links to these images. This is the simpler solution to implement, and is what this PR does.

Tested with CircleCI on my fork. See the pushed images at [radiantly/emsdk · DockerHub](https://hub.docker.com/r/radiantly/emsdk/tags)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,7 +96,6 @@ jobs:
             python3 -m pip install flake8==7.1.1
             python3 -m flake8 --show-source --statistics --extend-exclude=./scripts
 
-
   test-linux:
     executor: ubuntu
     environment:
@@ -249,7 +248,7 @@ jobs:
           name: push image
           command: |
             docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
-            make -C ./docker version=${CIRCLE_TAG} alias=latest push
+            make -C ./docker version=${CIRCLE_TAG} alias=${CIRCLE_TAG}-x64 only_alias=true push
 
   publish-docker-image-arm64:
     executor: linux_arm64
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -238,7 +238,8 @@ jobs:
           name: push image
           command: |
             docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
-            make -C ./docker version=${CIRCLE_TAG} alias=latest push
+            make -C ./docker version=${CIRCLE_TAG} alias=${CIRCLE_TAG}-x64 only_alias=true push
+            make -C ./docker version=${CIRCLE_TAG} alias=latest only_alias=true push
```
sbc100: What does `only_alias` do here?  is it needed on both lines?

.circleci/config.yml
```diff
@@ -256,6 +257,16 @@ jobs:
...
 
+  publish-docker-image-multiplatform:
+    executor: linux_arm64
+    steps:
+      - checkout
+      - run:
+          name: push image
+          command: |
+            docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
+            make -C ./docker version=${CIRCLE_TAG} platforms="x64 arm64" push-multiplatform
```
sbc100: Should `latest` be an alias for the multi-platform rather than the x86 one?

docker/Makefile
```diff
@@ -5,6 +5,7 @@
 version =
 alias =
 only_alias =
+platforms =
```
sbc100: Maybe just hardcode `x64 arm64` here?    Seems simpler than threading an argument through and have extra error case below? 

Actually just inline inlining it directly into `push-multiplatform: ` keeps it even simpler for now.  

We can always add more complexity / flexibility if the architecture list gets long.

tdrz:
Any chance this will be merged?

.circleci/config.yml
radiantly: Yep, I've fixed this! (and rebased this branch)

docker/Makefile
radiantly: You're right. I've removed the argument.

Issue #1584, opened 2025-08-19
bazel wasm_cc_binary mul threads


yangjiwei-source:
wasm_cc_binary 
linkopts = [
        "-lembind",
        "-sMODULARIZE=1",
        "-sEXPORT_ES6=1",
        "-sSINGLE_FILE=1",
        "-Wpthreads-mem-growth",
        "-sERROR_ON_UNDEFINED_SYMBOLS=1",
        "-sDISABLE_EXCEPTION_CATCHING=0",
        "-frtti",
        "--bind",
        "-msimd128",
        "-sINITIAL_MEMORY=33554432",
        "-sWASM_BIGINT=1",
        "-sALLOW_MEMORY_GROWTH",
        "-pthread",
        "-sUSE_PTHREADS=1", 
        "-sPTHREAD_POOL_SIZE=6",
    ],
when add "-sUSE_PTHREADS=1", 
        "-sPTHREAD_POOL_SIZE=6", brower run，need *.worker.js, can't run
can't gen .worker.js, size is 0

PR #1585, opened 2025-08-19
[emsdk] Remove references to old output files such as .worker.js. NFC

sbc100:
Fixes: #1584


bazel/emscripten_toolchain/link_wrapper.py
walkingeyerobot, line 1: need to delete the references in `_wasm_cc_binary_legacy_impl` as well

PR #1588, opened 2025-08-23
Allow switching between overrides

juj:
Previously, if one used e.g. command

```
emsdk install --override-repository emscripten-main-64bit@https://github.com/juj/emscripten/tree/my_own_emscripten_branch emscripten-main-64bit
```

then the overridden installation would create a git clone from `juj/emscripten`, and that clone would be located under the default remote name `origin`.

This would prevent being able to switch between overrides on subsequent `emsdk install` commands. I.e. it would then not be possible to later do a default

```
emsdk install emscripten-main-64bit
```

command without an override, since the remote name `origin` would point to `juj/emscripten` instead of `emscripten-core/emscripten`.

This PR changes the naming scheme of `git clone`s when `--override-repository` is used, to name the clones with remote names from the override, so in the above case, the clone would appear under remote name `juj` and not `origin`.

This enables flipping between multiple `emsdk install` commands in the same checkout, without needing to nuke the installed directory in between.

This enables me to run experiments like github.com/emscripten-core/emscripten/pull/25025 locally on my CI, without needing to merge them upstream to become available.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -862,8 +862,8 @@ def git_pull(repo_path, branch_or_tag, remote_name='origin'):
     if target_is_tag:
       ret = run([GIT(), 'checkout', '--recurse-submodules', '--quiet', branch_or_tag], repo_path)
     else:
-      local_branch_prefix = ('remote_name' + '_') if remote_name != 'origin' else ''
-      ret = run([GIT(), 'checkout', '--recurse-submodules', '--quiet', '-B', 'local_branch_prefix' + 'branch_or_tag',
+      local_branch_prefix = (remote_name + '_') if remote_name != 'origin' else ''
+      ret = run([GIT(), 'checkout', '--recurse-submodules', '--quiet', '-B', local_branch_prefix + branch_or_tag,
                  '--track', remote_name + '/' + branch_or_tag], repo_path)
     if ret != 0:
       return False


emsdk.py
```diff
@@ -812,40 +819,58 @@ def git_recent_commits(repo_path, n=20):
...
+
+
+def git_clone(url, dstpath, branch, remote_name='origin'):
   debug_print('git_clone(url=' + url + ', dstpath=' + dstpath + ')')
   if os.path.isdir(os.path.join(dstpath, '.git')):
-    debug_print("Repository '" + url + "' already cloned to directory '" + dstpath + "', skipping.")
-    return True
+    remotes = get_git_remotes(dstpath)
+    if remote_name in remotes:
+      debug_print(f"Repository {url} with remote '{remote_name}' already cloned to directory '{dstpath}', skipping.")
```
sbc100: I believe this script is still supposed to be python2-runnable, so you can't use f-strings.

Worrying that CI didn't pick that up.

sbc100 (commented):
lgtm % comment

Indeed I think you are likely the only user of this code.

emsdk.py
juj: I don't think Python 2 is a use case anymore. Nothing works in Python 2 mode in Emscripten anymore, for a long time now.

sbc100: No, not emscripten, only emsdk.py.   This is because emsdk.py is a script that actually downloads and installs the version of python we will eventually use.   So, unlike emscripten, it needs to run on the OS / system version of python.  For example, python2 used to ship installed on macOS.

See the top of emsdk.py for example where we have an explicit check for `if sys.version_info >= (3,):`

If we want to change this policy it should be separate / larger change.

sbc100: Looks like we lost CI support for checking python2 support in #1183..

juj: > For example, python2 used to ship installed on macOS.

I see.

It looks like Python 2.x was removed on Apple macOS 12.3: https://developer.apple.com/documentation/macos-release-notes/macos-12_3-release-notes (March 14, 2022).

If macOS users install Xcode with the Xcode Command Line Tools, then they will get Python 3, even on the oldest macOS 11 (that was the first version to support Apple Silicon). I'd be surprised if there exists macOS users who did Emscripten development, but didn't have Xcode installed.

Though I'll revert back to Python 2.

sbc100: https://github.com/emscripten-core/emsdk/issues/1592

PR #1590, opened 2025-08-26
Nightly Node.js and Big Endian cross compilation Node.js

juj:
This PR adds support for

`./emsdk install node-nightly-64bit`

which finds the latest nightly node.js version and installs that. It is a moving target, so the next day when a new nightly is produced, reissuing `./emsdk install node-nightly-64bit` will install the newer published version.

Also, this PR adds a fixed 22.18.0 version for the cross compilation s390x Node.js target as well. On an x64 Linux system, this allows running

./emsdk install node-big-endian-crosscompile-22.16.0-64bit`

to install the big endian Node.js. This greatly simplifies the steps at https://github.com/emscripten-core/emscripten/compare/main...juj:emscripten:bigendian_test_suite#diff-c36b90121be240017fa490a1c00e63e47fa3235f5c1be0593e2b7502d017c778R9985-R10000 and enables a trivial way to switch between LE and BE Node.js versions for testing.

CC @slavek-kucera

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1477,7 +1477,10 @@ def extract_newest_node_nightly_version(versions):
     else:
       return []
 
-  return max(versions, key=lambda v: parse(v))
+  try:
+    return max(versions, key=lambda v: parse(v))
+  except:
+    return None
 
 
 def build_node_nightly(tool):
@@ -1485,7 +1488,12 @@ def build_node_nightly(tool):
   latest_nightly = extract_newest_node_nightly_version(nightly_versions)
   print(f'Latest Node.js Nightly download available is "{latest_nightly}"')
 
-  output_dir = os.path.abspath(f'node/nightly-{latest_nightly}/bin')
+  output_dir = os.path.abspath(f'node/nightly-{latest_nightly}')
+  # Node.js zip structure quirk: Linux and macOS archives have a /bin,
+  # Windows does not. Unify the file structures.
+  if WINDOWS:
+    output_dir += '/bin'
+
   if os.path.isdir(output_dir):
     return True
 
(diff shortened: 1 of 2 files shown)


emsdk.py
```diff
@@ -1450,6 +1450,78 @@ def find_emscripten_root(active_tools):
...
+    else:
+      return []
+
+  try:
+    return max(versions, key=lambda v: parse(v))
+  except Exception:
+    return None
+
+
+def build_node_nightly(tool):
```
sbc100: Is this actually building node from source?  It looks like its just download it, so maybe this should be called "download_...` to `fetch_...`?

emsdk_manifest.json
```diff
@@ -204,6 +204,30 @@
...
   },
+  {
+    "id": "node-big-endian-crosscompile",
+    "version": "22.16.0",
+    "arch": "x86_64",
+    "bitness": 64,
+    "linux_url": "https://nodejs.org/download/release/v22.16.0/node-v22.16.0-linux-s390x.tar.gz",
+    "activated_path": "%installation_dir%/bin",
+    "activated_path_skip": "node",
+    "activated_cfg": "NODE_JS=['qemu-s390x', '-L', '/usr/s390x-linux-gnu/', '%installation_dir%/bin/node']",
```
sbc100: `NODE_JS_TEST` can be used here if you want to only use it for running tests.

emsdk.py
juj: check

emsdk_manifest.json
juj: Oh that's fantastic.

kripken:
Thanks!

PR #1593, opened 2025-08-27
Add support for a new command 'emsdk deactivate'

juj:
Add support for a new command 'emsdk deactivate' that can be used to remove a tool from the active list.

Since the big endian cross compile Node.js is added to `NODE_JS_TEST` instead of `NODE_JS`, there needs to be a way to deactivate it for regular runs.

This PR adds a `emsdk deactivate` command to enable doing that.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2835,6 +2835,9 @@ def main(args):
                                   --override-repository emscripten-main@https://github.com/<fork>/emscripten/tree/<refspec>
 
 
+   emsdk deactivate tool/sdk    - Removes the given tool or SDK from the current set of activated tools.
+
+
    emsdk uninstall <tool/sdk>   - Removes the given tool or SDK from disk.''')
 
     if WINDOWS:
@@ -3104,7 +3107,7 @@ def main(args):
   elif cmd == 'update-tags':
     errlog('`update-tags` is not longer needed.  To install the latest tot release just run `install tot`')
     return 0
-  elif cmd == 'activate':
+  elif cmd == 'activate' or cmd =='deactivate':
     if arg_permanent:
       print('Registering active Emscripten environment permanently')
       print('')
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -3116,7 +3119,14 @@ def print_tools(t):
         tool = find_sdk(arg)
         if tool is None:
           error_on_missing_tool(arg)
-      tools_to_activate += [tool]
+
+      if cmd == 'activate':
+        tools_to_activate += [tool]
+      elif tool in tools_to_activate:
+        print('Deactivating tool ' + str(tool) + '.')
+        tools_to_activate.remove(tool)
```
sbc100, line 3127: So "remove" is something that already exists?   Is it analogous to "deactivate" ?  What is it used for today?

sbc100 (commented):
lgtm % comment

emsdk.py
juj: `tools_to_activate` is a Python list here, so this is removing the found tool from the Python list by value.

sbc100: Sorry, I guess my question is if tools already have `.remove()` method, is there some existing want to do that?   i.e. is there an `emsdk remove` or `emsdk uninstall` already?  If so why do we need do `deactivate` synonym.   If not, what when is the `.remove()` method used?

juj: Err, still quite didn't parse that..

> if tools already have .remove() method

Looking at `class Tool`, I don't see it to have a `.remove()` method. 

>  i.e. is there an emsdk remove or emsdk uninstall already?

There is no `emsdk remove`, but there is `emsdk uninstall` that is symmetric to `emsdk install`.

The purpose for `emsdk deactivate` is to be symmetric to `emsdk activate`.

sbc100: Oh sorry.. I thought this was calling a method called `remove()` .. my bad

PR #1611, opened 2025-10-06
[bazel] Replace deprecated '-g4' flag with '-g3'

DoDoENT:
- this avoids deprecated flag warnings from emscripten when building with full_debug_info or wasm_ubsan features enabled

### DIFF
--- a/bazel/emscripten_toolchain/toolchain.bzl
+++ b/bazel/emscripten_toolchain/toolchain.bzl
@@ -408,7 +408,7 @@ def _impl(ctx):
             implies = ["profiling"],
         ),
 
-        # Turns on full debug info (-gsource-map).
+        # Turns on full debug info (-g3).
         feature(name = "full_debug_info"),
 
         # Enables the use of "Emscripten" Pthread implementation.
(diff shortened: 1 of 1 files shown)


bazel/emscripten_toolchain/toolchain.bzl
```diff
@@ -408,7 +408,7 @@ def _impl(ctx):
             implies = ["profiling"],
         ),
 
-        # Turns on full debug info (-g4).
+        # Turns on full debug info (-gsource-map).
```
sbc100: My `-gsource-map` is not full debug info, its just a source map.  Full debug info would be dwarf. 

How about just just `-g`?  (Or `-g3`?)

DoDoENT: This was literally the compile warning (which my build treats as errors): it said that I should use `-gsource-map` instead of the `-g4`, which is now deprecated.

DoDoENT: From the [changelog](https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#2017-04102021):

> Deprecate `-g4`, which is a little confusing as it does not do more than `-g3` but instead emits source maps instead of DWARF. `-g4` will now warn. A new flag `-gsource-map` enables source maps without warning.

sbc100: It seems like the author of toolchain.bzl was confused by `-g4` too and assumed that it enabled "full debug info".

Seems like the right thing to use here instead would be `-g` or `-g3`?

DoDoENT: I'll put `-g3`

walkingeyerobot (approved):

PR #1614, opened 2025-10-07
Fix macOS Firefox test run hangs

juj:
Fix macOS Firefox test run hangs by disabling macOS from prompting to restore windows after an unclean browser termination.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1400,6 +1400,16 @@ def download_firefox(tool):
   }
 }''')
 
+  if MACOS:
+    # Disable a macOS feature where if the browser is terminated mid-execution, (e.g. by
+    # CI aborting), then the next time the browser is launched, macOS might bring up a dialog
+    # "The last time you opened Firefox, it unexpectedly quit while reopening windows.
+    #  Do you want to try to reopen its windows again?"
+    # that will block automated CI runs.
+    # Disable this feature by changing the behavior of the program with macOS 'defaults'.
+    run(['defaults', 'write', '-app', root, 'ApplePersistenceIgnoreState', 'YES'])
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -1400,6 +1400,16 @@ def save_actual_version():
...
 
+  if MACOS:
+    # Disable a macOS feature where if the browser is terminated mid-execution, (e.g. by
+    # CI aborting), then the next time the browser is launched, macOS might bring up a dialog
+    # "The last time you opened Firefox, it unexpectedly quit while reopening windows.
+    #  Do you want to try to reopen its windows again?"
+    # that will block automated CI runs.
+    # Disable this feature by changing the behavior of the program with macOS 'defaults'.
+    run(['defaults', 'write', '-app', root, 'ApplePersistenceIgnoreState', 'YES'])
+    run(['defaults', 'write', '-app', root, 'NSQuitAlwaysKeepsWindows', '-bool', 'false'])
```
sbc100, line 1411: I'm not sure we want installing firefox using emsdk to have these kinds of global side effect.

This seems like something more related to CI/testing specifically not emsdk generally?

sbc100: Oh I see its only effecting the firefox binary... I guess thats not as bad, but still rather strange.

sbc100: I don't suppose there is a firefox specific user setting for this is here?  It has to be an OS setting?    I could imagine the firefox is calling some kind of `macos_get_persistence_state()` which might be controllable from the firefox side?

juj: No, this is just perfect, because we can package these types of customizations into just the Firefox test binary.. and it won't for example affect the Firefox browser that user might have installed globally on their system.

juj: This applies system wide as far as I know, i.e. all apps have a setting like this.

Issue #1581, opened 2025-08-04
Bazel support for multiple emscripten caches


DoDoENT:
I'm experimenting with Bazel with the goal of replacing my CMake+Conan setup with Bazel. I'm using LTO in release builds and in debug and fastbuild I'm not using LTO. However, when using Bazel, this doesn't work.

Namely, the default configuration does not ship with LTO-enabled cache, so when I enable LTO, I get the error `Exception: FROZEN_CACHE is set, but cache file is missing: "sysroot/lib/wasm32-emscripten/thinlto/crtbegin.o"`.

Then, after reading [the instructions for enabling LTO](https://github.com/emscripten-core/emsdk/blob/main/bazel/README.md), I added the declaration for cache as given in the instructions and my LTO build still didn't work.

After a couple of hours of trial and error, I discovered that the documentation does not list all minimum targets that need to be built in cache - for [my project](https://github.com/dodoent/bazel-playground) to work, I had to add at least these targets:

```
emscripten_cache.configuration(flags = ["--lto=thin"])
emscripten_cache.targets(targets = [
    "crtbegin",
    "libal",
    "libc-mt",
    "libc-mt-debug",
    "libc",
    "libc++-mt-noexcept",
    "libc++-noexcept",
    "libc++-debug-mt-noexcept",
    "libc++abi-mt-noexcept",
    "libc++abi-debug-mt-noexcept",
    "libc++abi-debug-noexcept",
    "libcompiler_rt-mt",
    "libcompiler_rt",
    "libembind-rtti",
    "libemmalloc-mt",
    "libemmalloc-mt-debug",
    "libemmalloc",
    "libfetch-mt",
    "libfetch",
    "libGL-getprocaddr",
    "libGL-mt-getprocaddr",
    "libhtml5",
    "libnoexit",
    "libsockets-mt",
    "libsockets",
    "libprintf_long_double-mt",
    "libprintf_long_double-mt-debug",
    "libprintf_long_double-debug",
    "libstubs",
    "libstubs-debug",
    "libc-debug",
    "libdlmalloc",
    "libc++abi-debug-noexcept",
])
```

Now the LTO build started working, but `dbg` and `fastbuild` builds stopped working, throwing the very same exception:

`Exception: FROZEN_CACHE is set, but cache file is missing: "sysroot/lib/wasm32-emscripten/crtbegin.o"` because now the LTO-enabled cache is used, but `dbg` and `fastbuild` don't have LTO enabled.

[...]

PR #1620, opened 2025-10-18
[Bazel] Support for a custom prebuilt emscripten cache

DoDoENT:
This adds support for using a prebuilt cache as an archive instead of building it from scratch every time.

This solves multiple problems:

- build speed on CI machines that have cold cache
    - instead of building the secondary cache every time, it can simply download it
- effective support for multiple caches (#1581)
    - emscripten supports having multiple caches on its own, but it builds `thinlto`, `lto`, and other cache combinations on demand. This doesn't work in the Bazel world, where the cache needs to be frozen. 
    - the current solution allowed building a secondary cache, but only with a single configuration
    - this allows the use of any archive that contain any emscripten cache
    - this PR does not handle how that cache is built - you can build it with embuilder, zip it and serve it from your server or whatever you want
- secondary cache hermeticity problem
    - the current solution for secondary cache generated `emscripten_config` file that contained full path to the cache. 
    - this was not hermetic as this path could be different on different machines
    - this PR calculates the cache path by using the same trick with environment variables, ensuring that `emscripten_config` file is always same, regardless of the machine
    - additionally, it ensures that the cache is provided as an input to the toolchain so that Bazel can correctly add it to the sandbox environment

After applying this patch to our internal codebase, the regular builds of WASM code went down from 12-15 minutes to cca 30 seconds, regardless on which machine it runs, as long as remote cache is warm.

### DIFF
--- a/bazel/emscripten_cache.bzl
+++ b/bazel/emscripten_cache.bzl
@@ -3,6 +3,21 @@ package(default_visibility = ['//visibility:public'])
 exports_files(['emscripten_config'])
 """
 
+BUILD_FILE_USE_BUILTIN_CACHE = """
+alias(
+    name = "emscripten_cache",
+    actual = "{}//:builtin_cache",
+)
+"""
+
+BUILD_FILE_USE_SECONDARY_CACHE = """
+filegroup(
+    name = "emscripten_cache",
+    srcs = glob(["cache/**"]),
+    visibility = ["//visibility:public"],
+)
+"""
+
 EMBUILDER_CONFIG_TEMPLATE = """
 CACHE = '{cache}'
 BINARYEN_ROOT = '{binaryen_root}'
@@ -29,6 +44,26 @@ def get_root_and_script_ext(repository_ctx):
     else:
         fail("Unsupported operating system")
 
+def get_bin_deps_repo_name(repository_ctx):
+    if repository_ctx.os.name.startswith("linux"):
+        if "amd64" in repository_ctx.os.arch or "x86_64" in repository_ctx.os.arch:
+            return "@emscripten_bin_linux"
+        elif "aarch64" in repository_ctx.os.arch:
+            return "@emscripten_bin_linux_arm64"
+        else:
+            fail("Unsupported architecture for Linux")
+    elif repository_ctx.os.name.startswith("mac"):
+        if "amd64" in repository_ctx.os.arch or "x86_64" in repository_ctx.os.arch:
+            return "@emscripten_bin_mac"
+        elif "aarch64" in repository_ctx.os.arch:
+            return "@emscripten_bin_mac_arm64"
+        else:
+            fail("Unsupported architecture for MacOS")
+    elif repository_ctx.os.name.startswith("windows"):
+        return "@emscripten_bin_win"
+    else:
+        fail("Unsupported operating system")
+
 def _emscripten_cache_repository_impl(repository_ctx):
     # Read the default emscripten configuration file
     default_config = repository_ctx.read(
@@ -38,6 +73,8 @@ def _emscripten_cache_repository_impl(repository_ctx):
     )
 
     repo_metadata = None
+    build_file_content = BUILD_FILE_CONTENT_TEMPLATE
+    use_builtin_cache = True
 
     if repository_ctx.attr.prebuilt_cache_url:
         repository_ctx.download_and_extract(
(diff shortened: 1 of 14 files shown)


sbc100:
@walkingeyerobot can you recommend anyone else who might be good to review this?

nickschaap:
This is incredibly useful. We added this patch to our usage as well.

sbc100:
I don't really feel qualified to review this.  Maybe @mmorel-35 can chime in?

bazel/emscripten_cache.bzl
```diff
@@ -37,7 +72,24 @@ def _emscripten_cache_repository_impl(repository_ctx):
...
+            url = repository_ctx.attr.prebuilt_cache_url,
+            output = "cache",
+            sha256 = repository_ctx.attr.prebuilt_cache_sha256,
+            strip_prefix = repository_ctx.attr.prebuilt_cache_strip_prefix,
+        )
+
+        # Use the prebuilt cache
+        use_builtin_cache = False
+
+        repo_metadata = repository_ctx.repo_metadata(reproducible = True)
```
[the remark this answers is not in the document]
DoDoENT: @copilot, instead of making the repository reproducible depending on whether sha256 was provided, please rather make sha256 mandatory, as you suggested.

bazel/emscripten_cache.bzl
```diff
@@ -74,40 +126,68 @@ def _emscripten_cache_repository_impl(repository_ctx):
...
+            prebuilt_cache_sha256 = prebuilt_cache.sha256
+            prebuilt_cache_strip_prefix = prebuilt_cache.strip_prefix
 
     _emscripten_cache_repository(
         name = "emscripten_cache",
         configuration = all_configuration,
         targets = all_targets,
+        prebuilt_cache_url = prebuilt_cache_url,
+        prebuilt_cache_sha256 = prebuilt_cache_sha256,
+        prebuilt_cache_strip_prefix = prebuilt_cache_strip_prefix,
```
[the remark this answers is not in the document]
DoDoENT: @copilot apply changes based on [this feedback](https://github.com/emscripten-core/emsdk/pull/1620#discussion_r3166013462)

DoDoENT:
I don't know why this `test-mac-arm64` build is failing - it has nothing to do with Bazel and I saw it also failing for other PRs.

@sbc100, do you know more details about this?

EDIT: I see the answer [here](https://github.com/emscripten-core/emsdk/pull/1704).

bazel/emscripten_cache.bzl
```diff
@@ -37,7 +72,24 @@ def _emscripten_cache_repository_impl(repository_ctx):
...
+    build_file_content = BUILD_FILE_CONTENT_TEMPLATE
+    use_builtin_cache = True
+
+    if repository_ctx.attr.prebuilt_cache_url:
+        repository_ctx.download_and_extract(
+            url = repository_ctx.attr.prebuilt_cache_url,
+            output = "cache",
+            sha256 = repository_ctx.attr.prebuilt_cache_sha256,
+            strip_prefix = repository_ctx.attr.prebuilt_cache_strip_prefix,
+        )
```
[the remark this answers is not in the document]
DoDoENT: Excellent suggestion! Thank you. I added tests that uncovered incompatibilities with Bazel 7, which I now fixed.

mmorel-35 (approved):

bazel/test_prebuilt_cache/.bazelrc
```diff
@@ -0,0 +1 @@
+build --incompatible_enable_cc_toolchain_resolution
```
sbc100: Do you know if we still need this `incompatible_enable_cc_toolchain_resolution` setting?  Is there something we can do to remove the need for it?

bazel/emscripten_cache.bzl
```diff
@@ -72,42 +126,72 @@ def _emscripten_cache_repository_impl(repository_ctx):
             },
         )
         if result.return_code != 0:
-            fail("Embuilder exited with a non-zero return code")
+            fail("Embuilder exited with a non-zero return code\nstdout: {}\nstderr: {}".format(result.stdout, result.stderr))
```
sbc100: I wonder if we should avoid capturing the stdout/stderr of embuilder?  Is it problem to simpy pass it through.  By default its not very verbose.

sbc100 (approved):

sbc100:
The macos build failures is known issue.  See https://github.com/WebAssembly/binaryen/pull/8609 and https://github.com/emscripten-core/emsdk/pull/1704

bazel/test_prebuilt_cache/.bazelrc
DoDoENT: I thought this was needed for Bazel 7, but I see now that in Bazel 7 this got enabled by default. Since we don't support older bazel versions, it's safe to remove it.

bazel/emscripten_cache.bzl
DoDoENT: I added it to debug why embuilder was failing when I was creating prebuilt cache for the test. For some reason I thought that the default for `quiet` is `False`, so I never tried explicitely set it to `False` 😅 .

I agree it's better to be passed through.

PR #1623, opened 2025-10-24
Fix windows emsdk returncode

juj:
When `emsdk install foo` would fail on Windows, the return code would get ignored since there were later commands in the script file.

To fix that, capture the error code from the emsdk call, and output that as the final return code.

Also a drive-by Python path fix, which seemed to be out of date in the .ps1 file.

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -220,6 +220,8 @@ int main() {
   def test_binaryen_from_source(self):
     if MACOS:
       self.skipTest("https://github.com/WebAssembly/binaryen/issues/4299")
+    if WINDOWS:
+      self.skipTest("https://github.com/emscripten-core/emsdk/issues/1624")
     print('test binaryen source build')
     run_emsdk(['install', '--build=Release', '--generator=Unix Makefiles', 'binaryen-main-64bit'])
 


emsdk.bat
```diff
@@ -34,8 +34,9 @@ set EMSDK_PY=python
 
 :end
 call %EMSDK_PY% "%~dp0\emsdk.py" %*
+set _EMSDK_ERRORLEVEL=%ERRORLEVEL%
 
-endlocal
+endlocal & set "_EMSDK_ERRORLEVEL=%_EMSDK_ERRORLEVEL%"
```
sbc100: Shouldn't we instead early exit if this command fails?

We don't want to go on run the other stuff here if the emsdk command fail.

juj: Good point. That is logically better, though doesn't really matter in practice either way, unlikely we'll generate the emsdk_set_env.bat file any way.

But it does give simpler code since it allows avoiding leaking that _EMSDK_ERRORLEVEL env. var. outside, so migrated to that code.

emsdk.bat
```diff
@@ -35,6 +35,13 @@ set EMSDK_PY=python
 :end
 call %EMSDK_PY% "%~dp0\emsdk.py" %*
 
+:: If emsdk returned with error code >= 1, then skip executing emsdk_set_env
+:: below.
+if ERRORLEVEL 1 (
```
sbc100, line 40: Is there some way to have this happen automatically with the `call` above?  Something like `set -o errexit` in bash?

juj: Not that I know of.

juj:
I'll skip that for now. https://github.com/emscripten-core/emsdk/issues/1624

sbc100 (approved):

PR #1628, opened 2025-11-16
CMake 4.2.0-rc3

juj:
Add links to CMake 4.2.0-rc3 into the manifest, and a mechanism inside emsdk.py to look up an installed/activated CMake tool.

This enables a simple way to run `emsdk install cmake-4.2.0-rc3-64bit` to acquire the CMake dependency needed for e.g. compiling LLVM and Binaryen. This removes one manual setup item for setting up Emscripten from source.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -997,8 +997,7 @@ def find_cmake():
         info('Found installed CMake tool at "' + cmake_exe + '"')
         return cmake_exe
 
-  errlog('Unable to find "cmake" in PATH, or as installed/activated tool! Please install CMake first')
-  sys.exit(1)
+  exit_with_error('Unable to find "cmake" in PATH, or as installed/activated tool! Please install CMake first')
 
 
 def make_build(build_root, build_type):


emsdk.py
```diff
@@ -962,14 +962,53 @@ def build_env(generator):
...
+  # then use it to configure to build LLVM.
+  for tool in reversed(tools):
+    if tool.id == 'cmake' and tool.is_installed():
+      cmake_exe = locate_cmake_from_tool(tool)
+      if cmake_exe:
+        info('Found installed CMake tool at "' + cmake_exe + '"')
+        return cmake_exe
+
+  errlog('Unable to find "cmake" in PATH, or as installed/activated tool! Please install CMake first')
+  sys.exit(1)
```
sbc100: exit_with_error

emsdk.py
```diff
@@ -962,14 +962,53 @@ def build_env(generator):
...
+    if os.path.isfile(cmake_exe):
+      return cmake_exe
+
+  # 1. If user has activated a specific cmake tool, then use that tool to configure the build.
+  for tool in reversed(tools):
+    if tool.id == 'cmake' and tool.is_active():
+      cmake_exe = locate_cmake_from_tool(tool)
+      if cmake_exe:
+        info('Found installed+activated CMake tool at "' + cmake_exe + '"')
+        return cmake_exe
```
sbc100, line 981: Its its active won't to already be in the PATH?

juj: No, because a tool being active does not mean that the current terminal session would have entered into the environment.

I.e. the `./emsdk activate` vs `source ./emsdk_env` distinction.

sbc100: I see.  As an alternative should we instead do one of these two things:

1. Somehow run emsdk *as if* `./emsdk activate` had been run (i.e. when running emsdk all active tools are always added to the path).
2. Create an alternative to the `which` helper that automatically includes active paths?

juj: Those are good ideas, though I think we probably only need to go there if we run into other dependencies like this - which I can't think of at the moment.

PR #1633, opened 2025-11-26
Update information about linux version used to build prebuilt binaries. NFC

sbc100:


README.md
```diff
@@ -68,10 +68,11 @@ https://emscripten.org/docs/building_from_source/toolchain_what_is_needed.html.
 - `python`: Version 3.8 or above.
 - `java`: For running closure compiler (optional)
 
-The emsdk pre-compiled binaries are built against Ubuntu/Focal 20.04 LTS and
-therefore depend on system libraries compatible with versions of `glibc` and
-`libstdc++` present in that release.  If your linux distribution is very old
-you may not be able to use the pre-compiled binaries packages.
+The emsdk pre-compiled binaries are built against debian/stretch (for x86_64)
+and debian/bullseye (for arm64) sysroots and therefore depend on system
+libraries compatible with versions of `glibc` and `libstdc++` present in those
```
dschuff: Our builds do not use the libstdc++ in that sysroot. We statically link our own version of libcxx (precisely because that libstdc++ is so old). So those binaries only depend on glibc from that platform.

sbc100: Updated to add that extra context.

README.md
```diff
@@ -68,10 +68,13 @@ https://emscripten.org/docs/building_from_source/toolchain_what_is_needed.html.
...
 - `java`: For running closure compiler (optional)
 
-The emsdk pre-compiled binaries are built against Ubuntu/Focal 20.04 LTS and
-therefore depend on system libraries compatible with versions of `glibc` and
-`libstdc++` present in that release.  If your linux distribution is very old
-you may not be able to use the pre-compiled binaries packages.
+The emsdk pre-compiled binaries are built against debian/stretch (for x86_64)
+and debian/bullseye (for arm64) sysroots and therefore depend on system
+libraries compatible with the version of `glibc` (and other libraries) present
+in those releases.  Notably they use they embed their own version of `libc++`
```
kripken:
```suggestion
in those releases.  Notably they embed their own (modern) version of `libc++`
```

README.md
```diff
@@ -68,10 +68,13 @@ https://emscripten.org/docs/building_from_source/toolchain_what_is_needed.html.
...
+and so avoid depending on the system `libstdc++` or `libc++`. If your linux
+distribution is very old you may not be able to use the pre-compiled binaries
+packages.
```
kripken: I would change this order though:

1. We use a sysroot ... using glibc etc. there
2. If your linux is very old, these precompiled binaries may not work
3. But, note that the libc++ build is a modern one (not from the sysroot), so that at least should not be a problem

In this order I think it is clearer that libc++ is not a problem.

kripken: (the change is to move mention of libc++ to the end, and clarify the context)

sbc100: Ok, done.

kripken (approved):

Issue #1631, opened 2025-11-25
"latest" tag on Docker Hub still points to version 4.0.15


chrisguttandin:
The "latest" tag of the emsdk image on Docker Hub still points to version 4.0.15. As of now version 4.0.20 seems to be the latest version.

emscripten/emsdk:latest
https://hub.docker.com/layers/emscripten/emsdk/latest/images/sha256-27bc6267cb285223b8aebb7627bfebae7cb3ad2aaa0d5923b8aa5321793033e8

emscripten/emsdk:4.0.15
https://hub.docker.com/layers/emscripten/emsdk/4.0.15/images/sha256-27bc6267cb285223b8aebb7627bfebae7cb3ad2aaa0d5923b8aa5321793033e8

sbc100:
I do see https://hub.docker.com/repository/docker/emscripten/emsdk/tags/4.0.20 has been published, but somehow that latest is not being updated maybe?

sbc100:
Seems like maybe `make -C ./docker version=${CIRCLE_TAG} alias="latest" push-multiplatform` is not actually pushing the alias maybe?

chrisguttandin:
Could it be related to [these changes](https://github.com/emscripten-core/emsdk/commit/44f11eac54ae57e3352ec15131e0b422ef67dd3e)?

Version 4.0.15 was the last release with an alias. It was [released on Sep 17th](https://github.com/emscripten-core/emsdk/releases/tag/4.0.15). The commit above is from Sep 25th.

radiantly:
Hi, it looks like I missed out on pushing the alias in the `push-multiplatform` target - will create a PR fixing this in a few hours

chrisguttandin:
Thanks a lot for the quick fix.

PR #1635, opened 2025-11-27
Correctly tag latest built docker image

radiantly:
Additionally removes the `alias` and `only_alias` arguments and adds a single `tag` argument - this makes it clear about the tag being pushed to.

Tested on my fork: https://hub.docker.com/r/radiantly/emsdk

Fixes: #1631


sbc100:
Is there a command I can run locally to update the alias right now to point to 4.0.20?

sbc100:
Done!

PR #1641, opened 2025-12-08
Python fixes from `ruff check`

sbc100:


emsdk.py
```diff
@@ -1152,7 +1154,7 @@ def build_llvm(tool):
 
   enable_assertions = ENABLE_LLVM_ASSERTIONS.lower() == 'on' or (ENABLE_LLVM_ASSERTIONS == 'auto' and build_type.lower() != 'release' and build_type.lower() != 'minsizerel')
 
-  if ARCH == 'x86' or ARCH == 'x86_64':
+  if ARCH in {'x86', 'x86_64'}:
```
juj: Curious why `ruff` insists that this requires a set, but e.g. the `.endswith()` below and `.startswith()` above must use a tuple... that seems arbitrary.

sbc100: I don't think ruff cares if you use a set or a tuple here.

PR #1646, opened 2025-12-18
Fix for window very-long-filename support. NFC

sbc100:
We have a function called `fix_potentially_long_windows_pathname` which is used when unzipping archives under windows.

Without this the unzip process will fail if the total filename length ever exceeds 256.   This change adds a test for installing deep in the filesystem (where pathnames exceed 256) and fixes a bug in this code which was causing the test to fail.

The fact that this code had this bug for god-known-how-long means I think nobody was depending on it, so we could probably just remove it, but maybe better to fix it and add this test?

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -32,6 +32,12 @@ def listify(x):
   return [x]
 
 
+def copy_emsdk_to(targetdir):
+  for filename in os.listdir('.'):
+    if not filename.startswith('.') and not os.path.isdir(filename):
+      shutil.copy2(filename, os.path.join(targetdir, filename))
+
+
 def check_call(cmd, **kwargs):
   if type(cmd) is not list:
     cmd = cmd.split()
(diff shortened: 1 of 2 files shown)


test/test.py
```diff
@@ -276,6 +280,28 @@ def test_keep_downloads(self):
...
+    os.makedirs(longpath, exist_ok=True)
+    print(longpath)
+    copy_emsdk_to(longpath)
+
+    olddir = os.getcwd()
+    try:
+      os.chdir(longpath)
+      run_emsdk('install latest')
+      run_emsdk('activate latest')
+      check_call([upstream_emcc, os.path.join(olddir, 'hello_world.c')])
```
dschuff: do we need to clean up longpath afterwards?

sbc100: The tests in this repo don't current clean up after themsleves (for example `hello_world.c` is create on startup and never deleted).

We should probably move all generated files into `out/test` or somesuch, but I'll leave that as a followup maybe?

dschuff (approved):

PR #1651, opened 2026-01-22
Fix EMSDK_OS override. NFC

sbc100:
It is sometimes useful when debugging to be able to things like install the windows binaries on linux using:

```
$ EMSDK_OS=windows ./emsdk install latest
```

Without these changes this zipfile extraction fails due to the use of `//?/` windows filename prefix.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -77,13 +77,13 @@ MSYS = False
 MACOS = False
 LINUX = False
 
-if 'EMSDK_OS' in os.environ:
-  EMSDK_OS = os.environ['EMSDK_OS']
-  if EMSDK_OS == 'windows':
+os_override = os.environ.get('EMSDK_OS')
+if os_override:
+  if os_override == 'windows':
     WINDOWS = True
-  elif EMSDK_OS == 'linux':
+  elif os_override == 'linux':
     LINUX = True
-  elif EMSDK_OS == 'macos':
+  elif os_override == 'macos':
     MACOS = True
   else:
     assert False, 'EMSDK_OS must be one of: windows, linux, macos'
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -524,7 +524,7 @@ def untargz(source_filename, dest_dir):
 # See https://msdn.microsoft.com/en-us/library/aa365247.aspx#maxpath and http://stackoverflow.com/questions/3555527/python-win32-filename-length-workaround
 # In that mode, forward slashes cannot be used as delimiters.
 def fix_potentially_long_windows_pathname(pathname):
-  if not WINDOWS or MSYS:
+  if (not WINDOWS or MSYS) or 'EMSDK_OS' in os.environ:
```
dschuff: EMSDK_OS is a global var above. It doesn't look like it's initialized though. Maybe it would be better to give it an initial value and then we don't have to check `os.environ` explicitly down here?

emsdk.py
```diff
@@ -524,7 +524,7 @@ def untargz(source_filename, dest_dir):
 # See https://msdn.microsoft.com/en-us/library/aa365247.aspx#maxpath and http://stackoverflow.com/questions/3555527/python-win32-filename-length-workaround
 # In that mode, forward slashes cannot be used as delimiters.
 def fix_potentially_long_windows_pathname(pathname):
-  if not WINDOWS or MSYS:
+  if (not WINDOWS or MSYS) or os_override:
```
dschuff, line 527: This would prevent a user on windows from overriding to unix. I guess that's probably ok?

sbc100: Yes, but only if they also happen have huge filename lengths that require this `fix_potentially_long_windows_pathname` hack.

I'm also working on trying to add some testing to `fix_potentially_long_windows_pathname` so that we don't actually use it unless we need to.   But its really really hard because this `//?/` pathnames don't work consistently in many APIs.  Its really bad.

sbc100: Returning early here just means you get have normal windows pathname limits

Issue #1642, opened 2025-12-12
emsdk can't be used with Bazel 9/10


theoparis:
```
ERROR: Traceback (most recent call last):
        File "/home/theo/.cache/bazel/_bazel_theo/a51fe4523e7bd3b5d868cc54ce49e796/external/emsdk+/emscripten_toolchain/toolchain.bzl", line 1148, column 43, in <toplevel>
                emscripten_cc_toolchain_config_rule = rule(
Error in rule: at index 0 of provides, got element of type NoneType, want Provider
ERROR: /home/theo/src/skia/src/gpu/BUILD.bazel:56:16: While resolving toolchains for target //src/gpu:gpu (1c5f67e): invalid registered toolchain '//emscripten_toolchain:cc-toolchain-wasm-emscripten_linux': error loading package '@@emsdk+//emscripten_toolchain': at /home/theo/.cache/bazel/_bazel_theo/a51fe4523e7bd3b5d868cc54ce49e796/external/emsdk+/remote_emscripten_repository.bzl:4:6: initialization of module 'emscripten_toolchain/toolchain.bzl' failed
ERROR: Analysis of target '//src/gpu:gpu' failed; build aborted
```

Bazel 9.0 is supposed to be released later this month. It seems that this file needs to pull in CcToolchainConfigInfo via `load` now. This causes my WIP Bazel 9 skia PR to fail to run tests.  I could make a PR if that would be helpful.

Vertexwahn:
Some other fixes that might be needed to support Bazel 9.0.0: https://github.com/emscripten-core/emsdk/pull/1649

sbc100:
We certainly test with bazel 9 these days, so maybe this can be closed?   Not sure about bazel 10.  @DoDoENT would that be hard to add?

sbc100:
Oh, bazel 10 is not even released yet.   I think we can close this and open a specific bazel 10 issue once its available.

DoDoENT:
Personally, I would avoid testing with bazel 10 until it reaches at least first beta. Before that it's simply too unstable.

PR #1666, opened 2026-02-03
Add missing loads for bazel 9 compatibility 

mmorel-35:
Add missing loads for bazel 9 compatibility

Add Bazel 8 testing workflows 

Includes buildifier of modified files

See #1642

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -273,50 +273,40 @@ jobs:
             make -C ./docker version=${CIRCLE_TAG} tag=${CIRCLE_TAG} push-multiplatform
             make -C ./docker version=${CIRCLE_TAG} tag="latest" push-multiplatform
 
-  test-bazel7-linux:
+  test-bazel-linux:
+    parameters:
+      bazel_version:
+        type: string
+        default: "latest"
     executor: ubuntu
     environment:
-      USE_BAZEL_VERSION: "7.x"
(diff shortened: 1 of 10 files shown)


sbc100:
If you think this will make stuff work with latest bazel maybe you could un-comment the bazel-latest jobs from .circleci/config.yml too?

Or is the idea that this is just a small step in the right direction but on its own does not make the bazel 9 toolchain work?

walkingeyerobot (approved):
thanks very much for this!

.circleci/config.yml
```diff
@@ -361,22 +351,33 @@ workflows:
 
   test-bazel7-linux:
```
sbc100: Can you remove the '7' here and below?

mmorel-35: I can but the naming with 7 is required to pass and they won't exist when I do

sbc100: I just removed those requirements.

Also, I wonder if we can/should keep these as separate top level tasks?  I'm not which is better?

mmorel-35: I’m not sure, that something that deserves to be discussed.

Just an idea, reduce the top level to 

- build-docker-image
- lint
- test (test-linux, test-linux-arm64,...)
- test-bazel (test-bazel-linux, test-bazel-mac-arm64,... )

Vertexwahn:
@mmorel-35 Thanks!

PR #1670, opened 2026-02-09
Add caching for Python lint dependencies in CI configuration

mmorel-35:
#### Description

Add caching for Python lint dependencies in CI configuration


.circleci/config.yml
```diff
@@ -79,16 +79,23 @@ jobs:
...
+      - restore_cache:
+          keys:
+            - v1-lint-deps-{{ arch }}-{{ checksum "requirements-dev.txt" }}
       - run:
           name: install python deps
           command: |
             apt-get update -q
             apt-get install -q -y python3-pip
+            python3 -m pip install --upgrade pip
+            python3 -m pip install -r requirements-dev.txt
```
sbc100: Is there some way that we can completely skip this step if/when there is a cache hit?  i.e. avoid running pip completely?

sbc100: Is there some way that we can completely skip this step if/when there is a cache hit?  i.e. avoid running pip completely?

sbc100: Is there some way that we can completely skip this step if/when there is a cache hit?  i.e. avoid running pip completely?

sbc100: Is there some way that we can completely skip this step if/when there is a cache hit?  i.e. avoid running pip completely?

sbc100, line 91: Is there some way that we can completely skip this step if/when there is a cache hit?  i.e. avoid running pip completely?

PR #1671, opened 2026-02-09
[CI] Reuse setup-docker steps

mmorel-35:
Also rename build-docker-image workflow to docker as not just docker build.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,6 +40,18 @@ commands:
           command: |
             source emsdk_env.sh
             test/test.py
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -340,7 +334,7 @@ workflows:
   test-windows:
     jobs:
       - test-windows
-  build-docker-image:
```
sbc100, line 343: When users normally see this job it *is* just building of the images since the publishing part only happens on main.   But not harm renaming it I think.

Issue #1675, opened 2026-02-20
Bazel toolchain uses system python on MacOS


DoDoENT:
Instead of using the Python 3.13 from `rules_python`, it uses system python (not even the homebrew version).

Since MacOS still uses Python 3.9.6, this causes build failures when using Emscripten 4.0.22 or newer:

```
ERROR: /Users/dodo/Library/Caches/bazel/_bazel_dodo/d37a55ce2ed2abf9c807bae8c8d24add/external/googletest+/BUILD.bazel:102:11: Compiling googlemock/src/gmock.cc failed: (Exit 1): emcc.sh failed: error executing CppCompile command (from cc_library rule target @@googletest+//:gtest)
  (cd /Users/dodo/Library/Caches/bazel/_bazel_dodo/d37a55ce2ed2abf9c807bae8c8d24add/sandbox/darwin-sandbox/649/execroot/_main && \
  exec env - \
    EMCC_WASM_BACKEND=1 \
    EM_BIN_PATH=external/emsdk++emscripten_deps+emscripten_bin_mac_arm64 \
    EM_CONFIG_PATH=external/emsdk++emscripten_cache+emscripten_cache/emscripten_config \
    NODE_JS_PATH=external/rules_nodejs++node+nodejs_darwin_arm64/bin/node \
    PATH=/bin:/usr/bin:/usr/local/bin \
    PWD=/proc/self/cwd \
  external/emsdk+/emscripten_toolchain/emcc.sh @bazel-out/wasm-fastbuild-ST-f416d161dfb4/bin/external/googletest+/_objs/gtest/gmock.o.params)
# Configuration: e22466f04beb073adc43c23b7e8a6ea4ea7fa8665f522c3b4b8b22ce82e3bd2f
# Execution platform: //platforms/macos:arm64

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
Traceback (most recent call last):
  File "/Users/dodo/Library/Caches/bazel/_bazel_dodo/d37a55ce2ed2abf9c807bae8c8d24add/sandbox/darwin-sandbox/649/execroot/_main/external/emsdk++emscripten_deps+emscripten_bin_mac_arm64/emscripten/emcc.py", line 32, in <module>
    from tools import (
[...]

PR #1676, opened 2026-02-22
[bazel] Remove dependency on system python

DoDoENT:
Thanks to @rickeylev for the help with figuring out the correct way to ensure the Python interpreter with all its dependencies is correctly added to the toolchain files.

Note: this does not touch Windows support, which may also have a similar issue. I see windows bat files using `py -3`, which may also not be hermetic, but I have no access to any Windows machines to test and fix that, so I'll leave that untouched for now.

Fixes: #1675, #1642

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -374,8 +374,7 @@ workflows:
               bazel_version: 
                 - "7.x"
                 - "8.x"
-              # https://github.com/emscripten-core/emsdk/issues/1642
-              # - "9.x"
+                - "9.x"
   test-bazel-windows:
     jobs:
       - test-bazel-windows:


bazel/emscripten_toolchain/emar.sh
```diff
@@ -2,4 +2,4 @@
 
 source $(dirname $0)/env.sh
 
-exec python3 $EMSCRIPTEN/emar.py "$@"
+exec $EMSDK_PYTHON $EMSCRIPTEN/emar.py "$@"
```
sbc100, line 5: I wonder if could now maybe remove these custom launcher scripts and just use the standard emcc launcher shell script (as a followup, no in this PR)

I guess it depends what else in env.sh is needed?

DoDoENT: I actually tried that in one of my unsuccessful iterations. I replaced all those scripts with pure Python, presented them to Bazel via `py_binary`, and then gave those binaries to the `cc_toolchain`.

The first problem was that binaries were resolved in the "target context (WASM)" instead of the "exec" context, which then tried to run the "WASM version of Python", which doesn't exist. I circumvented that by creating a transitioned binary in `exec` context, and that almost worked, except final compilation attempts failed because `execv("/path/to/python/wrapper")` failed. I didn't investigate deeply enough to find out what was wrong there, and instead backtracked to pursuing adding Python interpreter runfiles to the toolchain context, so they end up in the sandbox.

My first working solution required a patch to `rules_python`, which I created a PR for, but the `rules_python` author then told me this wasn't how it is supposed to be used and instructed me to research the Python exec toolchain. I did that, and this solution now works, so I created this PR.

In the future, I'll probably revisit this approach that uses pure Python without any shell wrappers, but I still need to get better at Bazel before that (I'm still learning it).

mmorel-35:
Can you add bazel 9 (macOs) testing in CircleCI with that change ?

.circleci/config.yml
```diff
@@ -374,8 +374,7 @@ workflows:
               bazel_version: 
                 - "7.x"
                 - "8.x"
-              # https://github.com/emscripten-core/emsdk/issues/1642
-              # - "9.x"
+                - "9.x"
```
mmorel-35, line 377: @sbc100 ,
This is also closing 1642 issue? Shall Closes #xxx be added to the description and maybe several others ?

sbc100: Yes, please update the PR description (we use `Fixes: #xxx`)

PR #1682, opened 2026-02-26
[Bazel] Replace uses of deprecated bazel_tools things

stevenlr:
Using the C++ rules, libraries and utilities from bazel_tools is being deprecated in favor of using rules_cc.

~Similarly the `@bazel_tools//src/conditions:host_windows` constraint has been deprecated for a while and will be removed with Bazel 10.~

### DIFF
--- a/bazel/remote_emscripten_repository.bzl
+++ b/bazel/remote_emscripten_repository.bzl
@@ -141,7 +141,7 @@ def create_toolchains(name, repo_name, exec_compatible_with):
         target_compatible_with = ["@platforms//cpu:wasm32"],
         exec_compatible_with = exec_compatible_with,
         toolchain = cc_wasm_target,
-        toolchain_type = "@bazel_tools//tools/cpp:toolchain_type",
+        toolchain_type = "@rules_cc//cc:toolchain_type",
     )
 
     cc_toolchain_suite(
--- a/bazel/MODULE.bazel.lock
+++ b/bazel/MODULE.bazel.lock
@@ -1,5 +1,5 @@
 {
-  "lockFileVersion": 13,
+  "lockFileVersion": 26,
   "registryFileHashes": {
     "https://bcr.bazel.build/bazel_registry.json": "8a28e4aff06ee60aed2a8c281907fb8bcbf3b753c91fb5a5c57da3215d5b3497",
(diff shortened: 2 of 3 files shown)


bazel/remote_emscripten_repository.bzl
```diff
@@ -55,6 +55,8 @@ def create_toolchains(name, repo_name, exec_compatible_with):
     repo_linker_files_target = remote_repo + ":linker_files"
     repo_ar_files_target = remote_repo + ":ar_files"
 
+    script_extension = "bat" if "@platforms//os:windows" in exec_compatible_with else "sh"
```
sbc100: Can we not continue to use a select statement based on the host os?

Doing a sub-string search here seems a little strange.. is this really the best way?

Also, I don't love pythons single line if/else personally.  Could you split this into a normal if/else? (unless this is recommended pattern in bazel files?)

stevenlr: Indeed this could be a bit flaky if we don't use the canonical constraint names in `exec_compatible_with`.

But no, we can't use select because there is no well-known constraint on the host configuration anymore. Technically we could have reimplemented the old `@bazel_tools//src/conditions:host_windows`. But it has been phased-out because Bazel encourages us to use their toolchains system, despite it making some things like our case here difficult.

Other projects have adopted different solutions (see https://github.com/bazelbuild/bazel-skylib/blob/56a2abbaf131332835ab2721a258ea3c763a7178/rules/private/copy_file_private.bzl#L117), but I just found that the platforms module allows us to fetch the host constraints in their canonical format, so that's convenient, and fairly simple.

Another simple solution would have been to give the script extension in the call to `create_toolchains`. Fairly low-tech. If you think it's better because simpler I can change that.

sbc100: Perhaps you could revert his part of the change for now, pending further discussion?  The other part looks much most straight forward.

stevenlr: Done in latest version. I'll send another PR for the host constraints tomorrow.

stevenlr:
I believe the test failure here was just a flake, it seems the download of a Baze module failed somehow.

PR #1683, opened 2026-02-27
[Bazel] Replace the use of @bazel_tools//src/conditions:host_windows

stevenlr:
This is a followup of #1682.

This has been deprecated and will be removed in Bazel 10.

This implementation of the platform information retrieval is similar to what is done in Skylib (see https://github.com/bazelbuild/bazel-skylib/blob/56a2abbaf131332835ab2721a258ea3c763a7178/rules/private/copy_file_private.bzl#L117) however this does not use the experimental platform API.

- We define a provider that gives the script extension for a given platform and a rule to instantiate it.
- We define a target with the `platform_info` rule that selects the correct settings based on constraints from `@platform`.
- During toolchain config creation we inject that target **using the `exec` config**.
- We can then retrieve the platform information in the toolchain config (namely the script extension here).

Note that the platform information could have given a simple `is_windows` flag, but I chose this implementation because it matches the previous one more closely.

Note also that it was not possible to keep the `script_extension` attribute as string, because the `cfg` field is not supported on string attributes.

### DIFF
--- a/bazel/emscripten_toolchain/BUILD.bazel
+++ b/bazel/emscripten_toolchain/BUILD.bazel
@@ -2,11 +2,21 @@ load("@rules_cc//cc:defs.bzl", "cc_library")
 load("@rules_python//python:py_binary.bzl", "py_binary")
 load("//:emscripten_deps.bzl", "emscripten_repo_name")
 load("//:remote_emscripten_repository.bzl", "create_toolchains", "emscripten_toolchain_name")
+load(":platform_info.bzl", "platform_info")
 
 package(default_visibility = ["//visibility:public"])
 
 # dlmalloc.bc is implicitly added by the emscripten toolchain
 cc_library(name = "malloc")
+
+platform_info(
+    name = "platform_info",
+    script_extension = select({
+        "@platforms//os:windows": ".bat",
+        "//conditions:default": ".sh",
+    }),
+    visibility = [":__subpackages__"],
+)
 
 create_toolchains(
     name = emscripten_toolchain_name("linux"),
--- a/bazel/emscripten_toolchain/platform_info.bzl
+++ b/bazel/emscripten_toolchain/platform_info.bzl
@@ -1,0 +1,22 @@
+"""Defines a provider and rule for platform information.
+"""
+
+PlatformInfo = provider(
+    doc = "Provides some info about a platform",
+    fields = {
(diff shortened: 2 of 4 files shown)


stevenlr:
Both Bazel and its ecosystem have been [moving away from depending on Bash](https://bazel.build/configure/windows#using-bazel-without-bash) on Windows. So it's not really recommended.

I've seen [the discussion on a previous PR](https://github.com/emscripten-core/emsdk/pull/1676#discussion_r2839909249) about using `py_binary`s for Emscripten scripts. However doing this properly would require quite a bit of infrastructure and I would actually recommend doing it in Emscripten directly instead of here, because the infrastructure would depend on the Emscripten version (because you would have to also define py_librarys, dependencies between them, data targets, handle runfiles correctly, etc, quite a bit of work!).

A simpler approach could be to use the Python interpreter as the tool for each cc toolchain action, and the first argument would be the script (emar.py, emcc.py, etc). That way we could remove the wrapper scripts but still keep all Bazel stuff in this repo.

I could take a swing at it this week-end or early next week, see where it leads, but it's a separate topic anyway.

stevenlr:
Hi, I've been trying to remove the need for the batch and bash files, however I don't think it's doable. It seems like Emscripten expects the `EM_CONFIG` environment variable to have an absolute path. Bazel doesn't allow us to do that because of reproducibility. I don't see any way of avoiding some kind of wrapper that does non-hermetic stuff. Sure we could make the wrappers Python (it would be platform-independent), or even make it only 1 wrapper instead of 4, but I'm not sure the complexity is worth it.

bazel/emscripten_toolchain/BUILD.bazel
```diff
@@ -2,12 +2,22 @@ load("@rules_cc//cc:defs.bzl", "cc_library")
...
 package(default_visibility = ["//visibility:public"])
 
 # dlmalloc.bc is implicitly added by the emscripten toolchain
 cc_library(name = "malloc")
 
+platform_info(
+    name = "platform_info",
+    script_extension = select({
+        "@platforms//os:windows": "bat",
+        "//conditions:default": "sh",
```
sbc100: Can you include the `.` here.   I.e. the extension should be `.sh` and then it should appended to the bare filename.  This just fits with how python normally does file extensions stuff.

stevenlr: Done in latest version

PR #1695, opened 2026-04-11
Add EMSDK_USE_CURL environment variable to force use of curl

ishanjain28:
Corporate environments may force use of a custom CA. `emsdk` uses urllib on all platforms and curl on macos. However, urllib does not have a single universal variable similar to the recently introduced `NODE_USE_SYSTEM_CA` in nodejs that can be set to make it use system CA rather than the included certificate store.

There is an option to use custom CA with urllib by setting some environment variables but there are 2 problems with it. 
1. It requires a dependency to get the certificate store. (This can also be hardcoded if I know the platform)
2. I _think_ I cannot use multiple certificates this way because the environment variables seem to accept a  path to a single file.

It's easier to install curl and use that.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -720,11 +720,11 @@ def download_file(url, dstpath, download_even_if_exists=False,
   mkdir_p(os.path.dirname(file_name))
 
   try:
-    # Use curl on macOS to avoid CERTIFICATE_VERIFY_FAILED issue with
-    # python's urllib:
+    # Use curl on macOS or when EMSDK_USE_CURL is set to avoid
+    # CERTIFICATE_VERIFY_FAILED issue with python's urllib:
     # https://stackoverflow.com/questions/40684543/how-to-make-python-use-ca-certificates-from-mac-os-truststore
     # Unlike on linux or windows, curl is always available on macOS systems.
-    if MACOS:
+    if MACOS or 'EMSDK_USE_CURL' in os.environ:
       download_with_curl(url, file_name)
     else:
       download_with_urllib(url, file_name)


sbc100 (approved):

sbc100:
Maybe we could add an FAQ entry about this?  

What platform are you using where you found this was needed?  Window or linux?

ishanjain28:
@sbc100 This was on Linux, Ubuntu 22.04 / 24.04 and Debian 13. All environments required trusting corporate CA for TLS connections.

sbc100:
> @sbc100 This was on Linux, Ubuntu 22.04 / 24.04 and Debian 13. All environments required trusting corporate CA for TLS connections.

Are you saying the CA file that ships with python on Ubuntu 22.04 / 24.04 and Debian 13 is unable to connect to google cloud storage?   That would seem strange to me.  Is there something odd about the way you set it up?  Or are you not using the default python install?

sbc100:
Oh sorry I missed the "Corporate environments may force use of a custom CA" part.

sbc100:
Seems sad/strange that python cannot be made to use system CA chain.

sbc100:
Is there perhaps an FAQ entry for python/urllib about this that we can link to in our docs?

sbc100:
Some googling tell ms that it may be enough to do set `SSL_CERT_FILE` to the full path of your corporate .pem certificate bundle.  Is that an alternative way to acheive this?

ishanjain28:
> Some googling tell ms that it may be enough to do set SSL_CERT_FILE to the full path of your corporate .pem certificate bundle. Is that an alternative way to acheive this?

Yess, that is an option but there are 2 issues. 

1. This environment variable takes a single file. In the same CI pipeline, _someone_ may be making requests to domains that don't require the corporate CA but will work with the default CA. So, They have to set the variable to corporate CA and then unset it. I personally don't _need_ to do this and just pointing to corporate CA is likely sufficient. 
2. I'll have to add another dependency(to reliably get the path to certificate store. I don't want to hard code it). I don't really want to do that if I can make it use curl 😄  


I'll find urllib docs and also update emsdk docs page 👍

sbc100:
SGTM.  Thanks for the PR.

ishanjain28:
hi @sbc100, Should I put this in the readme in this repo or the emscription docs website? (I am not sure where/how to update the docs website right now)

sbc100:
I think perhaps it would be best as an FAQ entry in the emsdk repo.   Maybe here https://github.com/emscripten-core/emsdk?tab=readme-ov-file#sdk-maintenance

PR #1696, opened 2026-04-13
Create a simple bool env variable parser

joeedh:
This little PR creates a unified way to parse environmental variable bools.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -47,9 +47,10 @@ download_dir = 'downloads/'
 
 extra_release_tag = None
 
-def get_env_boolean(b):
-  env_var = os.getenv(b) or ''
-  return env_var == "1" or env_var.lower() == 'true'
+def get_env_boolean(name):
+  env_var = os.getenv(name)
+  assert env_var in {None, '1', '0'}, f'invalid environment variable setting ${env_var} for ${name}'
+  return env_var == '1'
 
 
 # Enable this to do very verbose printing about the different steps that are
--- a/emsdk.py
+++ b/emsdk.py
@@ -47,6 +47,7 @@ download_dir = 'downloads/'
 
 extra_release_tag = None
 
+
 def get_env_boolean(name):
   env_var = os.getenv(name)
   assert env_var in {None, '1', '0'}, f'invalid environment variable setting ${env_var} for ${name}'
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -21,6 +21,11 @@
 import zipfile
 from collections import OrderedDict
 
+def parseBool(b, default=False):
+  if b is None:
+    return default
+  return b == "1" or str(b).lower() == 'true'
```
sbc100: How calling this function `getEnvBoolean` and including the `os.getenv` in this function.

Also, I don't think you need the `if == None` check since None would fail the check on line 27.

sbc100: Move this new function down to next to its first use.. or at least after all of the imports and the version_info check (which should come before any other code).

emsdk.py
```diff
@@ -170,7 +170,7 @@ def exit_with_error(msg):
 ENABLE_LLVM_ASSERTIONS = 'auto'
 
 # If true, keeps the downloaded archive files.
-KEEP_DOWNLOADS = get_bool_env('EMSDK_KEEP_DOWNLOADS')
+KEEP_DOWNLOADS = getEnvBoolean('EMSDK_KEEP_DOWNLOADS')
```
sbc100: Sorry can you use the `camel_case` name..  sorry I suggested the wrong name first :(

emsdk.py
```diff
@@ -47,11 +47,16 @@
 
 extra_release_tag = None
 
+def getEnvBoolean(b, default=False):
+  env_var = os.getenv(b) or str(default)
```
sbc100: Is there are not users of `default` then just remove it.

emsdk.py
```diff
@@ -47,11 +47,16 @@
...
+  return env_var == "1" or env_var.lower() == 'true'
```
sbc100: How about just `return env_var in {'1', 'true'}`?

Actually do we really need to support `true`?

How about just:

```
assert env_var in {None, '1', '0'}, f'invalid environment variable setting for ${name}'
return evn_var == '1'
```

emsdk.py
```diff
@@ -47,8 +47,8 @@
 
 extra_release_tag = None
 
-def get_env_boolean(b, default=False):
-  env_var = os.getenv(b) or str(default)
+def get_env_boolean(b):
+  env_var = os.getenv(b) or ''
```
sbc100: You can just write this as `os.getenv(b, '')`

sbc100 (approved):
Nice!  Thanks for your patience.

PR #1702, opened 2026-04-25
Add new env. var. EMSDK_RETRY_CLEAN_BUILD=1

juj:
Add new env. var. EMSDK_RETRY_CLEAN_BUILD=1, which automatically re-does a cmake config+make from a clean slate, if incremental build failed. This helps mitigate against the frequent LLVM repository issues that break incremental building.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1127,6 +1127,23 @@ def get_generator_and_config_args(tool):
   return (cmake_generator, args)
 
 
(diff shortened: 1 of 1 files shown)


sbc100 (approved):

juj:
Failures are all network outages, I'll land this presuming it's all good.

PR #1704, opened 2026-04-28
[CI] Skip building binaryen from source on macOS

sbc100:
Our macOS CI instance is too old to build binaryen from source.

See https://github.com/WebAssembly/binaryen/pull/8609

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -277,6 +277,10 @@ int main() {
     if WINDOWS:
       # It takes over 30 mins to build binaryen using Visual Studio in CI
       self.skipTest('test is too slow under windows')
+    if MACOS:
+      # Our CI uses and older version of macOS which only as 14.3.1
+      # TODO: Remove this if/when we update the emsdk CI.
+      self.skipTest('building binaryen requries xcode 15.3 because it uses std::ranges')
     run_emsdk(['install', '--build=Release', 'binaryen-main-64bit'])
 
   def test_no_32bit(self):


sbc100:
@juj, what is the minimum version of Xcode that you need to support building binaryen using?

We just started using std::ranges in binaryen which seems to require Xcode 15.3.   If that is too new then we would need to look into reverting I guess?

juj:
Looking at our minimum target, the minimum macOS version for Unity developers for our next upcoming Unity release is macOS 13. Apple's [Xcode support page](https://developer.apple.com/support/xcode/) says this would correspond to Xcode 15.2:

image

I wonder how feasible it would be to take this as minimum level of support?

There is the `-DCMAKE_OSX_DEPLOYMENT_TARGET=` directive in CMake, which supposedly should be able to be used to compile Binaryen on a newer system, while targeting an older system. But if possible, I would like to avoid needing to rely on that, and build on the same image where we do our tests at.

sbc100:
> Looking at our minimum target, the minimum macOS version for Unity developers for our next upcoming Unity release is macOS 13. Apple's [Xcode support page](https://developer.apple.com/support/xcode/) says this would correspond to Xcode 15.2:
> 
> image
> I wonder how feasible it would be to take this as minimum level of support?
> 
> There is the `-DCMAKE_OSX_DEPLOYMENT_TARGET=` directive in CMake, which supposedly should be able to be used to compile Binaryen on a newer system, while targeting an older system. But if possible, I would like to avoid needing to rely on that, and build on the same image where we do our tests at.

Yes, I think `CMAKE_OSX_DEPLOYMENT_TARGET` is good solution here.

I would hope that there would be no need for the machine building your binaries to be running the oldest support macOS version for the target binaries.   Do you actually do your binary building on macOS 13 today, or do you use `CMAKE_OSX_DEPLOYMENT_TARGET` on a more recent system?

sbc100:
emsdk already builds all cmake projects with `-DCMAKE_OSX_DEPLOYMENT_TARGET=11.0`, and we are not proposing to change that right now.

sbc100:
Binaryen itself also already has `set(CMAKE_OSX_DEPLOYMENT_TARGET "10.15" CACHE STRING "Minimum OS X deployment version")`

juj:
> Do you actually do your binary building on macOS 13 today, or do you use `CMAKE_OSX_DEPLOYMENT_TARGET` on a more recent system?

We do reuse the current emsdk installs with the `DCMAKE_OSX_DEPLOYMENT_TARGET=` directives that are present there. 

Currently we build Mac X64 from a macOS 10.14 system, and Mac ARM from a macOS 13 system. If I read the table correctly, this would mean that building Emscripten from source would require macOS 14 at minimum, for that Xcode 15.2.

sbc100:
> > Do you actually do your binary building on macOS 13 today, or do you use `CMAKE_OSX_DEPLOYMENT_TARGET` on a more recent system?
> 
> We do reuse the current emsdk installs with the `DCMAKE_OSX_DEPLOYMENT_TARGET=` directives that are present there.
> 
> Currently we build Mac X64 from a macOS 10.14 system, and Mac ARM from a macOS 13 system. If I read the table correctly, this would mean that building Emscripten from source would require macOS 14 at minimum, for that Xcode 15.2.

Hmm, yes I think that unless we revert that means that binaryen would require macOS 14 to build from source.

Can you maybe confirm that binaryen `main` does or doesn't build on your systems @juj?  

Assuming it doesn't then I think we probably need to revert the binaryen change, and add some CI testing on older macos.

Given that emsdk currently sets `-DCMAKE_OSX_DEPLOYMENT_TARGET=11.0` how can tha work if you build for X64 on macOS 10.14 ?  i.e. isn't it somehow wrong to set the deployment target to a version of macOS newer than the one you are running.

sbc100:
It seems that binaryen would like to move forward with C++20 support, not just for ranges but also for co-routines.

I guess we should block the emsdk release in https://github.com/emscripten-core/emsdk/pull/1705, until this has been resolved?

@juj, are you able to perform the building of your SDK on at recent macOS machine (relying on the existing `-DCMAKE_OSX_DEPLOYMENT_TARGET` to ensure compatibility?)

juj:
> @juj, are you able to perform the building of your SDK on at recent macOS machine (relying on the existing `-DCMAKE_OSX_DEPLOYMENT_TARGET` to ensure compatibility?)

I'll try to give this a look asap next week.

juj:
Commented in https://github.com/WebAssembly/binaryen/pull/8657#issuecomment-4372227162

sbc100:
Looks like binaryen is still failing to build here on the older macOS CI machine:

```
/Users/distiller/project/binaryen/main/src/support/name.h:35:12: error: call to implicitly-deleted default constructor of 'wasm::IString'
  Name() : IString() {}
           ^
/Users/distiller/project/binaryen/main/src/support/istring.h:72:3: note: explicitly defaulted function was implicitly deleted here
  IString() = default;
  ^
/Users/distiller/project/binaryen/main/src/support/istring.h:68:14: note: default constructor of 'IString' is implicitly deleted because field 'str' of const-qualified type 'const wasm::IString::View' would not be initialized
  const View str;
 ```

@kripken @tlively can we address this one?

kripken:
@sbc100 was that fixed by WebAssembly/binaryen#8719 ?

sbc100:
Yes, although another issue crept in so now we have to wait for https://github.com/WebAssembly/binaryen/pull/8725 too

PR #1707, opened 2026-04-30
Update Github ci actions to latest versions

mcbarton:
The actions being used in the Github ci are outdated. These cis are obviously not broken, but thought it might be worth doing. I've been told its best to use a hash instead of a general version to make it use an immutable version. I have added a comment next to each one to make it more obvious as to what version the action is pinned to.


mcbarton:
image

image


This PR should also fix these warnings when the ci run.

sbc100 (approved):

PR #1709, opened 2026-05-18
Add more ruff checks

sbc100:
This caught as few bugs form the f-string conversion.

### DIFF
--- a/bazel/emscripten_toolchain/link_wrapper.py
+++ b/bazel/emscripten_toolchain/link_wrapper.py
@@ -38,7 +38,7 @@ if any(' ' in a for a in param_file_args):
   sys.argv[1] = '@' + new_param_filename
 
 emcc_py = os.path.join(os.environ['EMSCRIPTEN'], 'emcc.py')
-rtn = subprocess.call([sys.executable, emcc_py] + sys.argv[1:])
+rtn = subprocess.call([sys.executable, emcc_py, *sys.argv[1:]])
 if rtn != 0:
   sys.exit(1)
 
@@ -157,7 +157,7 @@ if not files:
   sys.exit(1)
 
 # cc_binary must output exactly one file; put all the output files in a tarball.
-cmd = ['tar', 'cf', base_name + '.tar'] + files
+cmd = ['tar', 'cf', base_name + '.tar', *files]
 subprocess.check_call(cmd, cwd=outdir)
 os.replace(os.path.join(outdir, base_name + '.tar'), output_file)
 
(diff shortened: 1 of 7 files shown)


juj:
> This caught as few bugs form the f-string conversion.

What was the bug?

bazel/emscripten_toolchain/link_wrapper.py
```diff
@@ -38,7 +38,7 @@
   sys.argv[1] = '@' + new_param_filename
 
 emcc_py = os.path.join(os.environ['EMSCRIPTEN'], 'emcc.py')
-rtn = subprocess.call([sys.executable, emcc_py] + sys.argv[1:])
+rtn = subprocess.call([sys.executable, emcc_py, *sys.argv[1:]])
```
juj, line 41: What's the idea with these changes? Was there something wrong here?

sbc100:
> > This caught as few bugs form the f-string conversion.
> 
> What was the bug?

There were a few strings used template var but didn't have the `f` prefix on them.    My fault I imagine.

bazel/emscripten_toolchain/link_wrapper.py
sbc100: Apparently its more efficient that using `+=` on lists:  https://docs.astral.sh/ruff/rules/collection-literal-concatenation/.

We could suppress this warning instead since none of this code is very perf sensitive, but I'd like to train myself so always prefer the better form.

emsdk.py
```diff
@@ -1030,8 +1030,8 @@ def quote_parens(x):
     open(os.path.join(build_root, 'recmake.' + ('bat' if WINDOWS else 'sh')), 'w').write(' '.join(map(quote_parens, cmdline)))
     ret = subprocess.check_call(cmdline, cwd=build_root, env=build_env())
     if ret != 0:
-      errlog('CMake invocation failed with exit code {ret}!')
-      errlog('Working directory: ' + build_root)
+      errlog(f'CMake invocation failed with exit code {ret}')
```
sbc100, line 1033: This line, for example, was missing the f-prefix.

emsdk.py
```diff
@@ -1528,7 +1528,7 @@ def build_binaryen_tool(tool):
 
 
 def download_and_extract(archive, dest_dir, filename_prefix='', clobber=True):
-  debug_print('download_and_extract(archive={archive}, dest_dir={dest_dir})')
+  debug_print(f'download_and_extract(archive={archive}, dest_dir={dest_dir})')
```
sbc100, line 1531: And this one.

juj: Got it.

juj (approved):

PR #1710, opened 2026-05-29
Replace optional Tools fields (using `hasattr`) with normal class fields. NFC

sbc100:
Also remove `version_filter` completely.  The last usage of this field was removed in #1165.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -870,7 +870,7 @@ def llvm_build_dir(tool):
   generator_suffix = cmake_generator_prefix()
   bitness_suffix = '_32' if tool.bitness == 32 else '_64'
 
-  if hasattr(tool, 'git_branch'):
+  if tool.git_branch:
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -1884,57 +1904,52 @@ def installation_dir(self):
...
       return []
 
-    return self.expand_vars(activated_env).split(';')
+    return self.expand_vars(self.activated_env).split(';')
 
   def compatible_with_this_arch(self):
-    if hasattr(self, 'arch'):
-      if self.arch != ARCH:
-        return False
+    if self.arch and self.arch != ARCH:
```
dschuff: could even be shortened to `return not self.arch or self.arch == ARCH` or some equivalent.

dschuff (approved):

emsdk.py

Issue #1714, opened 2026-06-01
bazel build seems to have having trouble with the trasition to true shared libraries.


sbc100:
I'm seeing errors like this trying to integrate the latest emscripten changes that include https://github.com/emscripten-core/emscripten/pull/25930

```
WARNING: For repository 'rules_cc', the root module requires module version rules_cc@0.2.16, but got rules_cc@0.2.17 in the resolved dependency graph. Please update the version in your MODULE.bazel or set --check_direct_dependencies=off
INFO: Analyzed target //long_command_line:long_command_line_wasm (3 packages loaded, 6209 targets configured).
ERROR: /root/project/bazel/test_external/long_command_line/BUILD.bazel:48:10: Linking long_command_line/liblong_command_line.so failed: (Exit 1): emcc_link.sh failed: error executing CppLink command (from cc_binary rule target //long_command_line:long_command_line) external/emsdk+/emscripten_toolchain/emcc_link.sh @bazel-out/wasm-fastbuild-ST-97b75ba167b6/bin/long_command_line/liblong_command_line.so-0.params

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
wasm-ld: error: bazel-out/wasm-fastbuild-ST-97b75ba167b6/bin/long_command_line/liblong_command_line_file01_a.a(long_command_line_file01.o): relocation R_WASM_MEMORY_ADDR_SLEB cannot be used against symbol `std::__2::cout`; recompile with -fPIC
wasm-ld: error: bazel-out/wasm-fastbuild-ST-97b75ba167b6/bin/long_command_line/liblong_command_line_file01_a.a(long_command_line_file01.o): relocation R_WASM_MEMORY_ADDR_SLEB cannot be used against symbol `.L.str`; recompile with -fPIC
wasm-ld: error: bazel-out/wasm-fastbuild-ST-97b75ba167b6/bin/long_command_line/liblong_command_line_file01_a.a(long_command_line_file01.o): relocation R_WASM_MEMORY_ADDR_SLEB cannot be used against symbol `std::__2::ctype<char>::id`; recompile with -fPIC
wasm-ld: error: bazel-out/wasm-fastbuild-ST-97b75ba167b6/bin/long_command_line/liblong_command_line_file02_a.a(long_command_line_file02.o): relocation R_WASM_MEMORY_ADDR_SLEB cannot be used against symbol `std::__2::cout`; recompile with -fPIC
[...]

PR #1715, opened 2026-06-01
[bazel] Remove `linkshared = True` from bazel/test_external/long_command_line

sbc100:
I'm not sure why this was originally added as part of  #1373, but its breaking with our transition to using true dynamic linking because bazel currently does not know how to include PIC versions of the standard libraries. 

See #1714


sbc100:
This is currently blocking the next emscripten release so I'm thinking of landing this as TBR

PR #1718, opened 2026-06-02
Use version information from Workflow trigger to set version

dschuff:
This allows passing the version from Emscripten-releases.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -9,6 +9,9 @@ on:
       nonlto-sha:
         required: true
         type: string
+      version:
+        required: false
+        type: string
 
 jobs:
   create-release-pr:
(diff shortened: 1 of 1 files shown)


dschuff:
This, combined with https://chromium-review.git.corp.google.com/c/emscripten-releases/+/7891806, would fix
https://github.com/emscripten-core/emsdk/pull/1712

.github/workflows/create-release.yml
```diff
@@ -9,6 +9,9 @@ on:
       nonlto-sha:
         required: true
         type: string
+      version:
+        required: false
```
dschuff: Actually, should we just make the version information mandatory in the trigger?

sbc100 (approved):

PR #1724, opened 2026-06-25
[bazel] Remove use of deprecated `-s USE_PTHREADS=1` flag

DoDoENT:
This will silence linker warnings when building binaries with enabled threads.

### DIFF
--- a/bazel/emscripten_toolchain/toolchain.bzl
+++ b/bazel/emscripten_toolchain/toolchain.bzl
@@ -569,12 +569,6 @@ def _impl(ctx):
             features = ["crosstool_cpu_asmjs"],
         ),
         flag_set(
-            actions = all_compile_actions +
-                      all_link_actions,
-            flags = ["-sUSE_PTHREADS"],
-            features = ["use_pthreads"],
-        ),
-        flag_set(
             actions = all_link_actions,
             flags = ["-sEXIT_RUNTIME"],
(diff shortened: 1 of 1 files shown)


bazel/emscripten_toolchain/toolchain.bzl
```diff
@@ -568,12 +568,6 @@ def _impl(ctx):
             flags = ["-s", "WASM=0"],
             features = ["crosstool_cpu_asmjs"],
         ),
-        flag_set(
-            actions = all_compile_actions +
-                      all_link_actions,
-            flags = ["-s", "USE_PTHREADS=1"],
-            features = ["use_pthreads"],
```
sbc100, line 575: You could just change the flag to `-pthread` ?

DoDoENT: The `-pthread` flag is already given a couple of lines below.

The original author obviously added both: `USE_PTHREADS=1` always and `-pthread` only when using the LLVM backend. I guess the toolchain was written when emscripten still supported the fastcomp backend.

sbc100: OK to land this then?

DoDoENT: Yes. I'm already using this patch in production at my work.

sbc100 (approved):

PR #1725, opened 2026-06-25
[bazel] Remove EMCC_WASM_BACKEND

sbc100:
This environment variable has not existed for a long time now.

### DIFF
--- a/bazel/emscripten_toolchain/toolchain.bzl
+++ b/bazel/emscripten_toolchain/toolchain.bzl
@@ -1093,14 +1093,6 @@ def _impl(ctx):
                 ),
             ],
         ),
-        # Use llvm backend.  Off by default, enabled via --features=llvm_backend
-        env_set(
-            actions = all_compile_actions +
-                      all_link_actions +
-                      [ACTION_NAMES.cpp_link_static_library],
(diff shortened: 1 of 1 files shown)


DoDoENT (approved):
This could possibly be couple with removal of `llvm_backend` feature. It has been enabled by default for some time now and I think emscripten no longer supports `fastcomp` anyway.

But, we can also do that in a separate PR.

sbc100:
Yes, a followup to remove fastcomp references would be great.   I guess we still might want to support WASM=0 though?  (what we call asmjs here?)

DoDoENT:
> Yes, a followup to remove fastcomp references would be great. I guess we still might want to support WASM=0 though? (what we call asmjs here?)

Let's do it in two steps:

- first one PR that removes anything related to fastcomp
- second, which adds support for asmjs builds if anyone actually needs that with bazel (I never tried that, so I'll have to first learn how official emscripten's build system handles this).

sbc100:
Sounds good.  Would you have time to send a PR for phase 1?

DoDoENT:
> Would you have time to send a PR for phase 1?

I would be glad to. Probably during the weekend, if I find time.

PR #1726, opened 2026-06-25
[bazel] Use standard/compact `-s` flags

sbc100:


DoDoENT (approved):
Oh, yes...

How many sleepless nights I used to have before emscripten supported compact flags and I had to work with cmake...

PR #1728, opened 2026-06-27
[bazel] add support for enabling relaxed SIMD in wasm_cc_binary

DoDoENT:
It would probably be more elegant to reuse the existing `simd` attribute and change it's type from boolean into string enum (similar like it's for `threads`), but that would be a breaking change so I decided to add additional boolean acknowlidging that `simd=False`, `relaxed_simd=True` makes no sense and will also enable `simd` as `relaxed_simd` implies `simd`.

If you are comfortable with me breaking the API, I'd be happy to do it (see [how I did it in my wrapper](https://github.com/DoDoENT/bazel-playground/pull/62/changes#diff-f8403d920861efa3471626c055b696f82cf2e5d20c5f584a991576fcc6178a05))

### DIFF
--- a/bazel/emscripten_toolchain/toolchain.bzl
+++ b/bazel/emscripten_toolchain/toolchain.bzl
@@ -440,13 +440,14 @@ def _impl(ctx):
             name = "do_not_split_linking_cmdline",
         ),
 
-        # Adds simd support, only available with the llvm backend.
+        # Adds simd support
         feature(
             name = "wasm_simd",
         ),
-        # Adds relaxed-simd support, only available with the llvm backend.
+        # Adds relaxed-simd support
         feature(
             name = "wasm_relaxed_simd",
+            implies = ["wasm_simd"],
         ),
         feature(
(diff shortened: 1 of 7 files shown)


sbc100 (approved):

sbc100:
This type of breakage seem reasonable to me.  Seems easy enough for folks to fixup when the upgrade bazel, right?

DoDoENT:
>  Seems easy enough for folks to fixup when the upgrade bazel, right?

It has nothing to do with the Bazel upgrade. If we do that change, then the next emsdk bazel package will be breaking w.r.t. the current one (`emsdk@6.0.1`).

But I agree it's a minor update that should be no problem to figure out for any seasoned Bazel user.

sbc100:
Sorry I guess I meant "when upgrading the emsdk bazel package".

DoDoENT:
I made the API break in the last commit. Please review it.

sbc100 (approved):

PR #1735, opened 2026-07-15
Drop testing for bazel 7

sbc100:
The CI for bazel 7 + linux is failing for unknown reasons so maybe its time to simply drop support for this old version.

### DIFF
--- a/bazel/emscripten_cache.bzl
+++ b/bazel/emscripten_cache.bzl
@@ -86,10 +86,7 @@ def _emscripten_cache_repository_impl(repository_ctx):
 
         # Use the prebuilt cache
         use_builtin_cache = False
-
-        # Bazel 7 does not have the repo_metadata API, so prebuilt cache on Bazel 7 will not be marked as reproducible. This is not ideal, but it is a limitation of Bazel 7.
-        if hasattr(repository_ctx, "repo_metadata"):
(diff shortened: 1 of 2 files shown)


DoDoENT (approved):
You can also remove [this if](https://github.com/emscripten-core/emsdk/blob/main/bazel/emscripten_cache.bzl#L90-L92) - it was a workaround for Bazel 7.

bazel/emscripten_cache.bzl
```diff
@@ -87,10 +87,6 @@ def _emscripten_cache_repository_impl(repository_ctx):
         # Use the prebuilt cache
         use_builtin_cache = False
 
-        # Bazel 7 does not have the repo_metadata API, so prebuilt cache on Bazel 7 will not be marked as reproducible. This is not ideal, but it is a limitation of Bazel 7.
-        if hasattr(repository_ctx, "repo_metadata"):
-            repo_metadata = repository_ctx.repo_metadata(reproducible = True)
```
DoDoENT, line 92: Just remove the `if`, the `repo_metadata = repository_ctx.repo_metadata(reproducible = True)` should now always be executed.

The `if` guarded against this function not being available in Bazel 7. It's very important to define downloaded cache as reproducible.

DoDoENT (approved):

walkingeyerobot (approved):

Issue #1734, opened 2026-07-15
Can we simplify the windows building by always using `-G Ninja`?


sbc100:
If we always using ninja we can simplify the windows build and remove all the handling of different MSVC versions?

We could just mandate that if you want to build anything form source you need to run emsdk from visual sudio command prompt.  (i.e. you are expected to have run vsvarsall.bat already and/or have cl.exe in PATH).

@juj do you have any reason not to use ninja for all your windows builds?

sbc100:
Example PR for how much complexity we could remove: https://github.com/emscripten-core/emsdk/pull/1736

PR #1736, opened 2026-07-15
Always use Ninja generator when building cmake projects form source

sbc100:
When building tools from source, unconditionally use the Ninja
generator.

Since #1105 landed ninja is also always available via emsdk.

This massively simplifies the Windows build process allowing for the
removal of `--vs2019`, `--vs2022`, `--mingw` and `--generator` flags
along with `vswhere` detection, Visual Studio multi-config output paths,
and the `mingw` SDK package.

This also allows us to completely remove all the `NUM_CORES` handling
since ninja is really good and using all the cores automatically (no need to
pass -j at all).   If folks really want to control the parallelism they can still
do so using the `CMAKE_BUILD_PARALLEL_LEVEL` environment variable (https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html)

Instead, we now rely on MSCV being configured in the environment prior
to running emsdk.  i.e. We expect windows users to run in a Visual Studio
command prompt (or with `vcvarsall.bat` having been already run).  This
seems like a more sensible soltution that trying to build visual studio selection
into emsdk itself.  i.e. we can just expect users to have already selected their
prefered visual studio version.

See: #1734


sbc100:
I'm adding `ninja` to the manifest so it should be super easy to install too https://github.com/emscripten-core/emsdk/pull/1105

sbc100:
@juj, what do you think about this?  Can we just standardize of the Ninja generator?

sbc100:
In the emscripten-releases repo we already are ninja-only: https://chromium.googlesource.com/emscripten-releases/+/refs/heads/main/src/build.py

PR #1737, opened 2026-07-21
Use $CFGDIR in generated .emscripten for emscripten 6.0.4+

sbc100:
For emscripten version 6.0.4 and above, update the generated `.emscripten` config file to use `$CFGDIR` directly in config path strings instead of including `import os` and constructing paths dynamically with `emsdk_path`.

Older versions of emscripten (< 6.0.4) will continue to generate and use the legacy config format.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1784,8 +1784,17 @@ def get_emsdk_shell_env_configs():
 
 
 def generate_em_config(active_tools, permanently_activate, system):
-  cfg = 'import os\n'
-  cfg += "emsdk_path = os.path.dirname(os.getenv('EM_CONFIG')).replace('\\\\', '/')\n"
+  emroot = find_emscripten_root(active_tools)
+  version = None
+  if emroot and os.path.exists(os.path.join(emroot, 'emscripten-version.txt')):
+    version = parse_emscripten_version(emroot)
+
+  supports_cfgdir = version and version >= [6, 0, 4]
+
+  cfg = ''
+  if not supports_cfgdir:
+    cfg += 'import os\n'
+    cfg += "emsdk_path = os.path.dirname(os.getenv('EM_CONFIG')).replace('\\\\', '/')\n"
 
   # Different tools may provide the same activated configs; the latest to be
   # activated is the relevant one.
@@ -1806,9 +1815,7 @@ def generate_em_config(active_tools, permanently_activate, system):
     else:
       cfg += f"{name} = '{value}'\n"
 
-  emroot = find_emscripten_root(active_tools)
-  if emroot:
-    version = parse_emscripten_version(emroot)
+  if version:
     # Older emscripten versions of emscripten depend on certain config
     # keys that are no longer used.
     # See https://github.com/emscripten-core/emscripten/pull/9469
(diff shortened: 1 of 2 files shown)


emsdk.py
```diff
@@ -2331,8 +2341,16 @@ def parse_emscripten_version(emscripten_root):
...
     version = f.read().strip()
-    version = version.strip('"').split('-')[0].split('.')
-    return [int(v) for v in version]
+  version = version.strip('"')
+  suffix = None
+  if '-' in version:
+    version, suffix = version.split('-')
+  version = [int(v) for v in version.split('.')]
+  if suffix == 'git':
+    # Treat 1.2.3-git as 1.2.2 since its doesn't contain all the changes
```
dschuff:
```suggestion
    # Treat 1.2.3-git as 1.2.2 since it doesn't contain all the changes
```

dschuff (commented):
should we wait on emsdk until we have a final plan for how we're going to evolve the config file format?

test/test.py
```diff
@@ -223,9 +223,34 @@ def test_list(self):
...
+    self.assertNotIn('$CFGDIR', config)
+
+    # For newer versions we use $CFGDIR
+    version_file = os.path.join('upstream', 'emscripten', 'emscripten-version.txt')
+    with open(version_file) as f:
+      old_ver = f.read()
+    try:
+      with open(version_file, 'w') as f:
+        f.write('6.0.4\n')
+      run_emsdk('activate 6.0.0')
```
kripken: Shouldn't this be 6.0.4?

emsdk.py
```diff
@@ -2331,8 +2341,16 @@ def parse_emscripten_version(emscripten_root):
...
+    # that present in the final 1.2.3 release.
```
kripken:
```suggestion
    # present in the final 1.2.3 release.
```
(or "that are present")

kripken (commented):
lgtm % existing comments

sbc100:
> should we wait on emsdk until we have a final plan for how we're going to evolve the config file format?

I guess its not a bad idea to only evolve the format once.   Maybe lets see how the discussions play out.

This intermediate change that we have landed allows the simplification of the config files and by far the biggest offender in terms of "user who use os.path in their config files" is likely this emsdk-generated one, which is why I was trying to remove this usage.

But you are probably right, if I get my way there will be another evolution coming :)

test/test.py
sbc100: Strangely enough no, this is reactivation after hacking the embedded version number.  Its a bit strange though.. maybe better to wait until 6.0.4 is actually released.

sbc100:
I had a chat with @kripken today about the future of the config file format.

We decided that this approach (i.e. simplifying the existing config files) is probably the easiest way forward.   No need to a new config file format for now.

So I think we should land this change to simplify the config file use by emsdk.

dschuff:
sounds good to me.
How close are we to just being able to have the emsdk ship with no config file at all, and just use defaults for those users?

sbc100:
> sounds good to me. How close are we to just being able to have the emsdk ship with no config file at all, and just use defaults for those users?

I think a config file will likely always be a good idea for emsdk.

The only way to avoid the config file its to pollute the users PATH with our version of clang / binaryen / etc.   I kind of like having emsdk avoid polluting the PATH (except for emcc itself of course)

PR #1739, opened 2026-07-21
Remove reference to emsdk-shipped version of git

sbc100:
Git was removed in #395.

Also remove the `must_succeed` argument since all the callers pass zero args.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -739,45 +739,19 @@ def run_get_output(cmd, cwd=None):
   return (process.returncode, stdout, stderr)
 
 
-cached_git_executable = None
-
-
-def GIT(must_succeed=True):
-  """Return the path to the git executable.
-
-  :param must_succeed: If false, the search is performed silently without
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -742,42 +742,23 @@ def run_get_output(cmd, cwd=None):
...
-  :param must_succeed: If false, the search is performed silently without
-    printing out errors if not found. Empty string is returned if git is not
-    found.
-    If true, the search is required to succeed, and the execution will
-    terminate with sys.exit(1) if not found.
-  """
+def GIT():
   global cached_git_executable
-  if cached_git_executable is not None:
+  if cached_git_executable:
```
dschuff: Is it really worth having this cache? shutil.which doesn't seem terribly slow, and I would guess that this isn't called so often that it's performance-critical?

dschuff (approved):

emsdk.py
sbc100: Good point, probably not worth it.

PR #1743, opened 2026-07-21
Bump min python version to 3.10

sbc100:
This gives use the `text=True` alias in the subprocess API, which
was already being used in some places here.

We also get access to `subprocess.run` which was added in 3.5.

We also get ordered dictionaries by default.

Also, update the min version of python mentioned in the README.md,
which refers the minimum version required by emcc, not emsdk.py,
although they happen to match for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ https://emscripten.org/docs/building_from_source/toolchain_what_is_needed.html.
 
 ### Linux
 
-- `python`: Version 3.8 or above.
+- `python`: Version 3.10 or above.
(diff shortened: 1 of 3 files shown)


dschuff:
3.7 is [EOL](https://devguide.python.org/versions/) since 2023; should we update to 3.10 or 3.11 instead?

sbc100:
Yes maybe 3.10 to match the emscripten min version?

sbc100:
Done

dschuff (approved):

PR #1751, opened 2026-07-24
Update ruff to latest version. NFC

sbc100:


dschuff (approved):
who would have thought that giving things useful names would be such a great idea?!

PR #1753, opened 2026-07-27
Remove redundant cmake-missing error handling. NFC

sbc100:
These errors cannot occur because the find_cmake function above will exit_with_error when cmake is missing.

View with "hide whitespace" to see the change more clearly.


sbc100:
View with "hide whitespace" to see the change more clearly.

kripken (approved):

PR #1760, opened 2026-07-31
Update Node.js to v24.19.0

### LINKED ISSUE #1758: Node is too old for JSPI
With latest emsdk (6.0.4), which comes with node 22.16.0, modules compiled with -sJSPI will fail to run:

```
$ echo 'int main(){}' > test.c

$ emcc -sJSPI test.c -o test

$ ./test
Aborted(Assertion failed: JSPI not supported by current environment. Perhaps it needs to be enabled via flags?)

$ head -n 1 test
#!/emsdk/node/22.16.0_64bit/bin/node

$ /emsdk/node/22.16.0_64bit/bin/node --experimental-wasm-jspi test
Aborted(Assertion failed: JSPI not supported by current environment. Perhaps it needs to be enabled via flags?)
```

Even with the correct flag, it doesn't run because this version of node only provides `Suspender` (the older JSPI implementation), and not the newer `Suspending`.

**Suggestion**
Upgrade to node 24.18.1, which does support the new JSPI. (still behind a flag)

sbc100:
Update the version of Node.js shipped with emsdk to the latest LTS.

One additional change here is that since #1761 we can now have a different activated_path for windows to other platforms.  This allows us to use the upstream node archives without modifying the windows versions (they ship node.exe at the top level and not under `bin/`).

Fixes: #1758

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -69,7 +69,7 @@ COPY --from=stage_build /emsdk /emsdk
 # using `--entrypoint /bin/bash` in CLI).
 # This corresponds to the env variables set during: `source ./emsdk_env.sh`
 ENV EMSDK=/emsdk
-ENV PATH="/emsdk:/emsdk/upstream/emscripten:/emsdk/node/22.16.0_64bit/bin:${PATH}"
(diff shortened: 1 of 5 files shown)


kripken (approved):

sbc100:
I think I'll leave this open until next week at least to give @dschuff and @juj a chance to comment.

PR #1761, opened 2026-08-03
Add OS suffix support for all tool properties

sbc100:
We had specific hardcoded support for OS-specific URLs.  This change makes that generic and allows `_windows` / `_linux` / `_mac` suffixes to work with any key.

This change is really a no-op that makes way to using a windows specific activated_path for now (allowing us to use the upstream windows node archive, rather than re-packaging/hacking it in (scripts/update_node.py).  See #1760.

Also, remove support for `os: "all"`.. this is not used and doesn't seem useful (since you can just omit it).


sbc100:
If you like I could split out the removal of the `unix` suffix?

kripken (approved):

PR #1763, opened 2026-08-03
Update `tot` version to be based on binary availability on all platforms

sbc100:
When resolving tip-of-tree (tot) release in get_emscripten_releases_tot, check that prebuilt binary archives exist for all standard TOT platforms (linux x86-64, mac x86-64, mac arm64, win x86-64) before selecting a revision.

Previously, emsdk only checked binary availability for the host OS and architecture running emsdk. When emsdk-update CI ran on Linux, it could select a TOT revision before macOS or Windows binaries finished uploading.

This has been causing issues in emscripten where the update-emsdk job (which updates test/emsdk_version.txt) was choosing a version that was available on linux, but not yet on all other platforms, causing the CI jobs to fail on some platforms (e.g. windows).

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -2328,34 +2328,27 @@ def get_emscripten_releases_tot():
   recent_releases = git_recent_commits(sdk_path('releases'))
   # The recent releases are the latest hashes in the git repo. There
   # may not be a build for the most recent ones yet; find the last
-  # that does.
-  arch = ''
-  if ARCH == 'arm64':
-    arch = '-arm64'
+  # that has uploaded binaries for all TOT platforms.
+  all_platforms = [
+    ('linux', '', 'tar.xz'),
+    ('mac', '', 'tar.xz'),
+    ('mac', '-arm64', 'tar.xz'),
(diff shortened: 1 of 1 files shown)


emsdk.py
```diff
@@ -2341,34 +2341,27 @@ def get_emscripten_releases_tot():
...
-      release,
-      arch,
-      ext,
-    )
-
-  for release in recent_releases:
-    make_url('tar.xz' if not WINDOWS else 'zip')
+  # that has uploaded binaries for all TOT platforms.
+  all_platforms = [
+    ('linux', '', 'tar.xz'),
```
sbc100, line 2333: I would include the linux-arm64 build here but currently we are only uploading those builds for LTO releases.

See https://chromium.googlesource.com/emscripten-releases/+/refs/heads/main/src/build.py#1132

Should we lift this restriction?   @dschuff ?

kripken (approved):

sbc100:
@dschuff just wanted you to confirm that this makes sense to you too?

PR #1764, opened 2026-08-03
Remove `unix_url` tool property.

sbc100:
Instead, we can just have `window_url` and the use the default `url` as the fallback.

This simplification is part of the larger change to how we do OS-specific properties.  See #1761


sbc100:
This is the first change in the stack of 3.. checkout the github stacks UI!

emsdk.py
```diff
@@ -1943,11 +1942,11 @@ def compatible_with_this_os(self):
       assert self.os in {'all', 'linux', 'win', 'macos'}
       if self.os == 'all':
         return True
-      if (WINDOWS and self.os == 'win') or (LINUX and (self.os in {'linux', 'unix'})) or (MACOS and (self.os in {'macos', 'unix'})):
```
kripken: Do you know why this was needed in the first place?

sbc100: I think the idea was so that one could specify "windows_url" (with .zip) and "unix_url" (with .tar.gz).

But I imagine that we always supported the fallback to "url" as the default .. I don't think this was ever strictly necessary.

sbc100: Indeed, the `unix_` prefix was introduced in fb549cdf5e0 way back in 2013, but even then it was not needed since `url` was always the default.

PR #1766, opened 2026-08-03
Track active tools and SDKs in activated.json

sbc100:
Store active tools and SDKs in a dedicated `activated.json` file when they are activated by `emsdk activate`. Update `Tool.is_active()` to check against `activated.json` rather than matching entries in `.emscripten`.

This solves several issues with the previous `.emscripten`-based `is_active()` logic:

1. Tools with no `activated_cfg` (such as `ccache` or tools that only modify `PATH` or environment variables) previously returned `is_active() == False` because they had no `.emscripten` keys and no dependencies (`len(deps) == 0`).
2. Tools with dependencies but no `activated_cfg` previously falsely reported `is_active() == True` whenever their dependencies were active, even if the tool itself was never activated.

This allows us to remove all of the `activated_cfg` entries that were there only to keep track of tool activation.

Replaces: #1345

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 *.pyc
 __pycache__
 
+# Track which tools are currently active
+/activated.json
+
 # Support for --embedded configs
 /.emscripten
 /.emscripten.old
--- a/emsdk.py
+++ b/emsdk.py
@@ -1597,7 +1597,94 @@ def get_required_path(active_tools):
 # Returns the absolute path to the file '.emscripten' for the current user on
 # this system.
 EM_CONFIG_PATH = os.path.join(EMSDK_PATH, ".emscripten")
-EM_CONFIG_DICT = {}
+ACTIVATED_JSON_PATH = os.path.join(EMSDK_PATH, "activated.json")
+ACTIVATED_TOOLS = set()
+
+
+def get_active_legacy():
+  """Legacy support for users without existing `activated.json` files.
+
+  Older versions of emsdk would imply active status purely based on the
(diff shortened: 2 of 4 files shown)


kripken:
The new behavior seems reasonable, but also changing the current long-standing behavior seems risky. How much of a benefit do you estimate our users would get from this?

sbc100:
I think its much cleaner and clearer design.    Without this we need to have dummy config vars in the `.emscripten` file that just server as magic strings to show that a tool is active.   Its odd at best and confusing at worst. 

For example there on mac an windows you will see `PYTHON=` the `.emscripten` config file even though nobody reads this or uses it.   Its very confusing to have a setting like that that doesn't do anything.

We need the invent new functional `FOO=bar` settings for *all* tools, just to support making `is_active()` work as expected.

sbc100:
I tired to mitigate the potential risk by building migration system from the old to new system.   The idea is that maybe one day we can remove this.

kripken:
Maybe I'm influenced by my memories of the last time we tried this. Iirc we ended up with some bugs around how many installed-but-not-activated tools/sdks we had, leading to user bug reports of disk space getting exhausted. I'm not saying that is a risk specifically, but that this is complicated and involves stuff that is hard to test comprehensively.

Are there specific bugs that this fixes? That could change my mind. I am mainly worried about a refactor for cleanup purposes only.

sbc100:
I'll try to come up with some more clear rationale why I think this is an important step, and how I think we can mitigate any risks...

PR #1767, opened 2026-08-04
[test] Clean up top-level directory after test runs

sbc100:
Add setup and teardown handlers to `test/test.py` to clean up temporary files and output artifacts created during unit tests.

Clean up `hello_world.c`, `a.out*` build artifacts, longpath test directories, and temporary `hack_emsdk` files.

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -1,8 +1,10 @@
 #!/usr/bin/env python3
+import glob
 import json
 import os
 import platform
 import shutil
+import stat
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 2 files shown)


test/test.py
```diff
@@ -155,11 +174,17 @@ def setUpClass(cls):
    return 0;
 }
 ''')
+    self.addCleanup(remove_file, 'hello_world.c')
 
   def setUp(self):
+    self.make_hello_world()
```
kripken: This now runs on each test, not just once for the class. Isn't that slower? is there a benefit?

sbc100: There are very few tests here, and most of them take many seconds to run, so this is a microscopic cost.

It just turned out to be the easiest way to organize the code.

kripken (approved):

PR #1768, opened 2026-08-04
Merge tools now that OS-specific keys are possible

sbc100:
Since #1761 we now have os-specific keys so we don't need separate entries in these cases anymore.

### DIFF
--- a/emsdk.py
+++ b/emsdk.py
@@ -1962,7 +1962,10 @@ class Tool:
       assert self.os in {'linux', 'windows', 'macos'}
       return self.os == os_name()
 
-    return self.url is not None
+    # Tools without an explicit 'os' are compatible with the current
+    # OS if they have URL for the current OS *or* a set of dependencies
+    # for the current OS.
+    return self.url is not None or bool(self.deps)
 
   def get_version_file_path(self):
     """the "version file" is a file inside install dirs that indicates the
--- a/emsdk_manifest.json
+++ b/emsdk_manifest.json
@@ -280,16 +280,10 @@
     "arch": "x86_64",
     "activated_cfg": "EMSDK_CMAKE='%installation_dir%/bin/cmake%.exe%'",
     "activated_path": "%installation_dir%/bin",
+    "activated_cfg_macos": "EMSDK_CMAKE='%installation_dir%/CMake.app/Contents/bin/cmake%.exe%'",
+    "activated_path_macos": "%installation_dir%/CMake.app/Contents/bin",
     "url_windows": "https://github.com/Kitware/CMake/releases/download/v4.2.0-rc3/cmake-4.2.0-rc3-windows-x86_64.zip",
-    "url_linux": "https://github.com/Kitware/CMake/releases/download/v4.2.0-rc3/cmake-4.2.0-rc3-linux-x86_64.tar.gz"
-  },
-  {
-    "id": "cmake",
-    "version": "4.2.0-rc3",
-    "bitness": 64,
-    "arch": "x86_64",
-    "activated_cfg": "EMSDK_CMAKE='%installation_dir%/CMake.app/Contents/bin/cmake%.exe%'",
-    "activated_path": "%installation_dir%/CMake.app/Contents/bin",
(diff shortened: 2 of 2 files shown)


emsdk_manifest.json
```diff
@@ -679,54 +649,19 @@
...
-    "os": "macos",
-    "arch": "arm64",
-    "custom_install_script": "sdk_post_install"
-  },
   {
     "version": "releases-%releases-tag%",
     "bitness": 64,
     "uses": ["node-24.19.0-64bit", "python-3.13.3-64bit", "releases-%releases-tag%-64bit"],
-    "os": "windows",
+    "uses_linux": ["node-24.19.0-64bit", "releases-%releases-tag%-64bit"],
```
kripken: What does it mean to have both `uses` and `uses_linux`?

sbc100: "uses" is the default set of dependencies.

"uses_linux" is a linux-specific override.

I think we should rename "uses" to "deps" and it might make this more obvious.  Maybe I'll do that first.

sbc100: OK, now that #1772 has landed this part should be more readable

emsdk.py
```diff
@@ -1962,7 +1962,9 @@ def compatible_with_this_os(self):
       assert self.os in {'linux', 'windows', 'macos'}
       return self.os == os_name()
 
-    return self.url is not None
+    # Tools/SDKs without an explicit 'os' filter are compatible if they specify
+    # a download URL *or* have dependencies.
+    return self.url is not None or bool(self.deps)
```
sbc100, line 1968: The reason this is needed is the previously the "sdks" were listed as one-per-os, but now there is just a single one with `deps` and `deps_linux`.    

Prior to this change the return statement above on line 1963 would always fire for SDKs

emsdk.py
```diff
@@ -1962,7 +1962,9 @@ def compatible_with_this_os(self):
       assert self.os in {'linux', 'windows', 'macos'}
       return self.os == os_name()
 
-    return self.url is not None
+    # Tools/SDKs without an explicit 'os' filter are compatible if they specify
+    # a download URL *or* have dependencies.
```
kripken: But why does having dependencies make a tool compatible? E.g. some build system might only run on linux/mac, so it has deps for each of those, but windows wouldn't want it?

That is, having deps is a "structural" property. it doesn't say who or what should use it.

sbc100: If if you only specify `mac_deps` and `windows_deps` and then on linux the `self.deps` would be empty here.  i.e. it would be uninstallable.

What this logic is saying is that to be installable you need to either have a `url` or `deps` for the current platforms.   If you have neither of them then you are not installable.

i.e. its ok to have url for current platform *or* not deps for the current platform, but its not OK to have neither.

The reason this wasn't needed because is that the SDK had an entry for each OS with a specific `os = 'linux'`.  Now that the SDK is generic we need this line to be able to deal with a tools with no URL and only deps (i.e. an SDK).

Yes another way of putting it:  Prior to this change all the deps-only tools were OS-specific, but there is no reason you can't have a deps-only tool that is generic, and that requires this change (or something like it).

sbc100: I update the comment to make this more clear.  Both `self.url` and `self.deps` are by construction working on the current OS.

kripken (approved):

emsdk.py
```diff
@@ -1962,7 +1962,10 @@ def compatible_with_this_os(self):
       assert self.os in {'linux', 'windows', 'macos'}
       return self.os == os_name()
 
-    return self.url is not None
+    # Tools without an explicit 'os' are compatible with the current
+    # OS if they have URL for the current OS *or* a set of dependencies
+    # for the current OS.
```
kripken, line 1967:
```suggestion
    # for the current OS (`deps` by itself, unlike `deps_linux`, applies to all OSes, so all are compatible with it).
```

sbc100: Its a bit more subtle than that.   `self.deps` is the deps for the current OS.   So if either `deps_linux` or `deps` (no suffix) exist in the file then `deps` will be non-null here.

Issue #231, opened 2019-04-05
emsdk activate fails when install-dir contains unicode-characters


WereCatf:
I have Ubuntu localized for Finnish and desktop is called "Työpöytä" -- if I've downloaded emsdk-git to my desktop and issue `./emsdk activate latest` it errors out with `UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 13: ordinal not in range(128)`

Moving the emsdk-directory to a location without unicode-characters fixes this.

Not a biggie, but considering how many people there are out there with unicode-characters in their names and/or directories, this might be a good idea to fix at some point.

kripken:
Thanks, yeah, we should fix this. Looking at the full stack trace from that error, it should be clear which line needs to be fixed.

sbc100:
I'm pretty sure this was fixed when we updated to python3.   Adding a test in #1770 just to be sure.

PR #1770, opened 2026-08-05
[test] Add test for running emsdk from a unicode path

sbc100:
Add a test verifying that emsdk commands function properly when executed from a directory path containing unicode characters.

See: #231


sbc100:
@kripken gentle ping..

kripken (approved):
Sorry, not sure how I missed this last week...

PR #1775, opened 2026-08-05
[bazel] Use em++.py in link_wrapper.py for C++ linking

sbc100:
Use em++.py instead of emcc.py in Bazel's `link_wrapper.py`. In recent versions of Emscripten, emcc no longer links as C++ by default.    See  https://github.com/emscripten-core/emscripten/pull/27469.

Since Bazel uses this link wrapper for C++ executable/library linking, invoking em++.py ensures C++ standard library symbols are properly resolved.

### DIFF
--- a/bazel/emscripten_build_file.bzl
+++ b/bazel/emscripten_build_file.bzl
@@ -29,6 +29,7 @@ filegroup(
     name = "emcc_common",
     srcs = [
         "emscripten/emcc.py",
+        "emscripten/em++.py",
         "emscripten/embuilder.py",
         "emscripten/emscripten-version.txt",
         "emscripten/cache/sysroot_install.stamp",
--- a/bazel/emscripten_toolchain/link_wrapper.py
+++ b/bazel/emscripten_toolchain/link_wrapper.py
@@ -37,7 +37,7 @@ if any(' ' in a for a in param_file_args):
       f.write('\n')
   sys.argv[1] = '@' + new_param_filename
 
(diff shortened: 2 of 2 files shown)


sbc100:
I'm not sure if using em++ always as the linker is the best long term solution, but it will at least allow us to release 6.0.6.

What does bazel normally do here?

DoDoENT (approved):

DoDoENT:
> I'm not sure if using em++ always as the linker is the best long term solution, but it will at least allow us to release 6.0.6.
> 
> 
> 
> What does bazel normally do here?

Bazel in itself doesn't do anything. The question is what do rules_cc do? And I think they just delegate to toolchain, which is the emsdk here. 

I'm currently on a limited connectivity until late August so I can't dive deep into details of how everything works, but I'm confident that always linking with em++.py should not be a problem even for pure C programs. Namely, using c++ linker driver on other platforms is just a "syntax sugar" to automatically link with libc++ and libc++abi libs, which, if not used, can always be thrown away by the linking process.

sbc100:
I guess a better question would be: How does bazel choose gcc vs g++ (or clang vs clang++) when linking?  Or does it always assume C++?

DoDoENT:
> I guess a better question would be: How does bazel choose gcc vs g++ (or clang vs clang++) when linking?  Or does it always assume C++?

I don't know so I asked Gemini. Apparently it's a sophisticated heuristics... 

Here it's its answer: https://share.gemini.google/MsaVrSmRsSjV

Sorry for not being more helpful. I'm currently on a vacation behind a limited internet connection.

PR #1778, opened 2026-08-06
Improve duplicate and error reporting when parsing emsdk_manifest.json

sbc100:


emsdk.py
```diff
@@ -2214,16 +2214,21 @@ def recursive_dependencies(self):
     return deps
 
 
+def dict_to_str(d):
+  return '{' + ', '.join(f'{key}: {value}' for key, value in vars(d).items()) + '}'
```
kripken: why not do `str(d)`?

sbc100: `str(Tool)` resolves to its name (e.g. `binaryen-tag-1.38.4-64bit`)

kripken: How about `str(d.__dict__)`?

sbc100: Hmm.. maybe  we can do `str(vars(d))`  I will double check

sbc100: That worked great.  Thanks for pointing this out!

kripken (approved):

PR #1785, opened 2026-08-06
Convert `gsutil` usage to `gcloud storage`. NFC

sbc100:
`gsutil` is being phased out. Update out script to use `gcloud storage` instead.

### DIFF
--- a/scripts/update_node.py
+++ b/scripts/update_node.py
@@ -46,7 +46,7 @@ def main():
     if args.upload:
       upload_url = upload_base + filename
       print('Uploading: ' + upload_url)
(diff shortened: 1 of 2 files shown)


sbc100:
Hmm.. actually I think I misread the announcement.    gsutil is sticking around but as a separate package from gcloud.  I guess there is no need to switch after all?

sbc100:
https://docs.cloud.google.com/storage/docs/gsutil_install#install

dschuff:
I suspect this might still be worth doing. It seems like gsutil is effectively becoming third-party (and probably still ends up depending in some way on cloud CLI?) in which case maybe we want to just move to gcloud storage anyway?

PR #1788, opened 2026-08-12
Include linux-arm64 when determining `tot` version

sbc100:


kripken (approved):
Wait, do we ship official linux-arm64 builds from chromium ci?

sbc100:
> Wait, do we ship official linux-arm64 builds from chromium ci?

Yes, we've been doing that for several years now.

The new thing is that we don't just upload the LTO builds anymore

PR #1789, opened 2026-08-13
Fix PowerShell scripts to support StrictMode

### LINKED ISSUE #1337: The variable '$EMSDK_PY' cannot be retrieved because it has not been set.
commit: f2baa2f56a0c508c33ab2e16ef1efd3d548f42b8
PowerShell version: PowerShell 7.4.0

The installation script for PowerShell: "emsdk.ps1" is missing variable declaration for "$EMSDK_PY". Causing an exception
The variable '$EMSDK_PY' cannot be retrieved because it has not been set.

The simplest workaround is to add a statement $EMSDK_PY="" at the beginning of that script.

sbc100:
Add `Set-StrictMode -Version latest` to PowerShell scripts. Initialize `$EMSDK_PY` to `$null` and use `-ErrorAction SilentlyContinue` when removing environment variables so execution under strict mode does not fail on uninitialized variables or missing env items.

Fixes: #1337

### DIFF
--- a/test/test_activation.ps1
+++ b/test/test_activation.ps1
@@ -2,7 +2,7 @@
 # and checks if the environment variables and PATH are correctly updated. Set $env:SYSTEM_FLAG and $env:PERMANENT_FLAG to test each.
 # If no flag is provided the process/shell values are tested. See the CI file for an example.
 
-Set-StrictMode -Version latest
+Set-StrictMode -Version 3.0
 
 refreshenv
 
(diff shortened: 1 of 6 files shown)


test/test_activation.ps1
```diff
@@ -2,6 +2,8 @@
 # and checks if the environment variables and PATH are correctly updated. Set $env:SYSTEM_FLAG and $env:PERMANENT_FLAG to test each.
 # If no flag is provided the process/shell values are tested. See the CI file for an example.
 
+Set-StrictMode -Version latest
```
kripken: Should we set a specific version here, to avoid forward compat issues?

sbc100: Done

kripken (approved):

PR #1794, opened 2026-09-06
Teach a python script about git work-trees

### LINKED ISSUE #1793: Can not build if LLVM is a work-tree.
Hi, `emsdk` depends on some other `git` repositories. Such as `binaryen`, `emscripten` or `llvm`. The `llvm` repository especially is very huuuge. When building `emsdk` I don't want to download `llvm` if I already have it somewhere else for different purposes. I would like to use the `work-tree` technology of `git`.

So, instead of downloading the entirety of `lllvm`, I "just" create a new `git` `work-tree` instead.

But the `python` script in `emsdk` is not ready for this eventuality. It is not prepared for this scenario. It behaves as-if `git` `work-tree` functionality never existed, was never invented.

In this issue I propose to teach your `python` script about this eventuality, this would save me a lot of useless downloading.

MarekKnapek:
With git worktrees, `.git` is actually a text file that contains a kind of user-space symlink to the parent.

Fixes: #1793


emsdk.py
```diff
@@ -770,7 +770,7 @@ def get_git_remotes(repo_path):
 
 def git_clone(url, dstpath, branch, remote_name='origin'):
   debug_print(f'git_clone(url={url}, dstpath={dstpath})')
-  if os.path.isdir(os.path.join(dstpath, '.git')):
+  if os.path.isdir(os.path.join(dstpath, '.git')) or os.path.isfile(os.path.join(dstpath, '.git')):
```
sbc100, line 773: How about just `os.path.exists`?

MarekKnapek: Because I'm a C guy, and I had no idea it existed. Fixed in d9c7b0f49928e3f1371150711972bb6b1c39d686.

sbc100 (approved):

PR #1799, opened 2026-09-21
Re-indent python using 4-space indentation. NFC

sbc100:
This change was generated using:

```
$ pip3 install reindent
$ git ls-files '*.py' | xargs reindent -n
```

Why is this good idea?  Mostly for consistency with the rest of the python world.  Other projects in our ecosystem already use 4-space indentation:

- binaryen: Converted to 4-space a while back
- llvm-project: Has always used 4-space IIUC
- emscripten-releases: Converted to 4-space a while back
- wabt: Has always used 4-space
- chromium: Start converted to 4-space several years back.

This is also kind a trial run for doing the same thing in emscripten. Hopefully it won't make rebasing outstanding branches too painful (AI agents are really good a re-basing these days anyway).


sbc100:
Obviously best viewed with "Hide whitespace" enabled.

dschuff (approved):
After landing, probably worth adding the subsequent commit to .git-blame-ignore-revs

sbc100:
CC @juj, just wanted to give you heads up since you likely have some outstanding PRs against this repo.

In my experience AI agents are really good at rebasing outstanding change in the face of this kind of reformat.   If this change goes well I plan on doing the same thing for emscripten.

If this change does *not* go well and causes us / you huge headaches, then that will be good signal that we should maybe not do that same for emscripten.

